PR #88, opened 2026-03-21
Plan: pydantic/pydantic-harness#42

adtyavrdhn:
Implementation plan for pydantic/pydantic-harness#42

Generated via dual-model debate (anthropic/claude-opus-4-6 + anthropic/claude-opus-4-5-20251101).

See: https://github.com/pydantic/pydantic-harness/issues/42


PLAN.md
```diff
@@ -0,0 +1,3056 @@
...
+> https://github.com/pydantic/pydantic-harness/issues/42
+
+> Generated via dual-model debate (anthropic/claude-opus-4-6 + anthropic/claude-opus-4-5-20251101)
+> Date: 2026-03-21
+> Issue graph: 30 items crawled
+
+## Linked Issues & PRs
+
+- issue #1: [Tracking] Capabilities abstraction for pydantic-ai (https://github.com/pydantic/pydantic-harness/issues/1)
+- issue #29: Approval capability (https://github.com/pydantic/pydantic-harness/issues/29)
```
dsfaccini, line 13: too many issues linked

PLAN.md
```diff
@@ -0,0 +1,3056 @@
...
+| **Claude Code** | Built-in `AskUserQuestion` tool with multi-question and multi-select support, always-allowed | [Claude Code docs](https://code.claude.com/docs/en/hooks) | Multi-select, batch questions | Not open-source implementation |
+
+**Key insight:** The most useful pattern (Mastra, Claude Code) is a tool the LLM can choose to invoke, not a task-level config flag (CrewAI). Pydantic-ai's `CallDeferred` mechanism (`exceptions.py:77-99`) is semantically the correct choice: it's for tools whose results come from an external source and produces entries in `DeferredToolRequests.calls` with arbitrary string values — exactly the pattern needed for free-text user answers. `ApprovalRequired` (`exceptions.py:95-114`) is wrong because it's for yes/no authorization and triggers tool re-execution with `ctx.tool_call_approved=True`.
+
+## Approach
+
+The `UserInteraction` capability provides an `ask_user` tool via `get_toolset()` and guidance via `get_instructions()`. The capability operates in two modes:
```
DouweM, line 63: "UserInteraction" is too vague IMO, it doesn't communicate that this is specifically about asking the user questions -- there are many types of "human interaction", of which approval is one, for example

PLAN.md
```diff
@@ -0,0 +1,3056 @@
...
+
+**Deferred mode (default, no handler):** The tool raises `CallDeferred` with question metadata. The agent graph layer in `_agent_graph.py:1365-1370` catches the exception and automatically associates it with the current `ToolCallPart`'s `tool_call_id`. The run ends with `DeferredToolRequests`. The host application presents the question, then resumes with `DeferredToolResults` containing the answer keyed by `tool_call_id`. The answer becomes the tool's return value directly — no re-execution.
+
+**Inline mode (with handler):** When `input_handler` is provided, the tool calls it directly and returns the answer without pausing. Async handlers (detected via `inspect.iscoroutinefunction()`) are awaited. This enables terminal `input()`, callbacks, or any custom resolution.
```
DouweM, line 67: This is related to https://github.com/pydantic/pydantic-ai/issues/3959, which suggests a global way of handling deferred/approval tool calls using a handler rather than stopping the run on `DeferredToolRequests`. I think it should be possible to build a generic capability that lets you set a handler for any deferred tool call, using the tool hooks. Worth exploring if we don't have that already. Likely better than this each capability having its own handler, although I don't mind having duplicate paths in this particular case

PLAN.md
```diff
@@ -0,0 +1,3056 @@
...
+
+The capability is Tier S serializable: `get_serialization_name()` returns `'UserInteraction'`, `from_spec()` accepts no required arguments. The `input_handler` callable is not serializable, so `from_spec()` always creates the deferred variant. This follows the pattern established by the `Instructions` capability where `from_spec()` deliberately narrows to the serializable subset.
```
DouweM, line 69: What does "Tier S" mean lol.

PLAN.md
```diff
@@ -0,0 +1,3056 @@
...
+
+For the `choices` parameter, the tool uses `list[str] | None` as the model-facing type. This is the simplest representation for the LLM to produce correctly, matching the issue description's proposed signature. The type can be widened to structured objects (`[{label, description}]`) in a future non-breaking change if needed.
```
DouweM, line 71: I like having labels from the start, but not a huge deal

PLAN.md
```diff
@@ -0,0 +1,3056 @@
...
+    When ``input_handler`` is None (default), the tool raises ``CallDeferred``, pausing
+    execution so the host application can present the question and resume with the answer.
+
+    Example (deferred mode — default)::
+
+        agent = Agent('openai:gpt-4o', capabilities=[UserInteraction()])
+        result = await agent.run('Help me pick a color')
+        if isinstance(result.output, DeferredToolRequests):
+            call = result.output.calls[0]
+            question = call.metadata['question']
```
DouweM, line 111: - It's a shame that we don't get type safety here
- Is it right for the questions to be in the call metadata, rather than the deferred tool call args?

PLAN.md
```diff
@@ -0,0 +1,3056 @@
...
+    If None, the tool pauses execution via CallDeferred.
+    If the handler blocks (e.g., ``input()``), the agent's ``tool_timeout`` applies.
+    """
+
+    @classmethod
+    def get_serialization_name(cls) -> str:
+        return 'UserInteraction'
+
+    @classmethod
+    def from_spec(cls, **kwargs: Any) -> UserInteraction[Any]:
```
DouweM, line 141: (Not specific to the plan, more about capabilities:) Unsure if we need the `kwargs: Any` bit here as it implies it takes arbitrary args

PLAN.md
```diff
@@ -0,0 +1,3056 @@
...
+            """
+            if handler is not None:
+                if inspect.iscoroutinefunction(handler):
+                    return await handler(question, choices)
+                return handler(question, choices)  # type: ignore[return-value]
+
+            metadata: dict[str, Any] = {'question': question}
+            if choices is not None:
+                metadata['choices'] = choices
+            raise CallDeferred(metadata=metadata)
```
DouweM, line 177: As above, we may not need this since the questions are in the args already

PLAN.md
```diff
@@ -0,0 +1,3056 @@
...
+
+        return toolset
+```
+
+**Why `@toolset.tool_plain`:** The tool doesn't need `RunContext` — it only needs its arguments and the handler closure. This decorator works with async functions as confirmed in `test_capabilities.py:560-600`. The graph layer automatically associates the `CallDeferred` exception with the current `ToolCallPart`'s `tool_call_id` during tool execution in `_agent_graph.py:1365-1370`.
```
DouweM, line 182: This feels like unnecessary detail (tool_plain vs tool) -- that's the kind of thing that it may be better to leave to the codegen agent so it doesn't accidentally get too tightly bound to whatever the planning agent came up with, that may turn out to be different when we really get to code. So I'd like the plan to focus more on the API and behavior and the "flow" (e.g. which hooks to use, what state to store), than literal code snippets like the huge ones we have below.

PLAN.md
```diff
@@ -0,0 +1,3056 @@
...
+
+
+
+# Implementation Plan: UserInteraction Capability (Issue #42)
+
+## Goal
+
+Add a `UserInteraction` capability (`AbstractCapability` subclass) providing an `ask_user` tool that pauses agent execution to gather clarifying information from the user. Supports free-text questions and structured multiple-choice options. Uses pydantic-ai's existing `CallDeferred` mechanism for pause/resume.
+
+## Prior Art & Competitive Landscape
```
DouweM, line 50: I love this section

PLAN.md
adtyavrdhn: I have no idea something it coined for Serializable lol

PLAN.md
adtyavrdhn: Yep, this is useless

PR #138, opened 2026-04-02
Add SubAgent capability for agent-to-agent task delegation

### LINKED ISSUE #32: SubAgent / Agent-as-Tool capability

## Goal

Provide a capability for sub-agent delegation where the parent agent can dispatch tasks to specialized sub-agents while retaining control.

## Context

- **Serializability:** Tier P — `- SubAgents: {agents_dir: ./agents}` (agent specs serializable recursively; dynamic creation requires code)
- **Capability interface methods:** `get_toolset()`
- **Implementation status:** Not started

## Description

Two patterns: **Agent-as-Tool** (parent retains control, sub-agent response returned as tool result) and **full handoff** (control transfers to sub-agent). Tools: `delegate_to_agent`, `list_agents`. Present in OpenAI SDK (`agent.as_tool()`), Google ADK (`AgentTool`), llm-do, pydantic-deepagents (sync/async/auto modes with cancellation), Code Puppy.

Toolset state isolation for sub-agents is a consideration — see [pydantic/pydantic-ai#4347](https://github.com/pydantic/pydantic-ai/issues/4347).

## Depends on

- #10 — AbstractCapability + CombinedCapability + Agent integration

## Related pydantic-ai issues

- [pydantic/pydantic-ai#4347](https://github.com/pydantic/pydantic-ai/issues/4347) — Toolset state isolation (sub-agent isolation)

DouweM:
## Summary

- Adds `SubAgent` capability (`AbstractCapability` subclass) that lets a parent agent delegate tasks to named sub-agents via a `delegate_task` tool
- Sub-agent descriptions are injected into the system prompt (derived from `agent.description`, `agent.name`, or explicit `descriptions` dict)
- Parent deps are forwarded to sub-agents by default (`pass_deps=True`), configurable
- Unknown agent names trigger `ModelRetry` for self-correction
- 100% test coverage (20 tests), passes lint, format, and strict pyright

Closes #32

## Test plan

- [x] Construction: capability is `AbstractCapability`, descriptions resolve from agent metadata or explicit overrides, empty agents returns None for instructions/toolset
- [x] Instructions: system prompt lists all agents with descriptions
- [x] Toolset: `delegate_task` tool is registered with agent names in description
- [x] End-to-end delegation: parent delegates to sub-agent and gets result, unknown agent triggers retry then succeeds, deps forwarded when `pass_deps=True`, deps are None when `pass_deps=False`, multiple agents called sequentially
- [x] Imports: `SubAgent` importable from `pydantic_harness` package

> **Note:** Tests require `pydantic-ai-slim` with capabilities support (the `capabilities` branch). The `pyproject.toml` uses the PyPI dependency; for local development, override with `uv add --dev pydantic-ai-slim@/path/to/capabilities-branch/pydantic_ai_slim`.


DouweM:
Note: Sub-agent runs are currently opaque (parent sees only final text result). Real-time progress streaming from sub-agents to the parent's event stream requires #146 (sub-agent event propagation).

DouweM:
## Audit vs prior art: SubAgent

**Worth adding now:**
- `share_history: bool = False` to pass parent message history to sub-agent
- `delegate_tasks` (plural) for parallel delegation via asyncio.gather
- Preserve structured output type instead of always stringifying

**Follow-up opportunities:**
- Full handoffs (control transfer), streaming delegation, sub-agent event propagation (#146)

src/pydantic_harness/subagent.py
```diff
@@ -0,0 +1,219 @@
...
+    """
+
+    pass_deps: bool = True
+    """Whether to forward the parent agent's `deps` to sub-agents.
+
+    When True (the default), sub-agents receive the same dependency object
+    as the parent. Set to False if sub-agents use incompatible dependency types.
+    """
+
+    share_history: bool = False
```
adtyavrdhn, line 115: This is something that most frameworks do not do, subagents are spawned with their own prompt but no history of what is going on. Worth a discussion

src/pydantic_harness/subagent.py
```diff
@@ -0,0 +1,219 @@
...
+    """Whether to pass the parent agent's message history to sub-agents.
+
+    When True, the parent's conversation history is forwarded as
+    ``message_history`` to each sub-agent run, giving it access to the
+    full conversation context. When False (the default), sub-agents start
+    with a fresh conversation.
+    """
+
+    _resolved_descriptions: dict[str, str] = field(default_factory=dict[str, str], init=False, repr=False)
```
adtyavrdhn, line 124: Unsure why we need this?

src/pydantic_harness/subagent.py
```diff
@@ -0,0 +1,219 @@
...
+        async def _run_sub_agent(ctx: RunContext[AgentDepsT], agent_name: str, task: str) -> str:
+            """Run a single sub-agent, returning its formatted output."""
+            agent = agents.get(agent_name)
+            if agent is None:
+                available = ', '.join(sorted(agents))
+                raise ModelRetry(f'Unknown agent {agent_name!r}. Available agents: {available}')
+
+            deps = ctx.deps if pass_deps else None
+            message_history = _shareable_history(ctx.messages) if share_history else None
+            result = await agent.run(task, deps=deps, message_history=message_history)
```
adtyavrdhn, line 171: Worth checking how much control we want to hand over this.
Point to self: Need to check how background agents if at all will work for us and how the parent agent can still check for responses.

Ideal: CC sort of behaviour where the parent agent yields but can check the progress on the subagents, work on something else meanwhile and can respond back later. Dumping my thoughts here for further research, need not come with this PR at all.

src/pydantic_harness/subagent.py
```diff
@@ -0,0 +1,219 @@
...
+
+        async def delegate_task(ctx: RunContext[AgentDepsT], agent_name: str, task: str) -> str:
+            """Delegate a task to a named sub-agent and return its output.
+
+            Args:
+                ctx: The run context from the parent agent.
+                agent_name: The name of the sub-agent to run. Must be one of the registered agent names.
+                task: The prompt describing the task to delegate.
+            """
+            return await _run_sub_agent(ctx, agent_name, task)
```
adtyavrdhn, line 182: Could be inline

src/pydantic_harness/subagent.py
```diff
@@ -0,0 +1,219 @@
...
+            ctx: RunContext[AgentDepsT],
+            tasks: list[dict[str, str]],
+        ) -> list[str]:
+            """Delegate multiple tasks to sub-agents in parallel and return their outputs.
+
+            Args:
+                ctx: The run context from the parent agent.
+                tasks: A list of task objects, each with ``agent`` (sub-agent name) and ``task`` (prompt).
+            """
+            coros = [_run_sub_agent(ctx, t['agent'], t['task']) for t in tasks]
```
adtyavrdhn, line 194: Ah for this then, nvm

src/pydantic_harness/subagent.py
```diff
@@ -0,0 +1,219 @@
...
+                description=f'Delegate multiple tasks in parallel. Each item needs "agent" and "task" keys. {agent_desc}',
+            ),
+        ]
+        return FunctionToolset[AgentDepsT](tools)
+
+    def _delegate_task_description(self) -> str:
+        """Build a description for the delegate_task tool including available agent names."""
+        parts: list[str] = []
+        for name in self.agents:
+            desc = self._resolved_descriptions[name]
```
adtyavrdhn, line 216: Gotcha for the agents descriptions, not the cleanest way but I don't have an alt in mind atm

adtyavrdhn:
**Recursion prevention** -- there's nothing stopping a sub-agent from having its own `SubAgent` capability and recursing indefinitely. Hermes does this two ways: a `MAX_DEPTH=2` counter on `_delegate_depth`, and stripping `delegate_task` from child toolsets entirely. We should do at least one of these.

**Concurrency limits** -- `delegate_tasks` does an unbounded `asyncio.gather`. Hermes caps at 3 (`ThreadPoolExecutor(max_workers=3)`), pi caps at 4 concurrent / 8 total. We should add a semaphore or a `max_concurrency` parameter.

**Timeouts / turn limits** -- a hung sub-agent blocks the parent forever. Hermes uses an iteration budget (default 50). Mastra passes `maxSteps` per sub-agent and supports `abortSignal`. We should at minimum pass `usage_limits` through to `agent.run()`.

**Tool filtering** -- hermes intersects child toolsets with the parent's and strips dangerous tools (`delegate_task`, `clarify`, `memory`). We pass the agent as-is with whatever it was constructed with. Worth considering whether we need a `blocked_tools` or privilege-scoping mechanism.

**Result metadata** -- we return a single string. Hermes returns JSON with `status`, `duration_seconds`, `tokens`, `exit_reason`, `tool_trace`. Pi returns `exitCode`, full messages, `usage` stats. Mastra returns `{text, subAgentToolResults[]}`. Richer results would help the parent agent reason about what happened.

**Streaming / progress** -- all three (hermes, pi, mastra) stream intermediate progress from sub-agents. We're opaque until completion. Not a blocker but worth tracking.

PR #163, opened 2026-04-02
Feature: Code Mode Capability with Monty

adtyavrdhn:
## Code Mode capability

Adds a `CodeMode` capability to the harness. When enabled, the model sees one
tool called `run_code` instead of all your tools individually. The model writes
Python that calls your tools as functions, and the code runs in a Monty
sandbox. Useful when a task needs several tool calls plus arithmetic or simple
control flow, since the model can express the whole thing in one snippet
instead of going back and forth.

### How to use it in your Pydantic AI project

All dependencies are available from PyPI. Monty is an optional dependency
installed via the `code-mode` extra:

```toml
[project]
dependencies = [
    "pydantic-ai-slim>=1.78.0",
    "pydantic-harness[code-mode]",
]

[tool.uv.sources]
pydantic-harness = { git = "https://github.com/pydantic/pydantic-harness.git", branch = "code-mode-monty" }
```

Then run `uv lock && uv sync`.

In code:

```python
from pydantic_ai import Agent
from pydantic_harness import CodeMode

agent = Agent(
    capabilities=[CodeMode()],
    instructions='You are a shopping assistant. Use run_code to do calculations.',
)

@agent.tool_plain
def get_price(item: str) -> float:
    """Look up the unit price of an item in dollars."""
    ...

@agent.tool_plain
def apply_discount(amount: float, percent: int) -> float:
    """Apply a percentage discount to a dollar amount."""
    ...

result = await agent.run('What is the price of 3 apples after a 15% discount?')
```

The model only sees the `run_code` tool. Your wrapped tools show up inside
the `run_code` description as a Python prelude the model can call from its
snippet:

```python
async def get_price(*, item: str) -> float:
    """Look up the unit price of an item in dollars."""
    ...

async def apply_discount(*, amount: float, percent: int) -> float:
    """Apply a percentage discount to a dollar amount."""
    ...
```

Tools must be called with `await` and keyword arguments. The `run_code`
description tells the model this explicitly.

### Choosing which tools live inside code mode

By default, every tool the agent has goes into the sandbox. Pass a list of
tool names or a callable to put only some tools inside `run_code` and leave
the rest visible to the model as normal tool calls:

```python
CodeMode(tools=['compute_total', 'apply_discount'])
CodeMode(tools=lambda ctx, tool_def: tool_def.name.startswith('compute_'))
```

### Observability

- The `run_code` tool carries `code_arg_name`/`code_arg_language` metadata
  for Logfire code rendering with syntax highlighting.
- Sandboxed tool calls are dispatched through `ToolManager` for proper OTel
  tracing and capability hooks. Each nested call gets its own
  `execute_tool` span.
- Nested tool call/return parts are stored on the `run_code` `ToolReturnPart`
  metadata as `tool_calls` and `tool_returns` dicts keyed by `tool_call_id`.
- Nested `tool_call_id`s use the format `{parent_id}__{n}` for unique tracing.

### Known limitations

Read these before adopting code mode in a production agent.

**Approval-required tools are not yet supported.** Tools that require user
approval or deferral raise `UserError` when called from inside the sandbox.
The sandbox cannot pause and resume for approval round-trips.

**Tool names that are not valid Python identifiers are automatically
sanitized.** Hyphens, dots, and other non-identifier characters are replaced
with underscores (e.g. `get-weather` becomes `get_weather`). Python keywords
get an underscore appended (e.g. `for` becomes `for_`). If sanitization causes
a collision, the later tool is dropped with a warning.

**The sandbox is a strict subset of Python.** Verified against Monty `main`,
the following are not supported:

* Class definitions of any kind, including `@dataclass`-decorated classes and
  `TypedDict` subclasses
* `with` statements
* Wildcard imports (`from x import *`)
* Third-party libraries

These standard library modules can be imported (you must `import` them
explicitly inside the snippet, just like in regular Python): `sys`, `typing`,
`asyncio`, `math`, `json`, `re`, `datetime`, `os`, `pathlib`. Some submodule
APIs are partial. Errors surface as `ModelRetry`, so the model gets a chance
to fix its own snippet on the next turn.

**The `run_code` tool is marked `sequential=True`.** Multiple `run_code` calls
in a single model response are executed one at a time to prevent concurrent
REPL access.

### Trying it out

There is a runnable demo at `demo_code_mode.py` in the repo root. It sets up
three tools (`get_price`, `get_stock`, `apply_discount`), gives the agent a
shopping question that needs a few tool calls plus arithmetic, and prints
both the Python the model wrote and the final answer.

```bash
uv pip install anthropic logfire   # both optional, only needed for the demo
ANTHROPIC_API_KEY=... LOGFIRE_TOKEN=... uv run python demo_code_mode.py
```

The harness itself does not depend on `anthropic` or `logfire`. Logfire is
configured optionally and falls back cleanly if not installed.


pydantic_harness/toolsets/_code_mode.py
```diff
@@ -0,0 +1,149 @@
...
+            t_name: _make_tool_callable(t_name, t, ctx) for t_name, t in original_tools.items()
+        } or None
+
+        output_chunks: list[str] = []
+
+        def _print_callback(_stream: str, text: str) -> None:
+            output_chunks.append(text)
+
+        try:
+            result = await self._repl.feed_run_async(
```
adtyavrdhn, line 297: How are we now supposed to send the approval results back if monty handles it? Completely forgot about that :/

pydantic_harness/toolsets/code_execution/run_code.py
```diff
@@ -0,0 +1,149 @@
...
+
+# TODO: Add Python function signatures for available external functions to the tool description,
+# so the LLM knows what functions are available and their parameters.
+
+# TODO: Sanitize tool names that aren't valid Python identifiers (e.g. MCP tools with
+# hyphens/dots like `get-weather`, `api.call`) and map them back on dispatch.
+
+
+@dataclass
+class CodeExecutionToolset(WrapperToolset[AgentDepsT]):
```
DouweM: Let's call everything in this branch _just_ CodeMode! Let's not set the expectation that this same toolset will be used for CodeExec at some point.

DouweM (changes requested):

pydantic_harness/capabilities/code_mode.py
```diff
@@ -0,0 +1,62 @@
...
+    `run_code` tool — the model writes Python that calls them as functions instead
+    of issuing tool calls directly.
+
+    Pass a list of tool names or a callable predicate to `tools` to split the
+    toolset: matching tools become callables inside the sandbox, and the rest
+    stay visible to the model as normal tool calls.
+
+    ```python
+    from pydantic_ai import Agent
+    from pydantic_harness.capabilities import CodeMode
```
DouweM: I want the import just to be ` from pydantic_harness import CodeMode`

pydantic_harness/capabilities/code_mode.py
```diff
@@ -0,0 +1,62 @@
...
+
+    # Sandbox only specific tools
+    agent = Agent('openai:gpt-5', capabilities=[CodeMode(tools=['search', 'fetch'])])
+    ```
+    """
+
+    tools: ToolSelector[AgentDepsT] = field(default='all')
+    """Which wrapped tools should be sandboxed inside `run_code`.
+
+    - ``'all'`` (default): every tool the agent has is sandboxed.
```
DouweM: All docstrings are Markdown, no double backticks anywhere

pydantic_harness/capabilities/code_mode.py
```diff
@@ -0,0 +1,62 @@
...
+        selector = self.tools
+
+        async def _sandboxed(ctx: RunContext[AgentDepsT], td: ToolDefinition) -> bool:
+            return await matches_tool_selector(selector, ctx, td)
+
+        async def _native(ctx: RunContext[AgentDepsT], td: ToolDefinition) -> bool:
+            return not await matches_tool_selector(selector, ctx, td)
+
+        sandboxed = FilteredToolset(wrapped=toolset, filter_func=_sandboxed)
+        native = FilteredToolset(wrapped=toolset, filter_func=_native)
```
DouweM: Yeah the double filtering seems problematic; would it run `toolset.get_tools` twice without caching as well? I'd rather handle "which tools do we return from `get_tools` ourselves" unless there's another good route here.

pydantic_harness/toolsets/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""Toolsets shipped by the harness, including the code-mode toolset."""
+
+from .code_mode.run_code import CodeModeToolset
+
+__all__ = ['CodeModeToolset']
```
DouweM, line 5: Do you feel strongly that the toolset should be directly publicly accessible?

pydantic_harness/toolsets/code_mode/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""Code mode toolset: routes LLM-generated Python through a Monty sandbox."""
+
+from .run_code import CodeModeToolset
```
DouweM: Let's inline it

pydantic_harness/toolsets/code_mode/run_code.py
```diff
@@ -0,0 +1,393 @@
...
+        return {
+            _RUN_CODE_TOOL_NAME: ToolsetTool(
+                toolset=self,
+                tool_def=ToolDefinition(
+                    name=_RUN_CODE_TOOL_NAME,
+                    description=description,
+                    parameters_json_schema=_RUN_CODE_JSON_SCHEMA,
+                    metadata={'code_arg_name': 'code', 'code_arg_language': 'python'},
+                ),
+                max_retries=3,
```
DouweM: This should be configurable (as an argument to the CodeModeToolset + CodeMode capability), since syntax errors also count as retries

pydantic_harness/toolsets/code_mode/run_code.py
```diff
@@ -0,0 +1,393 @@
...
+        self, name: str, tool_args: dict[str, Any], ctx: RunContext[AgentDepsT], tool: ToolsetTool[AgentDepsT]
+    ) -> Any:
+        """Execute Python code in the sandbox, dispatching any external function calls to wrapped tools."""
+        code = tool_args['code']
+        restart = tool_args.get('restart', False)
+
+        if self._repl is None or restart:
+            self._repl = MontyRepl()
+
+        wrapped_tools = await self.wrapped.get_tools(ctx)
```
DouweM: I don't love that we're calling `get_tools` here again. Can we store the received tools on the `ToolsetTool`? We can use a subclass like we do in the `ToolSearchToolset`.

pydantic_harness/toolsets/code_mode/run_code.py
```diff
@@ -0,0 +1,393 @@
...
+        callable_defs: dict[str, ToolDefinition] = {}
+        sanitized_to_original: dict[str, str] = {}
+        for name, tool in wrapped_tools.items():
+            td = tool.tool_def
+            if td.defer or td.defer_loading:
+                if name not in self._warned_deferred:
+                    self._warned_deferred.add(name)
+                    warnings.warn(
+                        f'CodeMode does not currently support deferred tool {name!r} '
+                        f'(kind={td.kind!r}, defer_loading={td.defer_loading}); it will '
```
DouweM: Defer(red execution) and defer_loading are very different: defer_loading  relates to the "tool search" feature (look it up). If a user tags a tool for "defer loading = only discoverable via tool search", I agree that for now we should skip them. But I suggest having separate warnings for "deferred execution and human-in-the-loop approval" and "deferred loading (tool search)" (you come up with better phrasing please).

- Also we need to think about the `search_tool` tool that `ToolSearchToolset` injects -- right now I believe it is injected _after_ we inject our `run_code` (meaning that it would remain available as a top-level tool, as it should be!), but we should make sure we _exclude it_ code-mode-ification when `tools='all'` is used, by the time CodeMode becomes a core Pydantic AI feature.  (So add a TODO comment for this!)

pydantic_harness/toolsets/code_mode/run_code.py
```diff
@@ -0,0 +1,393 @@
...
+                        f'collides with {existing!r}; '
+                        f'{name!r} will be hidden from the sandbox.',
+                        UserWarning,
+                        stacklevel=2,
+                    )
+                    continue
+                sanitized_to_original[safe_name] = name
+                td = replace(td, name=safe_name)
+
+            callable_defs[safe_name] = td
```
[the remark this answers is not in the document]
DouweM: Good one. We can look at our existing tool renaming / prefixing / "duplicate" errors.

pydantic_harness/toolsets/code_mode/run_code.py
```diff
@@ -0,0 +1,393 @@
...
+
+_RUN_CODE_BASE_DESCRIPTION = """\
+Write and run Python code in a sandboxed environment.
+
+The sandbox uses Monty, a subset of Python. Key restrictions:
+- **No classes**: class definitions are not supported
+- **No third-party libraries**: only the standard library modules listed below can be used
+- **Importable standard library modules**: `sys`, `typing`, `asyncio`, `math`, `json`, `re`, `datetime`, `os`, `pathlib`. These must be imported at the top of your snippet before use, just like in regular Python. For example: `import asyncio` then `results = await asyncio.gather(tool_one(...), tool_two(...))`.
+- **No `import *`**: wildcard imports are not supported
+- **All tool functions listed below are async and pre-defined in the namespace**: invoke them with `await` directly (no import needed), e.g. `result = await tool_name(arg=value)`. Calling without `await` returns an unresolved future, not the value.
```
DouweM: This should be in `_FUNCTIONS_HEADER` because there may not be tool functions listed below if there are no tool defs.

pydantic_harness/toolsets/code_mode/run_code.py
```diff
@@ -0,0 +1,393 @@
...
+        # toolset is itself a wrapper (e.g. `CombinedToolset`), `tool.toolset` is the
+        # innermost owning toolset and will assert-fail on the wrapper-typed `tool`.
+        #
+        # For sanitized names (e.g. `get_weather` from `get-weather`), Monty sees the
+        # safe name but we dispatch using the original name from the wrapped toolset.
+        external_functions = {
+            safe_name: _make_async_tool_wrapper(
+                sanitized_to_original.get(safe_name, safe_name),
+                self.wrapped,
+                wrapped_tools[sanitized_to_original.get(safe_name, safe_name)],
```
DouweM: Move `sanitized_to_original.get(safe_name, safe_name)` into a var please

pydantic_harness/toolsets/code_mode/run_code.py
```diff
@@ -0,0 +1,393 @@
...
+
+        # `function_signature` is guaranteed non-None for non-output tools by
+        # `ToolDefinition.__post_init__`, and output tools are filtered out by the
+        # `get_wrapper_toolset` contract before they reach us — see `_partition_callable_tools`.
+        sigs = [cast(FunctionSignature, td.function_signature) for td in callable_defs.values()]
+
+        # `dedup_referenced_types` mutates the signature objects in place. The signature
+        # objects are owned by the (cached) `ToolDefinition.function_signature` field, so
+        # successive calls to `_build_description` see types already deduped — that's
+        # idempotent and safe.
```
DouweM: I believe that in the latest version of `FunctionSignature`, this is no longer the case, and we should be passing dupe type names into the render method. Check it.

pydantic_harness/toolsets/code_mode/run_code.py
```diff
@@ -0,0 +1,393 @@
...
+    which we need for `needs_prefix=True` types — those resolve their final class
+    name (e.g. `get_company_Address`) from the `_render_tool_name` ContextVar at
+    render time. Walking the per-tool list ourselves lets us:
+
+    1. Use the upstream helper for dedup-by-id (single source of truth).
+    2. Build an owner map so each prefixed type renders with the correct prefix.
+
+    A "render the whole prelude" helper would fit naturally on `FunctionSignature`
+    upstream — worth pushing back into PR #4964 as a follow-up — but for now this
+    keeps the demo unblocked without depending on more API surface from pydantic-ai.
```
DouweM: Eeeeh I'm working on both PRs at the same time, so it's very weird to be doing this here, if we need changes in FunctionSignature, now is the time to make them in the other repo/branch.

pydantic_harness/toolsets/code_mode/run_code.py
```diff
@@ -0,0 +1,393 @@
...
+    We dispatch through `outer_toolset.call_tool(...)` rather than
+    `outer_tool.toolset.call_tool(...)` because `outer_tool.toolset` may be the
+    innermost owning toolset, while `outer_tool` itself is a wrapper-typed tool
+    minted by an outer wrapper (e.g. `_CombinedToolsetTool` from `CombinedToolset`).
+    Calling the inner toolset with a wrapper-typed tool trips an isinstance assert
+    inside upstream toolset implementations like `_AgentFunctionToolset.call_tool`.
+    """
+
+    async def wrapper(**kwargs: Any) -> Any:
+        return await outer_toolset.call_tool(tool_name, kwargs, ctx, outer_tool)
```
DouweM: I believe these are not traced / capability-hooked correctly. We need to go through `ToolManager` I believe, and make sure we have tests that verify instrumentation (like we have instrumentation / logfire tests for regular tool calls)

pydantic_harness/environments/__init__.py
DouweM, line 1: Can be removed

pydantic_harness/__init__.py
```diff
@@ -0,0 +1,3 @@
+"""Agent harness for composable, reusable AI agent capabilities, built on Pydantic AI."""
```
DouweM: Not "on": they're capabilities _for_ Pydantic AI.

pyproject.toml
```diff
@@ -24,7 +24,11 @@ classifiers = [
     'Topic :: Software Development :: Libraries',
     'Typing :: Typed',
 ]
-dependencies = ['pydantic-ai-slim>=1.76.0']
+dependencies = [
+    "logfire[httpx]>=4.31.0",
```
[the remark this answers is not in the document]
DouweM: Should be a dev dependency

DouweM (changes requested):

pydantic_harness/toolsets/_code_mode.py
DouweM: Yeah I think this Monty function may need some tweaking

pydantic_harness/toolsets/_code_mode.py
```diff
@@ -306,14 +311,30 @@ async def dispatch_tool_call(original_name: str, kwargs: dict[str, Any]) -> Any:
...
-        response: dict[str, Any] = {}
         printed = capture.joined
-        if printed:
-            response['output'] = printed
+
+        # Validate result to reconstruct multimodal types (e.g. BinaryContent from
+        # serialized dicts) so they flow through to the model natively.
         if result is not None:
-            response['result'] = result
+            result = _TOOL_RETURN_CONTENT_VALIDATOR.validate_python(result)
```
DouweM: Are we handling `ToolReturn` returned by wrapped tools correctly? We can ignore the user parts stuff, but I'd expect the `ToolReturnPart` with `metadata` to still show up on our own `ToolReturnPart.metadata['tool_returns'][...].metadata`.

Is there logic in Pydantic AI that is currently outside of the ToolManager (i.e in the agent graph), but could potentially be used here / made resuable?

pydantic_harness/toolsets/_code_mode.py
```diff
@@ -42,6 +42,9 @@ class _RunCodeArguments(TypedDict):
 _RUN_CODE_JSON_SCHEMA = _RUN_CODE_ADAPTER.json_schema()
 _RUN_CODE_ARGS_VALIDATOR = _RUN_CODE_ADAPTER.validator
 _TOOL_RETURN_ADAPTER: TypeAdapter[Any] = TypeAdapter(Any)
+# Validates Monty results back to proper types (e.g. reconstructs BinaryContent from serialized dicts)
+# so that multimodal content flows through to the model natively.
+_TOOL_RETURN_CONTENT_VALIDATOR: TypeAdapter[Any] = TypeAdapter(ToolReturnContent)
```
DouweM: Do we need both _TOOL_RETURN_CONTENT_VALIDATOR and _TOOL_RETURN_ADAPTER?

README.md
```diff
@@ -55,6 +55,99 @@ result = agent.run_sync('Remember that my favourite colour is blue.')
 | ToolOrphanRepair | Repair orphaned tool calls in conversation history |
 | ToolOutputManagement | Control and format tool output for the model |
 
+## Code Mode
+
+The `CodeMode` capability replaces direct tool calls with a single `run_code` tool. Instead of calling tools one at a time, the model writes Python code that calls them as async functions inside a sandboxed [Monty](https://github.com/pydantic/monty) runtime. This lets the model chain multiple tool calls, use control flow (loops, conditionals), and post-process results -- all in a single round-trip.
+
+See also: [Tool use is also code generation](https://www.anthropic.com/engineering/tool-use-is-also-code-generation) (Anthropic) and [How we built our AI agent's tool use pipeline](https://blog.cloudflare.com/how-we-built-our-ai-agents-tool-use-pipeline/) (Cloudflare).
```
DouweM: I think the docs on https://github.com/pydantic/pydantic-ai/pull/4153 have better links to Anthropic and Cloudflare docs/articles. And possibly better explanation of the purpose and advantages of code mode, like saving context and parallelism.

README.md
```diff
@@ -55,6 +55,99 @@ result = agent.run_sync('Remember that my favourite colour is blue.')
...
+
+### Basic usage
+
+```python
+from pydantic_ai import Agent
+from pydantic_harness import CodeMode
+
+agent = Agent('anthropic:claude-sonnet-4-20250514', capabilities=[CodeMode()])
```
DouweM: Use sonnet 4.6

README.md
```diff
@@ -55,6 +55,99 @@ result = agent.run_sync('Remember that my favourite colour is blue.')
...
+CodeMode(tools=['search', 'fetch'])
+
+# By predicate
+CodeMode(tools=lambda ctx, td: td.name != 'dangerous_tool')
+
+# By metadata -- use with SetToolMetadata capability or .with_metadata() on toolsets
+CodeMode(tools={'code_mode': True})
+```
+
+When using the metadata dict selector, mark tools for sandboxing with the `SetToolMetadata` capability or the `.with_metadata()` method on toolset instances:
```
DouweM: Show `with_metadata` used on a `FunctionToolset` in the `toolsets` list passed to `Agents`

DouweM (changes requested):

PR #166, opened 2026-04-09
CodeMode: Snapshot-based execution loop for Temporal/DBOS compatibility

DouweM:
## Summary

- Replace Monty's `feed_run_async` (which uses background threads + `call_soon_threadsafe`) with the synchronous `feed_start`/`resume` snapshot API, making CodeMode compatible with Temporal and DBOS durable execution engines
- Add sequential tool resolution at FutureSnapshot time via `ToolManager.get_parallel_execution_mode`, matching the agent graph's behavior for DBOS's `parallel_ordered_events` mode
- Add static type checking on first code snippet (or after restart) using Monty's stateless type checker — catches type errors and undefined names before runtime
- Add Temporal integration test (real dev server) and DBOS integration test (SQLite)
- Add CI matrix: slim (code-mode only) vs all-extras (+ temporal + dbos)
- Add `pydantic-harness[temporal]` and `pydantic-harness[dbos]` optional extras

### Why

Monty's `feed_run_async` spawns a background thread and uses `call_soon_threadsafe` to dispatch async callbacks. Temporal's workflow sandbox disables threads and doesn't implement `call_soon_threadsafe`, causing CodeMode to hang. DBOS has similar restrictions. The snapshot approach drives Monty step-by-step on the main event loop, avoiding both.

### Other improvements from the audit

- Exceptions from tool calls passed back into Monty via `ExternalException`, preserving line numbers and context
- Print output before errors preserved with `[stdout before error]` markers
- Positional args explicitly rejected with `TypeError`

## Test plan

- [x] 69 tests pass, 100% coverage
- [x] Temporal integration test runs CodeMode inside a real Temporal workflow
- [x] DBOS integration test runs CodeMode inside a DBOS durable workflow
- [x] CI matrix tests slim (temporal/dbos tests skip) and all-extras (everything runs)
- [x] Static type checking catches bad arg types on first snippet
- [x] Sequential resolution verified with `ToolManager.parallel_execution_mode('sequential')`

### DIFF
--- a/pydantic_harness/code_mode/_toolset.py
+++ b/pydantic_harness/code_mode/_toolset.py
@@ -1,19 +1,4 @@
-"""Code mode toolset that runs LLM-generated Python in a Monty sandbox.
-
-Instead of using Monty's convenience `feed_run_async` method (which spawns a
-background thread and uses `call_soon_threadsafe` to dispatch async callbacks),
-this module drives the Monty REPL via the synchronous `feed_start`/`resume`
-snapshot API. This approach:
-
-- **Works with Temporal**: Temporal's workflow sandbox disables threads and
-  doesn't implement `call_soon_threadsafe`, so `feed_run_async` hangs.
-  The snapshot approach avoids both.
-- **Enables parallel tool calls**: `asyncio.gather`-style concurrency from
-  sandbox code is supported via `FutureSnapshot`.
-- **Improves error propagation**: exceptions from tool calls are passed back
-  into the sandbox via `ExternalException`, giving Monty full context for
-  error messages.
-"""
+"""Code mode toolset that runs LLM-generated Python in a Monty sandbox."""
 
 from __future__ import annotations
 
@@ -30,6 +15,7 @@ from pydantic_ai import AbstractToolset, RunContext, ToolDefinition, WrapperTool
 from pydantic_ai.exceptions import ApprovalRequired, CallDeferred, ModelRetry, UserError
 from pydantic_ai.function_signature import FunctionSignature
 from pydantic_ai.messages import ToolCallPart, ToolReturn, ToolReturnContent, ToolReturnPart, is_multi_modal_content
+from pydantic_ai.tool_manager import ToolManager
 from pydantic_ai.tools import AgentDepsT, ToolSelector, matches_tool_selector
 from pydantic_ai.toolsets.abstract import SchemaValidatorProt, ToolsetTool
 
@@ -101,13 +87,27 @@ Returns the last expression's value directly. If `print()` was also called, retu
 `{"output": "<printed text>", "result": <last expression>}`.\
 """
 
-_FUNCTIONS_HEADER = """\
 
-The following functions are available inside the sandbox. Call them directly \
-(do **not** redefine or import them) and `await` the result. All parameters are keyword-only. \
-All tool functions are async: invoke them with `await`, e.g. `result = await tool_name(arg=value)`. \
-Calling without `await` returns an unresolved future, not the value.\
-"""
+def _functions_header(*, has_sync: bool, has_async: bool) -> str:
+    """Build the functions-header paragraph for the `run_code` tool description."""
+    base = (
+        '\nThe following functions are available inside the sandbox. Call them directly '
+        '(do **not** redefine or import them). All parameters are keyword-only.'
+    )
+    if has_async and not has_sync:
+        return base + (
+            ' All tool functions are async: invoke them with `await`,'
+            ' e.g. `result = await tool_name(arg=value)`.'
+            ' Calling without `await` returns an unresolved future, not the value.'
+        )
+    if has_sync and not has_async:
+        return base + (' All tool functions are synchronous: call them directly, e.g. `result = tool_name(arg=value)`.')
+    return base + (
+        ' Async functions (`async def`) must be invoked with `await`,'
+        ' e.g. `result = await tool_name(arg=value)`.'
+        ' Sync functions (`def`) are called directly, e.g. `result = tool_name(arg=value)`.'
+    )
+
 
 _SEARCH_TOOLS_MODIFIER = (
     ' Note: discovered tools become callable as functions inside the run_code sandbox in subsequent invocations.'
@@ -288,8 +288,7 @@ class CodeModeToolset(WrapperToolset[AgentDepsT]):
         # tool calls go through the standard validation/execution path. We
         # inherit `root_capability` from the agent's ToolManager (for capability
         # hooks) but use the *wrapped* toolset and its tools.
-        from pydantic_ai.tool_manager import ToolManager
-
+        # See https://github.com/pydantic/pydantic-ai/pull/4307
         parent_tm = ctx.tool_manager
         assert parent_tm is not None, 'CodeModeToolset requires ctx.tool_manager to be set'
         tool_manager = ToolManager(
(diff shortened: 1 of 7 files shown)


pydantic_harness/code_mode/_toolset.py
```diff
@@ -447,6 +490,174 @@ def _build_description(self, callable_defs: dict[str, ToolDefinition]) -> str:
...
+        conflicting = FunctionSignature.get_conflicting_type_names(sigs)
+
+        parts = ['import asyncio\nfrom typing import Any, TypedDict, NotRequired, Literal']
+        type_blocks = FunctionSignature.render_type_definitions(sigs, conflicting)
+        parts.extend(type_blocks)
+        parts.extend(
+            td.render_signature('raise NotImplementedError()', is_async=True, conflicting_type_names=conflicting)
+            for td in callable_defs.values()
+        )
+        return '\n\n'.join(parts)
```
DouweM, line 552: There's way too much duplication here with the method above, clean it up.

pydantic_harness/code_mode/_toolset.py
```diff
@@ -1,11 +1,27 @@
...
+snapshot API. This approach:
+
+- **Works with Temporal**: Temporal's workflow sandbox disables threads and
+  doesn't implement `call_soon_threadsafe`, so `feed_run_async` hangs.
+  The snapshot approach avoids both.
+- **Enables parallel tool calls**: `asyncio.gather`-style concurrency from
+  sandbox code is supported via `FutureSnapshot`.
+- **Improves error propagation**: exceptions from tool calls are passed back
+  into the sandbox via `ExternalException`, giving Monty full context for
+  error messages.
```
DouweM: This whole justification doesn't need to be all the way up here, just in the method in question is enough.

pydantic_harness/code_mode/_toolset.py
```diff
@@ -447,6 +490,174 @@ def _build_description(self, callable_defs: dict[str, ToolDefinition]) -> str:
...
+                    else:
+                        # Eagerly schedule as a Task for concurrent execution.
+                        pending[monty_state.call_id] = asyncio.ensure_future(coro)
+                    monty_state = monty_state.resume(future=...)
+                else:
+                    # Unknown function — resume with NameError so the sandbox
+                    # sees a clear error at the call site.
+                    monty_state = monty_state.resume(exception=NameError(f'Unknown function: {fn_name}'))
+            else:
+                # FutureSnapshot — Monty is awaiting one or more deferred futures.
```
DouweM: Add an explicit elif instance here, in case new types are added in the future, and an else with assert_never

pydantic_harness/code_mode/_toolset.py
DouweM, line 1: *(Refers to line 254)*

Import this at the top

pydantic_harness/code_mode/_toolset.py
```diff
@@ -234,14 +262,28 @@ async def call_tool(
             tools=tool.wrapped_tools,
         )
 
+        # Determine whether sandbox tool calls should be resolved sequentially.
+        # We ask the inner ToolManager which checks both the global parallel
+        # execution mode context var (set by durable execution engines like
+        # DBOS) and per-tool sequential flags. We use original (unsanitized)
+        # tool names so ToolManager can look up their definitions.
+        probe_calls = [ToolCallPart(tool_name=n, args={}) for n in tool.wrapped_tools]
+        sequential = tool_manager.get_parallel_execution_mode(probe_calls) != 'parallel'
```
DouweM: Hmm, but this means that if _any_ tool in the entire batch is sequential, we jump into this mode?  Even if it's not required by e.g. DBOS? I would want it to work like it does on agent_graph process_tool_calls, where if we're not in "all sequential" mode, we keep going in parallel _until_ we hit a tool call that's marked as "sequential", and _then_ we switch into a mode where we wait fora ll prior calls to succeed, then run that sequential tool, and then continue, with either more async/parallel calls or more sequential calls -- depending on what we're getting from the Monty env.

It may be helpful to have sequential tools render as `def foo` instead of `async def foo`, if that helps get the right behavior from Monty/LLM.

In that case I suppose it makes sense to go into "all-sync" mode with sequential tool calling mode as well.

pydantic_harness/code_mode/_toolset.py
```diff
@@ -252,12 +294,11 @@ async def dispatch_tool_call(original_name: str, kwargs: dict[str, Any]) -> Any:
...
                 result = await tool_manager.handle_call(call_part, wrap_validation_errors=False)
             except (CallDeferred, ApprovalRequired) as e:
-                # Approval/deferral require a round-trip back to the caller, which
-                # the sandbox cannot do. We raise UserError here; because this runs
-                # inside Monty's external-function callback, Monty catches it as a
-                # RuntimeError and wraps it in MontyRuntimeError, which our caller
-                # then translates to ModelRetry. The error message is preserved
-                # through the chain so the model (or developer) sees the cause.
+                # Approval/deferral require a round-trip back to the caller,
+                # which the sandbox cannot do. Raise UserError so the execution
```
DouweM, line 337: Link to https://github.com/pydantic/pydantic-ai/pull/4307

pydantic_harness/code_mode/_toolset.py
```diff
@@ -281,41 +322,43 @@ async def dispatch_tool_call(original_name: str, kwargs: dict[str, Any]) -> Any:
...
         capture = _PrintCapture()
 
         try:
-            result = await self._repl.feed_run_async(
-                code,
-                external_functions=external_functions,
-                print_callback=capture,
+            monty_state = self._repl.feed_start(code, print_callback=capture)
+            completed = await _execution_loop(
+                monty_state, dispatch_tool_call, callable_defs, sanitized_to_original, sequential
```
DouweM: Require kwargs on the method for all but the "obvious" first arg

tests/_code_mode/test_temporal.py
```diff
@@ -0,0 +1,155 @@
...
+        workflows=[CodeModeWorkflow],
+        plugins=[AgentPlugin(temporal_code_mode_agent)],
+    ):
+        output = await client.execute_workflow(
+            CodeModeWorkflow.run,
+            args=['Calculate 3 + 4'],
+            id='test_code_mode_temporal_1',
+            task_queue=TASK_QUEUE,
+        )
+        assert '7' in output
```
DouweM: Assert against a snapshot of the entire message history (assuming we use inline-snapshot already), I want to see the ToolReturnPart metadata etc.

tests/_code_mode/test_dbos.py
```diff
@@ -0,0 +1,101 @@
...
+# Tests
+# ---------------------------------------------------------------------------
+
+
+def test_code_mode_runs_in_dbos_workflow(dbos_instance: DBOS) -> None:
+    """CodeMode's snapshot-based execution loop works inside a DBOS durable
+    workflow. DBOS defaults to `parallel_ordered_events` mode, which triggers
+    the sequential FutureSnapshot resolution path."""
+    result = dbos_code_mode_agent.run_sync('Calculate 3 + 4')
+    assert '7' in str(result.output)
```
DouweM: Same here as up, I want to see the full history

pyproject.toml
```diff
@@ -32,6 +32,12 @@ dependencies = [
 code-mode = [
     'pydantic-monty>=0.0.10',
 ]
+temporal = [
+    'pydantic-ai-slim[temporal]',
+]
+dbos = [
+    'pydantic-ai-slim[dbos,mcp]',
```
DouweM: This seems weird, why require mcp?

PR #159, opened 2026-04-02
Add KnowsCurrentTime capability

DouweM:
## Summary

- Adds `KnowsCurrentTime`, a minimal `AbstractCapability` subclass that injects the current date/time into the system prompt via `get_instructions()`
- Configurable `tz` (IANA timezone, default UTC), `format` (strftime string), and `include_tool` (optional `get_current_time` tool)
- Exported from `pydantic_harness` package: `from pydantic_harness import KnowsCurrentTime`
- Adds `tzdata` dependency for cross-platform timezone support
- Points `pydantic-ai-slim` to git main branch (capabilities API not yet released to PyPI)

Closes #23

## Test plan

- [x] 16 tests covering all code paths (helpers, defaults, instructions, toolset, serialization, from_spec)
- [x] 100% code coverage
- [x] ruff lint + format clean
- [x] pyright strict mode clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #170, opened 2026-04-10
Add KnowsCurrentTime capability

### LINKED ISSUE #23: KnowsCurrentTime capability

## Goal

Provide a capability that dynamically injects the current date/time into the system prompt.

## Context

- **Serializability:** Tier S — `- KnowsCurrentTime` or `- KnowsCurrentTime: {tz: UTC}`
- **Capability interface methods:** `get_instructions()` only
- **Implementation status:** Not started
- **Blocked by:** Nothing — implementable with current interface

## Description

The simplest possible capability — `get_instructions()` returns a formatted datetime string. Config: `timezone`, `format`. Proposed in the [traits research report](https://github.com/pydantic/pydantic-ai/pull/4233) as `KnowsCurrentTime`.

## Depends on

- #10 — AbstractCapability + CombinedCapability + Agent integration

## Related pydantic-ai issues

DouweM:
## Summary

- Adds `KnowsCurrentTime`, a minimal `AbstractCapability` subclass that injects the current date/time into the system prompt via `get_instructions()`
- Configurable `tz` (IANA timezone, default UTC), `format` (strftime string), and `include_tool` (optional `get_current_time` tool)
- Exported from `pydantic_harness` package: `from pydantic_harness import KnowsCurrentTime`
- Adds `tzdata` dependency for cross-platform timezone support
- Points `pydantic-ai-slim` to git main branch (capabilities API not yet released to PyPI)

Closes #23

## Test plan

- [x] 16 tests covering all code paths (helpers, defaults, instructions, toolset, serialization, from_spec)
- [x] 100% code coverage
- [x] ruff lint + format clean
- [x] pyright strict mode clean

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -1,0 +1,25 @@
+# KnowsCurrentTime capability
+
+Closes #23.
+
+## Overview
+
+A minimal `AbstractCapability` subclass that injects the current date/time into the system prompt via `get_instructions()`.
+
+## Design
+
+- **`get_instructions()`** returns a zero-arg callable (a `SystemPromptFunc`) that is evaluated on each model request, producing a string like:
+  `The current date and time is: 2026-04-02T20:30:00Z (Thursday, April 2, 2026)`
+
+- **Configuration:**
+  - `tz: str = 'UTC'` -- IANA timezone name
+  - `format: str = '%Y-%m-%dT%H:%M:%SZ'` -- strftime format string
+  - `include_tool: bool = False` -- optionally registers a `get_current_time` tool via `get_toolset()`
+
+- **Serialization:** inherits default `get_serialization_name()` and `from_spec()` from `AbstractCapability`, so it supports `- KnowsCurrentTime` and `- KnowsCurrentTime: {tz: 'America/New_York'}` in YAML specs.
+
+## Files
+
+- `src/pydantic_harness/knows_current_time.py` -- implementation
+- `src/pydantic_harness/__init__.py` -- re-export
+- `tests/test_knows_current_time.py` -- 16 tests, 100% coverage
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,7 +24,7 @@ classifiers = [
     'Topic :: Software Development :: Libraries',
     'Typing :: Typed',
 ]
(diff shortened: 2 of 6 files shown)


DouweM:
> *Originally posted by @DouweM in [#159 comment](https://github.com/pydantic/pydantic-harness/pull/159#issuecomment-4174801198) (PR closed due to history rewrite)*

## Audit vs prior art: KnowsCurrentTime

**Worth adding now:**
- Timezone validation at init time
- Elapsed time since session start

**Follow-up opportunities:**
- Scheduling hints

lucastamoios:
Thanks for this! Before merging I wanted to flag a prompt-caching interaction and propose a small tweak.

**The issue.** `get_instructions()` returns a closure that calls `datetime.now()` each time. pydantic-ai invokes that closure on every model request, so within one `agent.run()` with N tool turns the time block varies across all N+1 requests.

**What's fine.** Anthropic's backend already separates static vs dynamic instructions and places `anthropic_cache_instructions` breakpoints before the dynamic part (`models/anthropic.py:1539-1575`). The static prefix still hits the cache on that path.

**What breaks.**
- OpenAI implicit prefix cache: any prefix variation misses.
- Anthropic message-level caching (`anthropic_cache_messages`, `anthropic_cache`, downstream `CachePoint`): their breakpoints sit past the dynamic block.
- Across tool-use turns within one run: the model sees a shifting "now" between turns of the same request.

**Proposal: freeze the timestamp per-run.** `AbstractCapability.for_run(ctx)` is called once per run and, if it returns a new instance, `get_instructions()` is re-extracted from it (`agent/__init__.py:1384-1395`). That's a natural place to snapshot:

```python
async def for_run(self, ctx):
    return replace(self, _frozen_now=datetime.now(tz=self._tzinfo))
```

…and have `_formatted_now()` prefer `_frozen_now` when set.

Result: one run = one consistent "now", system prefix is byte-identical across all requests in the run, all three cache paths above hit within the run. Across runs the time changes, but messages change too, so nothing extra is lost.

An opt-out (`freeze: Literal['per_run', 'per_request'] = 'per_run'`) covers cases where ticking time inside a streaming run matters.

Happy to send a follow-up PR if useful.

PR #184, opened 2026-04-10
Add ToolOrphanRepair capability

DouweM:
## Summary

- Implements `ToolOrphanRepair`, a capability that sanitizes message history to fix orphaned tool calls and results before each model request
- Injects synthetic `ToolReturnPart` / `BuiltinToolReturnPart` for calls without matching results, strips `ToolReturnPart` / `RetryPromptPart` whose `tool_call_id` doesn't match any call, and handles trailing responses and empty request edge cases
- Exports as `from pydantic_harness import ToolOrphanRepair`

Refs: pydantic/pydantic-ai#4728

## Test plan

- [x] 25 tests covering all repair scenarios: orphaned calls, orphaned returns, orphaned builtin calls, trailing responses, empty requests, warnings, multi-turn conversations, and no-op passthrough
- [x] pyright strict mode: 0 errors
- [x] ruff lint + format: clean

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -1,0 +1,50 @@
+# ToolOrphanRepair Capability
+
+## Problem
+
+Multi-turn conversations with tools accumulate structurally invalid message history:
+
+- **Orphaned tool calls**: A `ToolCallPart` in a `ModelResponse` whose result was never
+  recorded (streaming timeout, deferred tool dropped). The next `ModelRequest` lacks a
+  matching `ToolReturnPart`.
+- **Orphaned builtin tool calls**: A `BuiltinToolCallPart` without a matching
+  `BuiltinToolReturnPart` in the same `ModelResponse`.
+- **Orphaned tool returns**: A `ToolReturnPart` or `RetryPromptPart` whose
+  `tool_call_id` does not match any call in the preceding `ModelResponse`
+  (frontend-generated IDs, mismatched call IDs from deferred tools).
+
+Providers (especially Anthropic) reject structurally invalid history with 400 errors.
+Once a conversation is poisoned, every subsequent run fails on the same history.
+
+## Solution
+
+A `ToolOrphanRepair` capability that hooks into `before_model_request` to sanitize
+`request_context.messages` before each model call.
+
+### Repair logic (single forward pass)
+
+For each `ModelResponse` paired with the `ModelRequest` that follows it:
+
+1. **Builtin call repair**: Inject synthetic `BuiltinToolReturnPart` for any
+   `BuiltinToolCallPart` without a matching return in the same response.
+2. **Regular call matching**: Collect `tool_call_id` values from `ToolCallPart` parts.
+3. **Orphaned return stripping**: Remove `ToolReturnPart` / `RetryPromptPart` from the
+   request whose `tool_call_id` is not in the call set.
+4. **Orphaned call patching**: Inject synthetic `ToolReturnPart` for call IDs with no
+   matching return or retry in the request.
+5. **Empty request guard**: If stripping leaves only `SystemPromptPart` parts, insert a
+   placeholder `UserPromptPart("Continue.")` to maintain alternation.
+
+For a trailing `ModelResponse` with no following request:
+- If it contains only unmatched tool calls, drop it entirely.
+- If it has other content (text, builtin results), keep it but strip the tool calls.
+
+### Configuration
+
+- `orphan_call_content: str` -- content for synthetic returns (default: `"Tool call was not completed."`)
+- `warn: bool` -- emit a `UserWarning` when repairs are made (default: `True`)
+
+## References
+
+- pydantic-ai #4728: "Built-in HistoryProcessor for orphaned tool call/result repair"
+- pydantic-harness #82: "Tool Output Management"
--- a/src/pydantic_harness/__init__.py
+++ b/src/pydantic_harness/__init__.py
@@ -7,4 +7,8 @@ Usage:
 # Each capability module is imported and re-exported here.
 # Capabilities are listed alphabetically.
 
-__all__: list[str] = []
+from pydantic_harness.tool_orphan_repair import ToolOrphanRepair
+
+__all__: list[str] = [
+    'ToolOrphanRepair',
+]
--- a/src/pydantic_harness/tool_orphan_repair.py
+++ b/src/pydantic_harness/tool_orphan_repair.py
@@ -1,0 +1,274 @@
+"""Capability that sanitizes message history to fix orphaned tool calls and results.
+
+Multi-turn conversations with tools can accumulate structurally invalid message
+history -- tool calls without matching results, or results referencing calls that
+no longer exist.  Providers (especially Anthropic) reject such history with a 400,
+and once a conversation is "poisoned" it stays broken for every subsequent run.
+
+This capability hooks into ``before_model_request`` to repair the history before
+each model call, so conversations self-heal instead of permanently breaking.
+"""
+
+from __future__ import annotations
+
+import warnings
+from dataclasses import dataclass, field, replace
+from typing import TYPE_CHECKING, Any
+
+from pydantic_ai.capabilities.abstract import AbstractCapability
+from pydantic_ai.messages import (
+    BuiltinToolCallPart,
+    BuiltinToolReturnPart,
+    ModelRequest,
+    ModelResponse,
+    RetryPromptPart,
+    SystemPromptPart,
+    ToolCallPart,
+    ToolReturnPart,
+    UserPromptPart,
+)
+
+if TYPE_CHECKING:
+    from pydantic_ai.messages import ModelMessage, ModelRequestPart, ModelResponsePart
+    from pydantic_ai.models import ModelRequestContext
+    from pydantic_ai.tools import RunContext
+
+
+_ORPHAN_CALL_CONTENT = 'Tool call was not completed.'
+"""Synthetic content injected for orphaned tool calls."""
+
+
+@dataclass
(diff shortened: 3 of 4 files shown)


DouweM:
> *Originally posted by @DouweM in [#132 comment](https://github.com/pydantic/pydantic-harness/pull/132#issuecomment-4174798956) (PR was recreated)*

## Audit vs prior art: ToolOrphanRepair

**Worth adding now:**
- Duplicate tool_call_id deduplication
- Debug-level logging of specific repairs (not just count)

**Follow-up opportunities:**
- Integration with Compaction to catch orphans created by summarization

DouweM:
> *Originally posted by @adtyavrdhn in [#132 comment](https://github.com/pydantic/pydantic-harness/pull/132#issuecomment-4183461422) (PR was recreated)*

## Notes from comparing with Hermes, Pi-mono, and Mastra

Looked at how other frameworks handle the same problem. A few things worth noting:

**Synthetic returns should be marked as errors**
Pi-mono marks injected results with `isError: true` so the model knows the tool didn't actually succeed — it's not a normal result, it's a "this never ran" signal. We can't do this yet because `ToolReturnPart` doesn't have an `is_error` field. That's tracked in pydantic/pydantic-ai#4363. Once that lands, the synthetic returns here should set `is_error=True`.

**Duplicate `tool_call_id` handling**
Already flagged in the audit comment. The current set-based matching means if two calls share an ID (provider bug, frontend-generated IDs), only one synthetic return gets created and the other call stays orphaned. Worth adding detection + a warning at minimum.

**Tool ID sanitization is not this PR's job**
Hermes has an explicit `_sanitize_tool_id()` that replaces non-`[a-zA-Z0-9_-]` chars for Anthropic compliance. That's a provider adapter concern — belongs in pydantic-ai's Anthropic model, not in history repair. Mentioning it here just because it causes the same symptom (Anthropic 400s).

**Errored/aborted turns — framework handles it (mostly)**
Pi-mono skips entire responses with `stopReason === "error"`. Pydantic-ai already handles `finish_reason='length'` (raises `IncompleteToolCall`) and auto-generates tool call IDs if missing, so partial tool calls aren't a concern for this capability. `finish_reason='error'` responses do silently enter history though — that's a framework-level gap, not something for here.

sarth6:
@DouweM I had an agent compare our internal version of this to yours and see what yours is doing better and vice versa, found a few tests that are probably worth adding - some are tests we added by hitting real production bugs (ex. #1)

1. End-to-end smoke through the actual provider that motivated the PR. The suite asserts a lot about the structure of the repaired output, but nothing exercises whether Anthropic actually accepts it. We’ve hit cases where the repaired history was “structurally valid” (every call had a return, every return had a call) and still got 400’d because part ordering inside a single message was wrong — for example, tool_result blocks ending up after a UserPromptPart in the same user turn, when Anthropic requires tool_results at the start. A single VCR cassette round-tripping a repaired payload through the Anthropic adapter would catch this whole class of bug. (Right now _repair_response_request_pair appends synthetic returns to kept_parts — worth checking what part order that produces when the original request already has a UserPromptPart.)

2. Idempotence as an explicit invariant. Re-running the repair on its own output appears to be a no-op given the current algorithm, but it’s a load-bearing property — the moment someone adds part normalization or reordering in a future change, it’s easy to break silently. A repair(repair(x)) == repair(x) test for a few non-trivial shapes locks it.

3. At least one test seeded from a real captured history. Every test case is hand-constructed. In our experience, the most pernicious bugs come from shapes that real producers generate (streaming timeouts dropping builtin parts mid-stream, frontend-generated IDs that don’t match server expectations, deferred-tool callbacks landing in unexpected order). Synthetic tests can be 100% green while a single captured production payload still fails. Even one fixture file from a real broken conversation would meaningfully increase the suite’s defensive value.

4. Multi-segment / stitched histories. When a conversation gets reassembled across boundaries — durable workflow resumption, session import, multi-run checkpoints — a tool call and its return can legitimately live in what looks like a discontinuous structure but is actually valid. Worth a test that confirms the algorithm doesn’t false-positive on a valid pair whose call/return parts originated in different runs but were stitched together correctly. (We have specifically seen the “broken cross-run pair” and “valid cross-run pair” cases produce different correct behaviors and they’re easy to conflate.)

5. Pair-scope choice as a documented test. test_consecutive_model_responses covers two responses in a row, but it doesn’t fully document the algorithm’s narrow (response, request) scope. Worth tests for: three consecutive responses; response → response → request where the first response’s calls would be matched by the eventual request; an interleaving with a system message in between. Your current algorithm drops the first response’s calls in those cases, which may be deliberate — but the test makes the choice explicit and protects it from accidental “fixing” later.

6. Synthetic-part identifiability for downstream consumers. The empty-request placeholder uses UserPromptPart(content="Continue."). Once that lands in request_context.messages, any downstream consumer (chat UI, transcript export, conversation analytics) has no way to tell whether “Continue.” was a real user message or synthesized by the repair. It’s not strictly a test for the repair logic, but it’s worth a thought-experiment test that asserts how a consumer is supposed to identify synthetic parts — otherwise this string will end up surfaced in user-facing rendering somewhere down the line.

agnim25:
Super interested in getting this change through! Let me know if there's anything where I can help. Would generally be interested in this and the other reliability initiatives where we'd want this as a capability. @DouweM @sarth6

dsfaccini:
This comment was posted by Claude Code using claude-opus-4-8 on behalf of David.

Review level: David set the direction and approved not building a harness capability here. The core-code findings below were produced and verified by Claude Code against pydantic-ai 2.14.1; David has not re-checked them line by line.

Hi @DouweM -- Harness Week update. We started a harness backport of this and then found the whole thing has landed in core: pydantic-ai 2.14.1 (the harness's current floor) now runs the full orphan-repair pipeline unconditionally -- `_clean_message_history` in `pydantic_ai/_agent_graph.py` at run start and before every model request (`repair_last_response=True`). That's the landed form of pydantic-ai#4728, the issue this PR named. Point-for-point:

- Dangling call to synthesized return: `_repair_dangling_tool_calls` / `_insert_synthesized_returns`.
- Orphaned return/retry dropped: `_drop_orphaned_tool_results`.
- Your synthetic wording is core's constant verbatim: `The tool call was interrupted before a result was produced.`
- @adtyavrdhn's `is_error` line: superseded by `ToolReturnPart.outcome` (`Literal['success','failed','denied','interrupted']`) -- core sets `outcome='interrupted'` on synthetics. #4363's successor, already shipped.
- Synthetic identifiability (@sarth6 #6): core marks each with `metadata={SYNTHESIZED_TOOL_RETURN_METADATA_KEY: True}`.
- Cross-run / stitched-pair false positives (@sarth6 #4/#5): core's ID-shadowed whole-history walk handles this more robustly than a forward pair-scope.

The one behavior this PR adds that core doesn't -- native/builtin (`NativeTool*`) call repair -- core omits deliberately (native parts are provider-serialized and a native result can arrive in a later response), so porting it to harness would fight a core decision.

So there's no delta left for a harness capability. Recommend closing this as absorbed into core -- but that's your call, not ours; leaving it open. Two things worth salvaging: @sarth6's six production-seeded cases are excellent **core** regression tests if they aren't upstream already, and if surfacing *what core silently repaired* is useful we can do a small observational monitor (in the `CacheStabilityMonitor` shape) instead of a repairer. Credit where due: @DouweM (design), @sarth6 (the six tests), @adtyavrdhn (the outcome/`is_error` call), @agnim25 (impl offer).

DouweM:
Superseded by https://github.com/pydantic/pydantic-ai/pull/6319

@dsfaccini If any of that comment is still relevant in terms of describing next steps / things that didn't make it in that PR, please file an issue/PR against pydantic-ai.

dsfaccini:
This comment was posted by Claude Code using claude-opus-4-8 on behalf of David.

Review level: David directed this and approved the outcome; Claude Code did the gap analysis and verified each claim against the merged pipeline and its tests. David has not re-checked the analysis line by line.

Went through @sarth6's six proposed scenarios against what https://github.com/pydantic/pydantic-ai/pull/6319 actually shipped. Five are already covered or deliberately out of scope:

- Idempotence (2): `test_repair_is_idempotent_and_deterministic` and `test_full_pipeline_idempotent_and_deterministic`.
- Valid stitched cross-run pair not false-broken (4): `test_valid_history_untouched`; the walk matches on `tool_call_id` order and never reads `run_id`/`conversation_id`, so a differing-run variant asserts nothing new.
- Synthetic identifiability (6): core marks synthetics with `metadata={SYNTHESIZED_TOOL_RETURN_METADATA_KEY: True}` + `outcome='interrupted'` (documented in `docs/message-history.md`, asserted in the persisted history and round-tripped through the Vercel adapter). The `UserPromptPart("Continue.")` ambiguity was the harness placeholder; core uses a bare empty `ModelRequest`, so it doesn't arise.
- Real-captured fixture (3): net-neutral, the hand-built cases already reproduce the exact provider-rejection shapes from the cited issues.
- Anthropic within-turn ordering (1): prevented inside a request (`_insert_synthesized_returns` + `_merge_consecutive_messages` front-load tool results, the Anthropic adapter emits in part order) and explicitly excluded across turns per the PR body and `docs/message-history.md`. A cassette can't prove ongoing provider acceptance anyway.

The one genuine gap was the pair-scope case (5): `resp(call_1) -> resp(text) -> req(return call_1)` had no test. The ordered walk treats `call_1` as answered across the intervening response and leaves the cross-turn ordering unrepaired by design, which is the documented boundary but wasn't pinned. Added a regression test for it: https://github.com/pydantic/pydantic-ai/pull/6690

Native/builtin call repair from the closed PR stays out, since core omits it deliberately.

PR #188, opened 2026-04-10
Execution environments

DouweM:


DouweM:
Housekeeping from the capability naming pass (#392/#480): this PR uses the ancient `pydantic_harness/` layout and `CodeMode` shipped on main long since -- recommend closing. Posted by Claude Code, acting on @DouweM's behalf.

PR #211, opened 2026-04-14
[WiP] Approvals for tools from within Code Mode

adtyavrdhn:
## Summary

A couple of people have pointed out the lack of this is concerning so I am inclined to just ship this which doesn't need core changes at all anymore thanks to @DouweM  :)

Follows the same pattern we agreed on before:

1. Tool raises DeferredRequest
2. Extract and show the user the inner tool which asked for approval
3. Approval is still keyed on `tool_call_id` of `run_code`
4. Restore snapshot from the metadata, if the tool name matches then this time we will not raise an `ApprovalRequired`

Not the prettiest but I like it for the less surface area we are dealing with, one thing I don't like currently is the back and forth between approvals. I would prefer to collect all the approvals required and send them in one go as a request but one thing to keep in mind would be that they would still need to be keyed to the same `tool_call_id` that `run_code` has because that is the upper tool which will have the `ctx.tool_call_approved`

## Linked Issue

Fixes #

## Checklist

- [ ] `make lint && make typecheck && make test` passes locally (don't stress about CI -- we'll help)
- [ ] No changes to `pyproject.toml` or `uv.lock` (dependency changes require a separate issue)
- [ ] Docstrings use single backticks (not RST double backticks)


adtyavrdhn:
CC @DouweM I am not a fan of having to do this round after round for tools, this is a UX concern for sure. We should discuss how we want to do HITL. I have some ideas :)

Issue #19, opened 2026-03-20
Expand hooks: `before_tool_call` / `after_tool_call` on AbstractCapability


adtyavrdhn:
> Parent: #1 — [Tracking] Capabilities abstraction for pydantic-ai

## Goal

Add `before_tool_call` and `after_tool_call` hook methods to `AbstractCapability`, enabling capabilities to intercept, validate, modify, or block individual tool calls and their results.

## Context

- **Category:** Hook extension (extends what capabilities can do)
- **Implementation status:** Complete on capabilities branch (PR #4640)

## Description

The capability interface now includes comprehensive tool-level hooks that go beyond the original before/after pattern with a three-tier approach (before/after/wrap) for both validation and execution phases:

**Tool validation hooks:**
- `before_tool_validate(ctx, call, args)` — Modify raw args before validation
- `after_tool_validate(ctx, call, args)` — Modify validated args after validation
- `wrap_tool_validate(ctx, call, args, handler)` — Wraps tool argument validation

**Tool execution hooks:**
- `before_tool_execute(ctx, call, args)` — Modify validated args before execution
- `after_tool_execute(ctx, call, args, result)` — Modify result after execution
- `wrap_tool_execute(ctx, call, args, handler)` — Wraps tool execution (can block by not calling handler)

The `wrap_*` variants provide full control — a guardrail can abort by not calling `handler()`, an approval workflow can pause and resume, and error recovery can catch exceptions.

**Implementation location:** `AbstractCapability` in `pydantic_ai/capabilities/abstract.py` (lines 221-286), hook invocation in `_tool_manager.py`.

## Depends on

- #10 — AbstractCapability + CombinedCapability + Agent integration ✅

## Unlocks

- #28 — Input/Output Guardrails (tool-level guardrails)
- #29 — Approval capability
- #57 — LSP Diagnostics
- #61 — Tool Error Recovery
- #71 — Stuck/Loop Detection
- #79 — Verification Loop
- #83 — System Reminders (partly)

adtyavrdhn:
## Who has this

Tool-level lifecycle hooks (before/after individual tool execution):

| Framework | Equivalent | Details |
|-----------|-----------|---------|
| **Google ADK** | `before_tool_callback` / `after_tool_callback` | Can validate args, skip execution, post-process results. Fixed for Live mode in v1.27.0 (Mar 2026) |
[...]

PR #219, opened 2026-04-24
feat: add input and output guardrails

### LINKED ISSUE #28: Input/Output Guardrails capability

## Goal

Provide capabilities for validating agent inputs and outputs with support for tripwire-style exceptions and configurable actions.

## Context

- **Serializability:** Tier P — `InputGuardrail` / `OutputGuardrail` with user-supplied callables are not fully spec-serializable
- **Current capability surface:** Current `pydantic-ai` already has the lifecycle/tool hooks needed for a first sequential release: `before_run`, `after_run`, `before_model_request`, `prepare_tools`, `before_tool_execute`, and `after_tool_execute`
- **Implementation status:** In progress via #182
- **Remaining blocker:** Advanced parallel / tripwire execution still depends on #144 (`wrap_model_request` cancellation contract)

## Status update (April 17, 2026)

The original blocked-by list on this issue is now stale.

- This issue is **no longer primarily blocked on missing lifecycle hooks** from #19 / #20.
- On **April 13, 2026**, @DouweM commented: "I believe all the requirements for this are in place now."
- The active implementation path is **#182**.
- The only clearly separate core dependency now is **#144**, which is specifically about the OpenAI-style pattern where a guard runs in parallel with the first model request and cancels it if the guard trips.

## Stable first-stage scope

The first stable target for this issue should be **only**:

- `InputGuardrail`
- `OutputGuardrail`

These are the minimal generic input/output validation primitives, and they map cleanly onto the current capability hooks without introducing unstable semantics.

## Explicitly deferred from the first stage

The following should **not** be part of the initial stable core for this issue:

- `ToolGuard`
- `CostGuard`
- `AsyncGuardrail`

Reasoning:

- `ToolGuard` overlaps with existing core toolset/approval primitives (`FilteredToolset`, `PreparedToolset`, `ApprovalRequiredToolset`, deferred-tool approval flow) and should be evaluated separately if a higher-level capability still seems worthwhile.
- `CostGuard` overlaps with existing core `UsageLimits` functionality and needs a clearer differentiated contract before it should ship as a harness capability.
- `AsyncGuardrail` depends on unresolved cancellation / concurrency semantics tracked in #144 and should be treated as a later follow-up.

## Description

Input guardrails validate user input before agent execution. Output guardrails validate the final result before returning to the caller.

A first harness release does **not** need to wait for zero-latency parallel execution. Sequential input/output guardrails can ship on top of the existing capability hooks. Parallel speculative/tripwire execution should be treated as a follow-up mode.

Most requested capability across the ecosystem — [pydantic/pydantic-ai#1197](https://github.com/pydantic/pydantic-ai/issues/1197) has 20+ comments. PR #3938 existed in core, but the longer-term direction is guardrails as capabilities / harness features rather than a standalone core API.

## Shipping scope for v1

- Sequential `InputGuardrail`
- Sequential `OutputGuardrail`
- Exception-based blocking / tripwires
- Tool/cost/async guard variants as separate follow-up work

## Related work

- #182 — current implementation path, but should be narrowed to the stable first-stage core
- #144 — handler cancellation contract for advanced parallel tripwire guardrails
- #47 — concrete safety implementations built on top of the generic guardrail infrastructure
- [pydantic/pydantic-ai#1197](https://github.com/pydantic/pydantic-ai/issues/1197) — Guardrails (20+ comments, active demand)
- [pydantic/pydantic-ai PR #3938](https://github.com/pydantic/pydantic-ai/pull/3938) — older core guardrails PR (won't land as-is)

## References

- [OpenAI Agents SDK: Guardrails](https://openai.github.io/openai-agents-python/guardrails/)
- [Google ADK: Callbacks](https://google.github.io/adk-docs/callbacks/)
- [Mastra: Guardrails](https://mastra.ai/docs/agents/guardrails)
- [Pydantic AI capabilities docs](https://pydantic.dev/docs/ai/core-concepts/capabilities/)

DEENUU1:
## Summary

Adds two guardrail capabilities with a minimal, callable-based API:

- `InputGuard(guard, parallel=False, block_message=...)` — runs before each model request. A guard that returns `False` triggers a graceful refusal via `SkipModelRequest` (a canned `ModelResponse` becomes the step output); a guard that raises propagates as a hard failure.
- `OutputGuard(guard, block_message=...)` — runs in `after_run` against the final agent output. A guard that returns `False` raises `OutputBlocked`.

Both accept sync or async callables. `parallel=True` on `InputGuard` races the guard against the model call via `wrap_model_request` and cancels the handler as soon as the guard trips, saving tokens when the guard is slower than the provider round-trip.

**Design notes:**
- Asymmetric halt semantics are intentional: input fail → graceful (no tokens spent, conversation continues), output fail → exception (tokens already spent, caller must decide).
- `parallel` is a flag on `InputGuard` rather than a separate `AsyncGuardrail` wrapper — keeps the API surface small and matches @Josh-blythe's feedback on #28 (fast guards don't need the cancellation wiring; let the capability declare whether it needs parallel).
- Cancellation on trip is always on in parallel mode; it's a fail-fast primitive and a configurable option was redundant.

**Out of scope for this PR (tracked as follow-ups):**
- Tool guardrails — belong in the tool-approval layer (#19).
- `transform` / `warn` actions — MVP is halt-only.
- Per-step output validation and per-token streaming validation — `after_run` + `wrap_run_event_stream` are the follow-up hooks.
- Content shields (prompt-injection detectors, PII scrubbers, etc.) — stay in `pydantic-ai-shields` on top of these primitives.

**Files added:**
```
pydantic_ai_harness/guardrails/{__init__,_capability,_exceptions}.py
pydantic_ai_harness/guardrails/README.md
tests/_guardrails/{__init__,test_input_guard,test_output_guard}.py
```

Top-level `pydantic_ai_harness/__init__.py` updated to export `InputGuard`, `OutputGuard`, `GuardrailFunc`, `GuardrailError`, `InputBlocked`, `OutputBlocked`.

## Linked Issue

Fixes #28

## Checklist

---

Context: Found during production work at [Vstorm](https://vstorm.co) · [github.com/vstorm-co](https://github.com/vstorm-co).

### DIFF
--- a/pydantic_ai_harness/guardrails/_capability.py
+++ b/pydantic_ai_harness/guardrails/_capability.py
@@ -169,16 +169,14 @@ class InputGuard(AbstractCapability[AgentDepsT]):
                 return_when=asyncio.FIRST_COMPLETED,
             )
             if guard_task in done:
-                guard_exc = guard_task.exception()
-                if guard_exc is not None:
+                if guard_task.exception() is not None:
                     handler_task.cancel()
-                    raise guard_exc
+                    await guard_task  # re-raises the guard's exception
                 return await handler_task
-            # Handler finished first: if it raised, propagate and cancel the guard.
-            handler_exc = handler_task.exception()
-            if handler_exc is not None:
+            # Handler finished first: if it raised, cancel the guard and propagate.
+            if handler_task.exception() is not None:
                 guard_task.cancel()
-                raise handler_exc
+                await handler_task  # re-raises the handler's exception
             # Handler succeeded; still need the guard verdict before committing the response.
             await guard_task
             return handler_task.result()
--- a/pydantic_ai_harness/guardrails/_capability.py
+++ b/pydantic_ai_harness/guardrails/_capability.py
@@ -1,0 +1,215 @@
+"""Input and output guardrail capabilities.
+
+`InputGuard` intercepts each model request and lets a user-supplied callable
+decide whether the current user prompt is safe to send to the model. A guard
+that returns `False` is treated as a graceful refusal: the LLM call is
+skipped via [`SkipModelRequest`][pydantic_ai.exceptions.SkipModelRequest] and
+a canned message becomes the model response for that step. A guard that raises
+propagates the exception so the caller can observe a hard failure.
+
+`OutputGuard` runs once the run completes and validates the final output.
+A guard that returns `False` raises
+[`OutputBlocked`][pydantic_ai_harness.guardrails.OutputBlocked].
+
+Both guards accept sync or async callables.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import inspect
+from collections.abc import Awaitable, Callable
+from dataclasses import dataclass
+from typing import TYPE_CHECKING, Any
+
+from pydantic_ai.capabilities import AbstractCapability, WrapModelRequestHandler
+from pydantic_ai.exceptions import SkipModelRequest
+from pydantic_ai.messages import ModelResponse, TextPart, UserPromptPart
+from pydantic_ai.tools import AgentDepsT, RunContext
+
+from pydantic_ai_harness.guardrails._exceptions import OutputBlocked
+
+if TYPE_CHECKING:
+    from pydantic_ai.models import ModelRequestContext
+    from pydantic_ai.run import AgentRunResult
+
+
+GuardrailFunc = Callable[[str], bool | Awaitable[bool]]
+"""Signature of the callable passed to `InputGuard` / `OutputGuard`.
+
+The callable receives the text to validate and returns `True` when safe.
+It may be sync or async. Raising an exception is treated as a hard failure
+and propagates up to the caller.
+"""
+
+
+async def _evaluate(guard: GuardrailFunc, value: str) -> bool:
+    """Call `guard` and await it if it returned an awaitable."""
+    result = guard(value)
+    if inspect.isawaitable(result):
+        return await result
+    return result
+
+
+def _extract_prompt(ctx: RunContext[AgentDepsT], messages: list[Any]) -> str | None:
+    """Return the text of the most recent user prompt, or `None` if absent.
+
+    Prefers `ctx.prompt` (set at run start) and falls back to scanning the
+    message history for the last [`UserPromptPart`][pydantic_ai.messages.UserPromptPart]
+    so that sub-agent calls or resumed runs without a fresh prompt still work.
+    """
+    if ctx.prompt is not None:
+        return ctx.prompt if isinstance(ctx.prompt, str) else str(ctx.prompt)
+    for message in reversed(messages):
+        parts = getattr(message, 'parts', None)
+        if not parts:
+            continue
+        for part in reversed(parts):
+            if isinstance(part, UserPromptPart):
+                return part.content if isinstance(part.content, str) else str(part.content)
+    return None
+
+
+@dataclass
+class InputGuard(AbstractCapability[AgentDepsT]):
+    """Validate the user prompt before it reaches the model.
+
+    The `guard` callable receives the prompt text and returns `True` when
+    the input is safe. Returning `False` triggers a graceful refusal: the
+    current model request is short-circuited via
+    [`SkipModelRequest`][pydantic_ai.exceptions.SkipModelRequest] with
+    `block_message` as the response text, so the agent returns cleanly to
+    the caller. Raising an exception from the guard propagates it as-is.
+
+    ```python
+    from pydantic_ai import Agent
+    from pydantic_ai_harness import InputGuard
+
+
+    def no_secrets(prompt: str) -> bool:
+        return 'api_key' not in prompt.lower()
+
+
+    agent = Agent('openai:gpt-4.1', capabilities=[InputGuard(guard=no_secrets)])
+    ```
+
+    Set `parallel=True` to start the guard alongside the model call. The
+    handler is cancelled as soon as the guard reports a violation, which saves
+    tokens when the guard is slower than the provider round-trip.
+    """
+
+    guard: GuardrailFunc
+    """Callable that returns `True` when the prompt is safe to send to the model."""
+
+    parallel: bool = False
+    """Run the guard concurrently with the model request and cancel the model call on failure."""
+
+    block_message: str = 'Request blocked by input guardrail.'
+    """Text returned as the model response when the guard trips gracefully."""
+
+    def _blocked_response(self) -> ModelResponse:
+        return ModelResponse(parts=[TextPart(content=self.block_message)])
+
+    async def _run_guard(self, prompt: str) -> None:
+        """Evaluate the guard and raise `SkipModelRequest` on failure."""
+        if not await _evaluate(self.guard, prompt):
+            raise SkipModelRequest(self._blocked_response())
+
+    async def before_model_request(
+        self,
+        ctx: RunContext[AgentDepsT],
+        request_context: ModelRequestContext,
+    ) -> ModelRequestContext:
+        """Check the prompt before the model call in sequential mode."""
+        if self.parallel:
+            return request_context
+        prompt = _extract_prompt(ctx, list(request_context.messages))
+        if prompt is None:
+            return request_context
+        await self._run_guard(prompt)
+        return request_context
+
+    async def wrap_model_request(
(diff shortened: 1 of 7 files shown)


pydantic_ai_harness/guardrails/_capability.py
```diff
@@ -0,0 +1,236 @@
...
+
+    block_message: str = 'Output blocked by output guardrail.'
+    """Message attached to `OutputBlocked` when the guard trips."""
+
+    async def after_run(
+        self,
+        ctx: RunContext[AgentDepsT],
+        *,
+        result: AgentRunResult[Any],
+    ) -> AgentRunResult[Any]:
```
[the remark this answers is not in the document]
DEENUU1:
```
    async def after_run(
        self,
        ctx: RunContext[AgentDepsT],
        *,
        result: AgentRunResult[Any],
    ) -> AgentRunResult[Any]:
```

https://github.com/pydantic/pydantic-ai/blob/2e1d8018f574d7a3c02eb97e3a87582afdd17ef7/pydantic_ai_slim/pydantic_ai/capabilities/abstract.py#L272

pydantic_ai_harness/guardrails/_capability.py
```diff
@@ -0,0 +1,236 @@
...
+            handler_exc = handler_task.exception()
+            if handler_exc is not None:
+                guard_task.cancel()
+                raise handler_exc
+            # Handler succeeded; still need the guard verdict before committing the response.
+            await guard_task
+            return handler_task.result()
+        finally:
+            if not guard_task.done():
+                guard_task.cancel()
```
adtyavrdhn: This could raise while the handler is still running because we aren't waiting for the cancel to go through?

pydantic_ai_harness/guardrails/_capability.py
```diff
@@ -0,0 +1,236 @@
...
+        try:
+            done, _ = await asyncio.wait(
+                [guard_task, handler_task],
+                return_when=asyncio.FIRST_COMPLETED,
+            )
+            if guard_task in done:
+                guard_exc = guard_task.exception()
+                if guard_exc is not None:
+                    handler_task.cancel()
+                    raise guard_exc
```
adtyavrdhn: If there is an exception raised from guard_task we should just await guard_task and it would raise the exception.

pydantic_ai_harness/guardrails/_capability.py
```diff
@@ -0,0 +1,236 @@
...
+            return await handler(request_context)
+        async def run_handler() -> ModelResponse:
+            return await handler(request_context)
+
+        guard_task: asyncio.Task[None] = asyncio.create_task(self._run_guard(prompt))
+        handler_task: asyncio.Task[ModelResponse] = asyncio.create_task(run_handler())
+        try:
+            done, _ = await asyncio.wait(
+                [guard_task, handler_task],
+                return_when=asyncio.FIRST_COMPLETED,
```
adtyavrdhn, line 168: We can simplify it :)
```suggestion

     try:
            done, _ = await asyncio.wait(
                {guard_task, handler_task},
                return_when=asyncio.FIRST_COMPLETED,
            )
        
            if guard_task in done:
                await guard_task
                return await handler_task
        
            response = await handler_task
            await guard_task
            return response
        
        finally:
            for task in (guard_task, handler_task):
                if not task.done():
                    task.cancel()
        
         await asyncio.gather(guard_task, handler_task, return_exceptions=True)                 
```

Sorry for the ugly snippet wrote in GitHub but hope it makes sense

pydantic_ai_harness/guardrails/_capability.py
```diff
@@ -0,0 +1,236 @@
...
+        """Run the guard alongside the model call when `parallel=True`."""
+        if not self.parallel or ctx.run_step > 1:
+            return await handler(request_context)
+        prompt = _extract_prompt(ctx, request_context.messages)
+        if prompt is None:
+            return await handler(request_context)
+        async def run_handler() -> ModelResponse:
+            return await handler(request_context)
+
+        guard_task: asyncio.Task[None] = asyncio.create_task(self._run_guard(prompt))
```
adtyavrdhn, line 163: I am on the fence thinking if we should call it `PromptGuard` because we don't run this for subsequent model requests and only for the first user prompt step, what do you think?

pydantic_ai_harness/guardrails/_capability.py
adtyavrdhn: Getting back to this because I was thinking if we are actually using this in a semantically correct way.

We do `return_when=asyncio.FIRST_COMPLETED` but we don't use it in a way warranting it. We always have to wait for the sibling task to be over, aren't we better off doing a group in that case?

Might lose the exceptions but we can unwrap it I think, worth checking for sure I'd love for it to be as readable as possible.

DEENUU1: Kept wait(FIRST_COMPLETED) though. asyncio.gather(return_exceptions=True) waits for all. 

TaskGroup would be ideal but it's 3.11+. So wait(FIRST_COMPLETED) is the simplest 3.10-compatible form that still gives us fail-fast. The "wait for sibling" only kicks in on the success path.

adtyavrdhn: What about the anyio equivalent? Is that also 3.11+?

adtyavrdhn: Even in cancel paths we would need to cancel the sibling?

If guard fails then handler needs to stop

If handler fails guard has nothing to do

pydantic_ai_harness/guardrails/_capability.py
DEENUU1: To me it looks clear, considering that we also have `OutputBlocked`, and it's similar to what exists in the OpenAI Agents SDK https://github.com/openai/openai-agents-python/blob/9a207b6938699d87d2d17dd67dd628ca3af0232d/src/agents/guardrail.py#L72

pydantic_ai_harness/guardrails/_capability.py
DEENUU1: CC told me that: 

> Tradeoff is it wraps exceptions in `ExceptionGroup`, so a `SkipModelRequest` from the guard comes out as `ExceptionGroup([SkipModelRequest])` which `_agent_graph.py`'s bare except `SkipModelRequest` won't catch. We'd have to manually unwrap before re-raising (no except* on 3.10 either). Net more code than the current wait + finally.


We would have to check it out

adtyavrdhn: Yeah that makes sense, except* would have been grand, the snippet could have collapsed to a fraction.

We should check if we can reasonably unwrap correctly otherwise this is fine with me

dsfaccini:
hey @DEENUU1 sorry for the delay on this one, I'm gonna do another review tomorrow in the morning (EST)

pydantic_ai_harness/guardrails/_capability.py
```diff
@@ -0,0 +1,230 @@
...
+
+    block_message: str = 'Output blocked by output guardrail.'
+    """Message attached to `OutputBlocked` when the guard trips."""
+
+    async def after_run(
+        self,
+        ctx: RunContext[AgentDepsT],
+        *,
+        result: AgentRunResult[Any],
+    ) -> AgentRunResult[Any]:
```
Kludex, line 226: Is the `Any` inside those supposed to change types? If not, then guard can be generic over this `Any`, and `pydantic-ai` can make it actually a TypeVar over there.

Kludex:
Please make sure we document how to use this with Presidio Microsoft, and real case PII Obfuscation; and how to integrate with the common moderation endpoints from Azure and OpenAI.

dsfaccini:
hey @DEENUU1 ! can you "Allow edits by maintainers" please?

dsfaccini:
Claude here (on behalf of @dsfaccini): superseded by #249.

Thanks @DEENUU1 — #249 is this branch plus a `main` merge and review follow-ups (`block_message` callable, `before_model_request`/`wrap_model_request` collapse). It was opened from a branch on this repo only because the `vstorm-co` fork branch wasn't reachable for push; full credit for the implementation stays with you, and your commits carry over. Closing this in favor of #249. Prepackaged LLM guardrails + Presidio/moderation docs are tracked in #248.

PR #240, opened 2026-05-13
fix(code_mode): honor Tool Search's deferred-loading contract

### LINKED ISSUE #232: Code Mode flattens Tool Search deferral — filter `defer_loading=True` from sandboxing
> _Filed by Claude on DouweM's behalf — approved and reviewed by DouweM before opening._

## Summary

`CodeMode` flattens `Tool Search`'s deferred-loading semantics. After [pydantic/pydantic-ai#5143](https://github.com/pydantic/pydantic-ai/pull/5143) lands, every `defer_loading=True` tool surfaced by `ToolSearchToolset` is rendered into `run_code.description` regardless of discovery state, and on native-supporting providers it's *also* on the wire as a top-level deferred tool — duplication, broken progressive disclosure, and a cache bust on every discovery.

## What goes wrong (with PR #5143 merged, harness unchanged)

1. **Progressive disclosure defeated.** `CodeModeToolset.get_tools` sees all 50 tools in `wrapped_tools` regardless of discovery state and renders all 50 signatures into `run_code.description`. The corpus is fully visible from turn 1 — the thing tool search is meant to prevent.
2. **Wire-level duplication on the native path.** `Model.prepare_request` keeps deferred entries on Anthropic / OpenAI (`with_builtin='tool_search'` is supported). They go on the wire as top-level `tools[]` with `defer_loading: true`. Their signatures are *also* in `run_code.description`. Same tool, two wire surfaces.
3. **Cache busts on every discovery (local fallback path).** `search_tools` returns a result; next step `_parse_discovered_tools` flips `defer_loading=True→False` on the discovered tool; code mode rebuilds `run_code.description`. The description string is part of the tool-definitions block in the prompt — Anthropic/OpenAI invalidate cache from the first changed byte.

The existing test [`test_deferred_loading_tools_sandboxed`](https://github.com/pydantic/pydantic-ai-harness/blob/main/tests/_code_mode/test_code_mode.py#L669) codifies the broken behavior — it pre-dates `with_builtin`/`unless_builtin` and asserts the deferred tool is rendered into `run_code` and not exposed natively.

## Tier-1 fix (this issue)

Filter `defer_loading=True` tools out of `CodeModeToolset` sandboxing. They stay native pass-through, so `ToolSearchToolset`'s `with_builtin` / `defer_loading` flags reach `Model.prepare_request` unaltered:

```python
# in CodeModeToolset.get_tools, before the existing partition loop:
for name, tool in wrapped_tools.items():
    if name == _SEARCH_TOOLS_NAME:
        native_tools[name] = tool
    elif tool.tool_def.defer_loading:
        # Stay native + invisible-until-discovered; let ToolSearchToolset's
        # `with_builtin`/`defer_loading` flags reach Model.prepare_request unaltered.
        native_tools[name] = tool
    elif await matches_tool_selector(...):
        sandboxed_tools[name] = tool
    else:
        native_tools[name] = tool
```

Behavior this gives:
- **Native-supported provider:** deferred tools flow to `tools[]` with `defer_loading=true`; provider drives discovery; `search_tools` is dropped by `_resolve_builtin_tool_swap`. After discovery they're `defer_loading=False`, code mode picks them up and renders them inside `run_code` from that point. No double-listing while undiscovered.
- **Local fallback:** deferred tools dropped from the wire by rule 3 while undiscovered (clean), `search_tools` rendered native. On discovery they return as `defer_loading=False`, code mode renders them inside `run_code` next turn.
- **Cache cost:** still pays the bust on the *first* discovery (run_code description grew), but the steady-state cost (no-discovery turns) stays cache-warm. Today's design pays the bust *every* description-changing event with no upside.

`test_deferred_loading_tools_sandboxed` flips its assertions — that's the contract change, not a regression. The current contract pre-dates `with_builtin` and silently discards Tool Search's deferral intent.

## New test required

The existing `TestToolSearchIntegration` (line 1626) only exercises a hand-rolled `_StaticToolset` with `search_tools` — it does **not** exercise `ToolSearchToolset` end-to-end and doesn't catch this. Add a real integration test:

```python
# FunctionToolset + ToolSearchToolset + CodeMode
# Pre-discovery: assert run_code.description does NOT contain the deferred tool's signature.
# Post-discovery: assert it does.
```

## Compatibility

**Forward-compat only.** Behavior change is observable only when a user has BOTH `CodeMode` AND tools with `defer_loading=True` (whether through `ToolSearch` capability auto-injection or manual `Tool(..., defer_loading=True)`). Before this fix, those deferred tools are flattened into `run_code` regardless of discovery state; after this fix, they obey the deferred-loading contract.

- Users not using `CodeMode`: unaffected.
- Users using `CodeMode` without any deferred tools: unaffected.
- Users using `CodeMode` with deferred tools today: see a behavior change. Specifically:
  - Pre-discovery: deferred tool signatures are no longer in `run_code.description`. Tool not callable from inside `run_code` until discovered.
  - Post-discovery: as before, signature appears in `run_code.description` from that step forward.

This is the correct semantics — `defer_loading=True` was a no-op under code mode before and is now respected. We don't expect anyone to be relying on the broken behavior, but the change is observable, so it should land in a minor version bump alongside the pydantic-ai release that ships PR #5143.

## Merge timing

Tied to **pydantic/pydantic-ai#5143**:

- This issue **cannot land before** PR #5143 is merged in pydantic-ai — the new test depends on `ToolSearchToolset` emitting deferred tools end-to-end, which is what #5143 introduces.
- This issue **should not lag** PR #5143 — without the fix, the wire-level duplication and progressive-disclosure breakage above hit any user who uses both capabilities together on the new pydantic-ai version.
- Recommended sequence: merge PR #5143 → bump `pydantic-ai-slim` minimum version in this repo's `pyproject.toml` to the release that includes it → land this fix → release pydantic-ai-harness on the same day as the corresponding pydantic-ai release, with both changelogs cross-referencing each other.

We can't enforce that users bump both packages together (no transitive constraint), but in practice users who pin pydantic-ai-harness will pull a compatible pydantic-ai range, and users who pin both directly are expected to bump both.

## Tier-2 future direction (NOT this issue)

The deeper fix is structural: `run_code.description` becomes static prose, and the function listing moves to instructions (with cache breakpoints between static and dynamic instructions). Newly-discovered tools are announced via `ctx.enqueue(SystemPromptPart(...))` (PR [pydantic/pydantic-ai#4980](https://github.com/pydantic/pydantic-ai/pull/4980), append-only, cache-safe) instead of mutating the cached prefix. This generalizes to the broader dynamic-instructions pattern shared with the lazy-loaded-capabilities work.

Full grapple: [pydantic-ai-notes/features/tool-search/2026-05-07 claude grapple - tool-search x code-mode interaction.md](https://github.com/pydantic/pydantic-ai-notes/blob/main/features/tool-search/2026-05-07%20claude%20grapple%20-%20tool-search%20x%20code-mode%20interaction.md). Cross-link with the prog-disc dynamic-instructions topic: [pydantic-ai-notes/features/progress-disclosure/2026-05-07 topic for next call - dynamic instructions for prog-disc.md](https://github.com/pydantic/pydantic-ai-notes/blob/main/features/progress-disclosure/2026-05-07%20topic%20for%20next%20call%20-%20dynamic%20instructions%20for%20prog-disc.md).

Tier-2 lands after both PR #4980 and the lazy-loaded-capabilities work in pydantic-ai. Tracked here so we don't lose the thread; not blocking this Tier-1 fix.

DouweM:
Closes #232. Depends on **pydantic/pydantic-ai#5143** (native Tool Search), shipped in **pydantic-ai 1.95.0** — the `pydantic-ai-slim` floor is bumped to `>=1.95.0` and the temporary `[tool.uv.sources]` git pin is removed, so this is ready to merge. Recommended: release pydantic-ai-harness with cross-referenced changelog.

## What was wrong

`CodeModeToolset.get_tools` rendered **every** `defer_loading=True` tool's signature into `run_code.description` regardless of discovery state:
- **Progressive disclosure defeated** — the whole deferred corpus is visible from turn 1, the exact thing Tool Search prevents.
- **Wire-level duplication on native-search providers** — Anthropic/OpenAI keep deferred entries on the wire as top-level `tools[]` with `defer_loading: true`; their signatures were *also* in `run_code.description`. Same tool, two surfaces.
- **Cache bust on every discovery** — `run_code.description` is part of the cached tool-definitions prefix; rebuilding it whenever a tool is discovered invalidates the cache from the first changed byte.

## Fix

In `CodeModeToolset.get_tools`, `defer_loading=True` tools stay **native pass-through** (same treatment as the `unless_native` local-fallback tools), so `ToolSearchToolset`'s `defer_loading` / `with_native` flags reach `Model.prepare_request` unaltered. Once a tool is discovered it comes back with `defer_loading=False` and folds into `run_code` from that step on — no double-listing while undiscovered, steady-state (no-discovery) turns stay cache-warm.

Also migrates the sibling filter `tool_def.prefer_builtin` → `tool_def.unless_native` (renamed in pydantic-ai 1.95.0; the deprecated alias would otherwise trip `filterwarnings = ['error']` on every `get_tools` call), and updates the README + `test_readme_quick_start` + the OTel-spans test for pydantic-ai 1.95.0's other deprecations that come along (`builtin=` → `native=` on `MCP`/`WebSearch`, `Agent(instrument=...)` → `capabilities=[Instrumentation(...)]`).

## Behavior change

Observable only when a user has **both** `CodeMode` **and** `defer_loading=True` tools (via the auto-injected `ToolSearch` capability or a manual `Tool(..., defer_loading=True)`):
- Pre-discovery: deferred-tool signatures are no longer in `run_code.description`; the tool isn't callable from inside `run_code` until discovered.
- Post-discovery: as before — the signature appears in `run_code.description` from that step forward.

`defer_loading=True` was effectively a no-op under code mode before; now it's respected. Users without `CodeMode`, or with `CodeMode` but no deferred tools, are unaffected. Should ship in a minor bump.

## Tests

- `test_deferred_loading_tools_sandboxed` → `test_deferred_loading_tools_not_sandboxed` — assertions flipped (the contract change): deferred tool not in `run_code.description`, exposed as a native tool with `defer_loading=True` intact.
- Added `test_deferred_loading_tool_sandboxed_once_discovered` (unit) and end-to-end `FunctionToolset` + `ToolSearchToolset` + `CodeMode` tests for both pre-discovery (deferred tool kept native, out of `run_code`, `search_tools` native) and post-discovery (`ToolSearchReturnPart` in history → tool folds into `run_code`).
- Renamed the three `prefer_builtin` tests to `unless_native`.
- `pydantic_ai_harness/code_mode/README.md`: new "Tool Search" subsection.
- `ruff` format/check, `pyright` strict, `pytest`, 100% branch coverage.

## Not in scope (tracked in #232)

Tier-2 structural fix — `run_code.description` becomes static prose, the function listing moves to dynamic instructions announced via `ctx.enqueue(SystemPromptPart(...))` (append-only, cache-safe). Lands after pydantic/pydantic-ai#4980 and the lazy-loaded-capabilities work.

### DIFF
--- a/.github/workflows/compat-test.yml
+++ b/.github/workflows/compat-test.yml
@@ -32,6 +32,15 @@ on:
 permissions:
   contents: read
 
+# `UV_FROZEN=1` so the lock-resolved sync + the editable-overlay of pydantic-ai-slim /
+# pydantic-graph stays put when the subsequent `uv run ruff/pyright/pytest` calls fire.
+# Without this, `uv run` auto-re-resolves against PyPI ("Ignoring existing lockfile due
+# to change in resolution mode"), silently uninstalls the overlay, and ends up testing
+# harness main against whatever `pydantic-ai-slim` is on PyPI — i.e. the compat check
+# becomes a no-op against the actual pydantic-ai checkout we wanted to validate.
+env:
+  UV_FROZEN: '1'
+
 jobs:
   compat:
     name: harness vs pydantic-ai @ ${{ inputs.pydantic-ai-ref }}
--- a/README.md
+++ b/README.md
@@ -166,11 +166,7 @@ agent = Agent(
         # Wraps every tool into a single run_code, sandboxed by Monty.
         CodeMode(),
 
-        # Progressive tool discovery for large tool sets (e.g. a chatty MCP server):
-        # tools marked `defer_loading=True` stay out of the model's context until it
-        # searches for them -- native provider tool search where available, a local
-        # `search_tools` tool elsewhere. Composes with CodeMode: a discovered tool
-        # folds into `run_code` from that step on.
+        # Progressive tool discovery for large tool sets; discovered tools fold into run_code.
         ToolSearch(),
 
         # --- Reasoning ---
--- a/README.md
+++ b/README.md
@@ -143,7 +143,7 @@ The Quick start above is deliberately small. Here's the other end of the spectru
 ```python
 import logfire
 from pydantic_ai import Agent
-from pydantic_ai.capabilities import MCP, Thinking, WebSearch
+from pydantic_ai.capabilities import MCP, Thinking, ToolSearch, WebSearch
 from pydantic_ai_harness import CodeMode
 
 # Community packages, alphabetical:
@@ -184,6 +184,13 @@ agent = Agent(
 
         # Provider-adaptive web search; falls back to a local DuckDuckGo implementation.
         WebSearch(),
+
+        # Progressive tool discovery for large tool sets (e.g. a chatty MCP server):
+        # tools marked `defer_loading=True` stay out of the model's context until it
+        # searches for them -- native provider tool search where available, a local
+        # `search_tools` tool elsewhere. Composes with CodeMode: a discovered tool
+        # folds into `run_code` from that step on.
+        ToolSearch(),
 
         # Filesystem + shell. By @vstorm-co: https://github.com/vstorm-co/pydantic-ai-backend
         ConsoleCapability(),
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ agent = Agent(
         CodeMode(),
         # Connect to any MCP server -- here, the open-source Hacker News server
         # (https://github.com/cyanheads/hn-mcp-server). native=False forces the
-        # local FastMCP toolset so CodeMode can wrap the tools; without it,
+        # local MCP toolset so CodeMode can wrap the tools; without it,
         # providers that natively support MCP server connectors execute the tools
         # server-side and bypass the sandbox.
         MCP('https://hn.caseyjhand.com/mcp', native=False),
--- a/README.md
+++ b/README.md
@@ -162,9 +162,16 @@ logfire.instrument_pydantic_ai()
 agent = Agent(
     'anthropic:claude-opus-4-7',
     capabilities=[
-        # --- Execution ---
+        # --- Tool execution & discovery ---
         # Wraps every tool into a single run_code, sandboxed by Monty.
         CodeMode(),
+
+        # Progressive tool discovery for large tool sets (e.g. a chatty MCP server):
+        # tools marked `defer_loading=True` stay out of the model's context until it
+        # searches for them -- native provider tool search where available, a local
+        # `search_tools` tool elsewhere. Composes with CodeMode: a discovered tool
+        # folds into `run_code` from that step on.
+        ToolSearch(),
 
         # --- Reasoning ---
         # Provider-adaptive thinking; uses native extended thinking on supporting models.
@@ -184,13 +191,6 @@ agent = Agent(
 
         # Provider-adaptive web search; falls back to a local DuckDuckGo implementation.
         WebSearch(),
-
-        # Progressive tool discovery for large tool sets (e.g. a chatty MCP server):
-        # tools marked `defer_loading=True` stay out of the model's context until it
-        # searches for them -- native provider tool search where available, a local
-        # `search_tools` tool elsewhere. Composes with CodeMode: a discovered tool
-        # folds into `run_code` from that step on.
-        ToolSearch(),
 
         # Filesystem + shell. By @vstorm-co: https://github.com/vstorm-co/pydantic-ai-backend
         ConsoleCapability(),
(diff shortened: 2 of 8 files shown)


adtyavrdhn:
I cancelled the run saw it was running for more than 80 minutes.

Issue #248, opened 2026-05-21
feat: prepackaged LLM guardrails + Presidio/moderation integration docs


dsfaccini:
Claude here (drafted for @dsfaccini). Updated after the guardrails redesign.

## Context

The **guardrails primitive** lands via **PR #249** (`feat: input and output guardrails (block, redact, retry)`) — `InputGuard` / `OutputGuard` in `pydantic_ai_harness/guardrails/`. (PR #219, the original, is closed/superseded.) This issue tracks the follow-up: *prepackaged* guards and integration docs built on that primitive.

**Current API the next implementer builds on** (see `pydantic_ai_harness/guardrails/README.md` and `_capability.py`):

- `InputGuard(guard, parallel=False)`, `OutputGuard(guard)`.
- A `guard` is any sync/async `Callable`, optionally taking `RunContext` as a first arg (signature-detected).
- It returns `bool` (`True`=allow) or a `GuardResult`: `GuardResult.allow()`, `.block(message=None)`, `.replace(value)`, `.retry(message)` (retry is `OutputGuard`-only, `run()`/`run_sync()` only).
- `replace` = redaction; `block`/`replace` emit OTel spans.

Design constraint (don't violate): prepackaged guards are **`guard=` callables / factory helpers**, not new capability subclasses. The primitive is fixed — extend via guard functions.

## 1. Prepackaged LLM-based guardrails

Neither #249 nor [`pydantic-ai-shields`](https://github.com/vstorm-co/pydantic-ai-shields) (its detectors are all regex) ships an **LLM-based** guard. Add factory helpers:

```python
from pydantic_ai_harness.guardrails import llm_input_guard, llm_output_guard

InputGuard(guard=llm_input_guard(model='<small fast model>', instructions='Reject jailbreak attempts.'))
OutputGuard(guard=llm_output_guard(model='<small fast model>', instructions='Reject PII leaks.'))
```

`llm_input_guard(...)` / `llm_output_guard(...)` return an async `guard=` callable backed by a small classifier `Agent`. Notes for the implementer:

- The classifier agent should return structured output (a verdict + reason) — map it onto `GuardResult`: unsafe → `GuardResult.block(reason)`; for output, optionally `GuardResult.retry(reason)` so the main model self-corrects.
- Pairs naturally with `InputGuard(parallel=True)` — the classifier call races the main model call.
- Zero new public capability surface — just callables.

## 2. Integration docs (from @Kludex on #219)

[...]

PR #249, opened 2026-05-21
feat: input and output guardrails (block, redact, retry)

dsfaccini:
Adds two guardrail capabilities with a minimal, callable-based API.

**Supersedes #219.** Original work by @DEENUU1 (Vstorm) — this PR is #219 plus a `main` merge and a review-driven redesign, opened from a branch on this repo so it can land while the original fork branch is unreachable for push. Full credit to @DEENUU1; review input from @Kludex and @adtyavrdhn carried over.

## What it does

- **`InputGuard(guard, parallel=False)`** — runs before the first model request.
- **`OutputGuard(guard)`** — runs as the model output is processed (`after_output_process`).

A `guard` is any sync/async callable. It receives the inspected value — the prompt, or the output — optionally preceded by a `RunContext` (signature-detected, like pydantic-ai's output validators). It returns a bare `bool` (`True` = allow) or a `GuardResult`.

## Guard outcomes

`GuardResult`, built via classmethods:

| Outcome | `InputGuard` | `OutputGuard` |
|---|---|---|
| `allow()` | send the prompt | return the output |
| `block(message=None)` | skip the model call, refusal message becomes the response (`SkipModelRequest`) | raise `OutputBlocked` |
| `replace(value)` | rewrite the prompt sent to the model (redaction) | substitute a sanitized output |
| `retry(message)` | — usage error | send the output back to the model (`ModelRetry`) |

A guard that raises propagates the exception as a hard failure.

## Observability

`replace` and `block` emit spans on the run's OpenTelemetry tracer (`guardrail redacted input`, `guardrail blocked output`, …) with `guardrail.*` attributes, so redactions and refusals are visible in Logfire. Redacted content is attached only when `RunContext.trace_include_content` is set. `retry` needs no special tracing — the retried request appears in the trace on its own.

## Notable design points

- `OutputGuard` uses `after_output_process` (not `after_run`) so it can redact and trigger `ModelRetry`; it runs on the final output only, not streaming partials.
- `InputGuard` `replace` requires sequential mode (a parallel guard races a model call already started with the original prompt); `retry` is rejected for input.
- `parallel=True` trades tokens for latency — sequential never calls the model on a blocked prompt.

## Follow-ups

Prepackaged LLM-based guardrails and Presidio/Azure/OpenAI moderation docs are tracked in #248.

## Checks

`make format && make lint && make typecheck` clean. Guardrails: 49 tests, 100% branch coverage; full suite green.


pydantic_ai_harness/guardrails/_capability.py
```diff
@@ -0,0 +1,409 @@
...
+)
+"""Signature of the callable passed to `OutputGuard`.
+
+The callable receives the agent output unchanged — for typed outputs this is
+the Pydantic model — and returns `True` / `GuardResult`. It may optionally take
+a [`RunContext`][pydantic_ai.tools.RunContext] first, and may be sync or async.
+"""
+
+
+def _takes_ctx(func: Callable[..., object]) -> bool:
```
adtyavrdhn, line 148: We could import this from `pydantic-ai` although private we run the suite so we won't break harness

pydantic_ai_harness/guardrails/_capability.py
```diff
@@ -0,0 +1,409 @@
...
+    def retry(cls, message: str) -> GuardResult:
+        """Send the output back to the model to try again — `OutputGuard` only.
+
+        `message` is the instruction the model sees on the retry.
+        """
+        return cls(action='retry', message=message)
+
+
+GuardOutcome = bool | GuardResult
+"""What a guard callable returns: a bare `bool` (`True` = allow), or a `GuardResult`."""
```
adtyavrdhn, line 120: Public docstrings could take a look, it could be simpler

pydantic_ai_harness/guardrails/_capability.py
```diff
@@ -0,0 +1,409 @@
...
+async def _evaluate(
+    guard: Callable[..., GuardOutcome | Awaitable[GuardOutcome]],
+    ctx: RunContext[AgentDepsT],
+    value: object,
+) -> GuardResult:
+    """Call `guard` (passing `ctx` when declared), await it, and normalize to `GuardResult`."""
+    outcome = guard(ctx, value) if _takes_ctx(guard) else guard(value)
+    if inspect.isawaitable(outcome):
+        outcome = await outcome
+    if isinstance(outcome, GuardResult):
```
adtyavrdhn, line 173: Isn't this then enforced based on the type?

pydantic_ai_harness/guardrails/_capability.py
```diff
@@ -0,0 +1,409 @@
...
+    path. The model call is cancelled the moment the guard reports a
+    violation. Trade-off: sequential mode never calls the model on a blocked
+    prompt, whereas parallel mode has already started it — if the guard trips
+    only after the model has responded, those tokens were still spent. Prefer
+    sequential for fast local guards. `replace` (redaction) is incompatible
+    with `parallel=True`, since the model call has already started with the
+    original prompt.
+
+    Scope: the guard runs exactly once per run — on the first model request —
+    and evaluates the original user prompt. Subsequent model requests in the
```
adtyavrdhn, line 272: Claude has become increasingly verbose. I am sure it could explain the same thing in fewer words. I dunno if they wanna increase their token costs with this ://

pydantic_ai_harness/guardrails/_capability.py
```diff
@@ -0,0 +1,409 @@
...
+                'already started with the original prompt. Use sequential mode for prompt redaction.'
+            )
+        replacement = verdict.replacement
+        if not isinstance(replacement, str):
+            raise UserError('GuardResult.replace() for an input guard must provide replacement prompt text (str).')
+        if not _replace_prompt(request_context.messages, replacement):
+            raise UserError('InputGuard could not find a user prompt to redact in the request.')
+        _trace_redaction(ctx, direction='input', original=prompt, replacement=replacement)
+
+    async def wrap_model_request(
```
adtyavrdhn, line 332: I see something tricky here, I think the position for this capability needs to be innermost so that no user morphed messages can sneak by and we can be confident this runs only after all the other capabilities are done.

`OutputGuard` similarly would need to be outermost but there are outermost capabilities already so it needs to be wrapped by `instrumentation` or something? Look at deferred caps PR(https://github.com/pydantic/pydantic-ai/pull/5230) for reference, ended up doing something similar without breaking instrumentation

pydantic_ai_harness/guardrails/_capability.py
```diff
@@ -0,0 +1,409 @@
...
+        ctx: RunContext[AgentDepsT],
+        *,
+        output_context: OutputContext,
+        output: Any,
+    ) -> Any:
+        """Evaluate the guard against the processed output and act on its verdict."""
+        if ctx.partial_output:
+            return output
+        verdict = await _evaluate(self.guard, ctx, output)
+        if verdict.action == 'allow':
```
adtyavrdhn: Can we use `match` instead?

Also, there are multiple if conditions, but this is a verdict, right? More than one cannot be true.

pydantic_ai_harness/guardrails/_capability.py
```diff
@@ -0,0 +1,409 @@
...
+                part.content = new_content
+                return True
+    return False
+
+
+def _trace_block(ctx: RunContext[AgentDepsT], *, direction: str, message: str) -> None:
+    """Record a zero-duration span marking a guardrail refusal."""
+    ctx.tracer.start_span(
+        f'guardrail blocked {direction}',
+        attributes={'guardrail.direction': direction, 'guardrail.action': 'block', 'guardrail.message': message},
```
adtyavrdhn: Is it all right for us to dump the message as is? Is there a risk that the LLM explains in the message why it blocked it and ends up leaking it anyway?

pydantic_ai_harness/guardrails/_capability.py
```diff
@@ -0,0 +1,409 @@
...
+    from pydantic_ai.models import ModelRequestContext
+    from pydantic_ai.output import OutputContext
+
+
+_DEFAULT_INPUT_BLOCK_MESSAGE = 'Request blocked by input guardrail.'
+_DEFAULT_OUTPUT_BLOCK_MESSAGE = 'Output blocked by output guardrail.'
+_DEFAULT_OUTPUT_RETRY_MESSAGE = 'Output rejected by output guardrail.'
+
+
+@dataclass
```
adtyavrdhn: Let us use frozen and kw_only

pydantic_ai_harness/guardrails/_capability.py
```diff
@@ -0,0 +1,409 @@
...
+
+    action: Literal['allow', 'block', 'replace', 'retry']
+    """What the capability should do with the inspected value."""
+
+    message: str | None = None
+    """For `block`, the refusal text. For `retry`, the instruction sent back to the model."""
+
+    replacement: object | None = None
+    """For `replace`, the value substituted for the inspected one."""
+
```
adtyavrdhn, line 72: We should do a `__post__init` to verify the integrity of this result

pydantic_ai_harness/guardrails/_capability.py
```diff
@@ -0,0 +1,409 @@
...
+                await guard_task
+                return await handler_task
+
+            response = await handler_task
+            await guard_task
+            return response
+        finally:
+            for task in (guard_task, handler_task):
+                if not task.done():
+                    task.cancel()
```
adtyavrdhn, line 373: Worth letting Claude check if this needs a shield, anyio level cancellations from the graph can seep in and throw cancellations here which would crash the cancellation. I am still wrapping my head around it so I am not sure if that is 100% possible

pydantic_ai_harness/guardrails/_capability.py
```diff
@@ -0,0 +1,409 @@
...
+        if verdict.action == 'allow':
+            return
+        if verdict.action == 'retry':
+            raise UserError(
+                'An InputGuard guard cannot return GuardResult.retry() — retry applies to model output only.'
+            )
+        if verdict.action == 'block':
+            message = verdict.message or _DEFAULT_INPUT_BLOCK_MESSAGE
+            _trace_block(ctx, direction='input', message=message)
+            raise SkipModelRequest(ModelResponse(parts=[TextPart(content=message)]))
```
adtyavrdhn: If we do this then the message remains in history and the model can see it in the next turn because we only go through the messages and pick the last prompt for `InputGuard`?

adtyavrdhn (changes requested):
Structurally great, have a few concerns.

pydantic_ai_harness/guardrails/_capability.py
dsfaccini: Claude here: applied. `InputGuard.get_ordering()` → `position='innermost'` so any message-morphing capability runs first and the guard sees the final prompt. `OutputGuard.get_ordering()` → `position='outermost', wrapped_by=[Instrumentation]` so the guard's block/redact spans are always captured by an enclosing `Instrumentation` span regardless of user list order. Tests added (`TestInputGuardOrdering`, `TestOutputGuardOrdering`).

pydantic_ai_harness/guardrails/_capability.py
dsfaccini: Claude here: investigated with a regression test (`test_no_dangling_tasks_on_outer_cancellation`) that cancels the outer task mid-flight and asserts no leftover guard/handler tasks. It passes without `asyncio.shield` — the outer cancel is consumed by the `asyncio.wait` above the `finally`, so `task.cancel()` + `await asyncio.gather(...)` complete cleanly. Adding `shield` would be defensive but could mask a genuine re-cancellation; the test locks the current behavior in.

pydantic_ai_harness/guardrails/_capability.py
dsfaccini: Claude here: applied — `_trace_block` now gates `guardrail.message` behind `RunContext.trace_include_content`, matching how `_trace_redaction` handles content. Default traces stay content-free; debugging can opt in via `include_content`. The user-facing refusal text (the `SkipModelRequest` response / `OutputBlocked` exception) is unchanged.

pydantic_ai_harness/guardrails/_capability.py
dsfaccini: Claude here: pydantic-ai does not expose a shared `takes_ctx` helper — `_output.py::OutputValidator` and `_system_prompt.py` each inline the same `len(inspect.signature(...).parameters) > N` check. Our `_takes_ctx` mirrors that convention exactly. Happy to swap if a public helper lands.

pydantic_ai_harness/guardrails/_capability.py
dsfaccini: Claude here: the type annotation is just a hint — runtime discrimination of the `bool | GuardResult` union needs the `isinstance` check to narrow. Python doesn't enforce union types at runtime.

pydantic_ai_harness/guardrails/_capability.py
dsfaccini: Claude here: this is intended `SkipModelRequest` behavior in pydantic-ai — the canned `ModelResponse` is the assistant's turn, and a multi-turn chat that skips it would be broken. Per the [`SkipModelRequest` docstring](https://github.com/pydantic/pydantic-ai/blob/main/pydantic_ai_slim/pydantic_ai/exceptions.py), the response is preserved in history. Callers that don't want the refusal carried forward can filter `result.all_messages()` or start a fresh agent for the next turn.

pydantic_ai_harness/guardrails/_capability.py
adtyavrdhn: Oh yeah it doesn't we'll need to import from a private module which I am fine with in this case given it is our private module

adtyavrdhn (approved):
Approving post discussion

dsfaccini:
David's AICA here: evaluated a repetition/degeneration **output guard** for this PR (detect degenerate model output — whitespace ratio, repeated char/n-gram runs, low entropy over a long tail — and turn it into a `retry`/`replace`/`block` verdict at response time). Motivating incident: Pydanty dogfooding, glm-5.1 emitted exactly 65,536 output tokens (the provider cap) of repeated whitespace over 9.2 min; that 1.5MB assistant message entered history and the next request died on the 204,800-token context cap (Logfire EU `pydantic-ai`, trace `019eb9902c1f6937bb40162e36a290c5`).

**Recommendation: keep it out of this PR; land it as a follow-up.** Reasoning:

1. **This PR is the guardrail framework, not a guard library.** It ships `InputGuard`/`OutputGuard`/`GuardResult` and the `OutputGuardFunc` signature, with *zero* concrete pre-built guards (PII/secrets appear only as docstring examples). A repetition guard would be the first concrete guard and shift the PR's identity, inviting "why this one and not PII/secrets?" on an already large, unmerged change.

2. **Its detection heuristics carry their own design decisions** — ratio/run-length/entropy thresholds, a min-length gate to avoid false positives on legitimately repetitive output (tables, ASCII art, code, base64), and typed-vs-`str` output handling. That deserves its own focused review rather than riding along here.

3. **The guard alone does not fix the incident's actual harm, and bundling it here would imply it does.** `OutputGuard.after_output_process` runs *after* the model has fully generated (and spent) the degenerate tokens, and none of its verdicts scrub history: `replace` returns a clean value to the caller but leaves the 1.5MB `ModelResponse` part in history; `retry` keeps the bloat *and* re-asks (can loop expensively); `block` ends the run but still leaves history poisoned for a catch-and-continue caller. Repairing the already-in-history message is exactly the scope of #285 (`ClampOversizedMessages`, zero-LLM). The guard is complementary *at-response-time detection*; it delivers full value only alongside #285 — so the two should land and be reviewed independently.

4. **Deferring is free.** A repetition guard is a pure-additive `OutputGuardFunc` factory needing no change to the settled API (allow/block/replace/retry, no `block_message`, `after_output_process`, signature-detected `ctx`). It can ship as its own small PR after both #285 (clamp) and this PR (framework) merge, designed to compose with both — the framework detects/signals, #285 keeps the run alive.

Related-issue search in `pydantic/pydantic-ai`: **no existing issue** reports endless repetition / degenerate generation / runaway-whitespace output (closest is the tangential #5657, mistral silently dropping `frequency_penalty`/`presence_penalty` — the params that *combat* repetition, not a report of this failure). So this is a novel class; worth its own tracking issue in harness for the follow-up guard.

Issue #147, opened 2026-04-02
ModelSettings.cache_control — structured provider-level cache parameters


DouweM:
> Parent: #1 — [Tracking] Capabilities abstraction for pydantic-ai

## Goal

Add structured cache control to `ModelSettings` so that capabilities and users can control provider-level context caching (Anthropic prompt caching, Google cached_content, OpenAI prefix caching hints) without resorting to `extra_headers` or provider-specific settings.

## Why this needs core

Cache control parameters are embedded in the request body at the provider API level — they're not just headers. A capability can't modify how the model class constructs its API request body. The parameters need to be in `ModelSettings` and translated by each model class.

## Related

- pydantic-harness #99: "Cache-Aware Prompt Construction (prefix cache preservation)"
- pydantic-ai has existing `anthropic_cache_control` patterns but no unified cross-provider API

### 1. Add to ModelSettings

```python
class ModelSettings(TypedDict, total=False):
    cache_control: CacheControlSettings | None
```

Where:
```python
class CacheControlSettings(TypedDict, total=False):
    cache_system_prompt: bool          # Cache the system prompt (all providers)
    cache_tools: bool                  # Cache tool definitions (Anthropic)
    cache_breakpoints: list[int] | None # Message indices to set cache breakpoints (Anthropic)
    cached_content_name: str | None     # Google cached content reference
```

### 2. Provider translation

- **Anthropic/Bedrock**: Translate to `cache_control: {"type": "ephemeral"}` on system messages and tool definitions
- **Google**: Translate to `cached_content` parameter on generation request
- **OpenAI**: Prefix caching is automatic, but structured hints could optimize it

### 3. Capability integration

A `CacheAwarePrompt` capability (#99) can use `get_model_settings()` to return cache control settings based on conversation structure, and `before_model_request` to reorder messages for optimal cache hit rates.

## Unblocks

- #99 (Cache-Aware Prompt Construction)
- Provider-adaptive caching strategies as capabilities
- Cost optimization for long conversations

Issue #99, opened 2026-03-26
Cache-Aware Prompt Construction (prefix cache preservation)


adtyavrdhn:
> Parent: #1 — [Tracking] Capabilities abstraction for pydantic-ai

## Goal

Preserve LLM prefix cache hit rates by freezing the system prompt (and other static context) across turns within a session, preventing cache-busting mutations from capabilities that inject dynamic content.

## Context

- **Category:** Primitive (foundational infrastructure)
- **Serializability:** N/A — infrastructure behavior, not a standalone capability
- **Capability interface methods:** Affects how `get_instructions()` and `before_model_request()` outputs are assembled
- **Implementation status:** Not started
- **Blocked by:** Nothing

### The problem

Anthropic and OpenAI offer **prompt caching** — if the prefix of a request matches a previous request, cached KV pairs are reused at ~90% cost reduction. However, capabilities that inject dynamic content into the system prompt (Memory, Skills, Repository Context) can **invalidate the cache every turn** if their output changes.

### How Hermes solves this

Hermes implements a **snapshot memory pattern**: the system prompt is built once at session start and cached in SQLite. Mid-session memory writes update disk immediately but **do not mutate the running prompt** — the frozen snapshot is reused for every subsequent turn. The next session picks up the updated files.

This is combined with an **ephemeral system prompt** mechanism for turn-specific nudges — these are injected as separate messages (not appended to the cached system prompt) so the prefix cache is preserved.

Key implementation details from Hermes:
- System prompt frozen on first turn, stored in session DB
- `ephemeral_system_prompt` parameter for per-turn injections that don't break cache
- Anthropic-specific `cache_control` directives applied via `prompt_caching.py`
- Honcho user context is injected at API-call time using a separate mechanism that doesn't modify the cached prefix

### Why this matters for pydantic-harness

Multiple planned capabilities inject content into the prompt:
- #30 Memory — injects relevant memories via `before_model_request`
- #22 Skills — injects skill instructions via `get_instructions()`
- #64 Repository Context — loads conventions at session start
- #83 System Reminders — injects mid-session guidance

Without cache-aware assembly, composing 3-4 of these capabilities could mean **zero cache hits** across an entire session, multiplying costs 2-10x.

### Proposed design

```python
class PromptAssemblyStrategy(Enum):
    DYNAMIC = "dynamic"      # Rebuild every turn (current implicit behavior)
[...]

Issue #90, opened 2026-03-21
Prompt Templates / Parameterized User Prompts


adtyavrdhn:
## Summary

Reusable, parameterized user prompt templates stored as markdown files that can be invoked as slash commands with argument substitution. These are **user message templates** (what to ask the agent) as opposed to skills/instructions (how the agent should behave).

This is **distinct from**:
- #22 (Skills) — Skills provide *agent instructions* injected into the system prompt, telling the agent how to behave. Prompt templates provide *parameterized user messages*, telling the agent what to do with variable inputs.
- #11 (Instructions capability) — Instructions are system-level. Prompt templates are user-level.
- #5 (TemplateStr) — TemplateStr is the underlying template engine. Prompt templates are the user-facing authoring and invocation system built on top.

## Motivation

Coding agents frequently need to perform the same types of tasks with different inputs — "review this PR", "add tests for this function", "refactor this file to use pattern X". Without prompt templates, users either:
1. Retype the same complex prompts repeatedly
2. Build their own templating on top of the framework
3. Hard-code prompts in application code

Prompt templates make agents scriptable and workflows reusable without requiring code changes.

**pi-mono** implements this as a standalone system alongside (but separate from) skills:

```typescript
// pi-mono: packages/coding-agent/src/core/prompt-templates.ts

// Prompt template files are markdown with optional YAML frontmatter
// Located at: ~/.pi/agent/prompts/ (global) or .pi/prompts/ (project)
// Invoked as: /template-name arg1 arg2 "arg with spaces"

// Argument substitution supports bash-style patterns:
// $1, $2        — positional arguments
// $@ or $ARGUMENTS — all arguments
// ${@:N}        — arguments from Nth onwards
// ${@:N:L}      — L arguments starting from Nth
```

Example prompt template (`fix-bug.md`):
```markdown
---
name: fix-bug
description: "Fix a bug with test verification"
---
Look at the failing test: $1

1. Read the test to understand what it expects
2. Find the relevant source code
3. Fix the bug
4. Run the test again to verify: $2
```

Invoked as: `/fix-bug test_auth.py "pytest test_auth.py -v"`

**Claude Code** has a similar concept via "skills" that can accept parameters and be invoked as slash commands (e.g., `/commit`, `/review`). The skill system blurs the line between instructions and prompts but serves a similar workflow automation purpose.

### Template Definition

```python
@dataclass
class PromptTemplate:
    """A parameterized user prompt template."""
    name: str
[...]

PR #257, opened 2026-05-26
feat(logfire): add ManagedPrompt capability

dmontagu:
Adds `ManagedPrompt`, a capability that backs an agent's instructions with a Logfire-managed prompt so you can version, label, roll out, and A/B test prompts from the Logfire UI without redeploying.

## Usage

```python
from pydantic_ai import Agent
from pydantic_ai_harness.logfire import ManagedPrompt

agent = Agent(
    'openai:gpt-5',
    capabilities=[
        ManagedPrompt(
            'support_agent',
            default='You are a helpful customer support agent. Be friendly and concise.',
            label='production',
        )
    ],
)
```

The slug `support_agent` is declared as the managed variable `prompt__support_agent` (Logfire's Prompt-management naming; hyphens → underscores). The `default` is the offline safety net used when no remote value is available.

## How it works

- Resolves the prompt **once per run** inside `wrap_run`, keeping the `ResolvedVariable` open as a context manager for the whole run — so the selected **label/version** are attached as baggage to every child span (model requests, tools, and the run span). Runs outermost (`wraps=[Instrumentation]`) so the run span is tagged too.
- The per-run resolution is held in a `ContextVar`, so a single capability instance is safe across concurrent runs.

## API

- `ManagedPrompt(slug, *, default, label=, targeting_key=, attributes=, render_template=, logfire_instance=)`.
- `targeting_key`/`attributes` accept a static value or a callable of `RunContext` (e.g. `lambda ctx: ctx.deps.user_id`).
- Pass an existing `logfire.Variable` instead of a slug to use your own declaration. Declaring the same slug twice is fine (each builds its own backing variable).
- `render_template=True` renders the prompt as Handlebars against `deps` (like `TemplateStr`); off by default. Requires `pydantic-ai-slim[spec]`.
- `ManagedPrompt.resolved` exposes the active run's `ResolvedVariable` (value/label/version/reason).

## Packaging

- New optional extra: `pydantic-ai-harness[logfire]` (`logfire>=4.31.0`, the CI-tested floor).
- Built directly on the public `logfire.Variable` API (no registry side effects, so duplicate slugs don't raise on any supported logfire version).
- Lives in `pydantic_ai_harness.logfire`; the test directory is named `logfire_variables` to avoid shadowing the third-party `logfire` package under pyright's `tests` execution-environment root.

## Testing

- 100% branch coverage; pyright-strict and ruff clean.
- Covers code-default fallback, provider-backed resolution via `LocalVariablesOptions` (remote value + label baggage), span/baggage assertions via `capfire` inline snapshots, targeting, dedup, the `prompt__` prefix warning, and Handlebars rendering.

## Future: first-party `Managed` capability

A broader `pydantic_ai.managed.logfire.Managed` capability is in flight upstream in [pydantic-ai#5107](https://github.com/pydantic/pydantic-ai/pull/5107) — it'll cover instructions, model settings, and whole-spec variables, with a planned phase-2 `contribute_run_spec` hook. Once it ships, users will be able to import it directly from `pydantic-ai` and `ManagedPrompt` here can either delegate to it or stay as the prompt-only convenience.

The README's `ManagedPrompt` section now carries a callout pointing at #5107 so users picking it up today know where things are headed.

## Related

- [#90](https://github.com/pydantic/pydantic-ai-harness/issues/90) — Prompt Templates / Parameterized User Prompts (complementary: user-prompt side)
- [#99](https://github.com/pydantic/pydantic-ai-harness/issues/99) — Cache-Aware Prompt Construction (this capability is the first concrete example of the trade-off; see the new README section)
- [#147](https://github.com/pydantic/pydantic-ai-harness/issues/147) — `ModelSettings.cache_control` (the cache-control surface that managed-prompt rollouts will interact with)


pydantic_ai_harness/logfire/_managed_prompt.py
```diff
@@ -0,0 +1,194 @@
...
+def _new_resolved_var() -> ContextVar[ResolvedVariable[str] | None]:
+    # `None` means nothing has been resolved for the active run.
+    return ContextVar('managed_prompt_resolved', default=None)
+
+
+@dataclass
+class ManagedPrompt(AbstractCapability[AgentDepsT]):
+    """Back an agent's instructions with a Logfire-managed prompt.
+
+    Pass a prompt slug and a code default and the capability declares the backing
```
strawgate:
```suggestion
    Pass the managed prompt name and a default value and the capability declares the backing
```

In Logfire we just call it the name of the prompt

pydantic_ai_harness/logfire/_managed_prompt.py
```diff
@@ -0,0 +1,194 @@
...
+    [managed variable](https://logfire.pydantic.dev/docs/reference/advanced/managed-variables/)
+    for you -- a slug of `support_agent` resolves the variable `prompt__support_agent`, matching
```
strawgate: name?

pydantic_ai_harness/logfire/_managed_prompt.py
```diff
@@ -0,0 +1,194 @@
...
+    of the agent run.
+
+    Declaring the same slug more than once is fine -- each `ManagedPrompt` constructs its own
+    backing variable, so sharing a prompt across several agents just works. Pass an existing
+    [`logfire.Variable`][logfire.Variable] as `prompt` instead of a slug when you want to use a
+    variable you defined yourself (for example a `template_var`, or one registered for
+    [`variables_push`][logfire.Logfire.variables_push]).
+    """
+
+    prompt: str | Variable[str]
```
strawgate: name=?

otherwise
`ManagedPrompt(prompt="support_agent")` vs `ManagedPrompt(name="support_agent")`

pydantic_ai_harness/logfire/_managed_prompt.py
```diff
@@ -0,0 +1,194 @@
...
+    """The prompt slug (declared as the variable `prompt__<slug>`), or a pre-built `logfire.Variable`."""
+
+    default: str | None = None
+    """Code-default prompt text. Required when `prompt` is a slug; ignored when `prompt` is a `Variable`."""
+
+    label: str | None = None
+    """Explicit label to resolve (e.g. `'production'`). When `None`, the variable's
+    rollout and targeting rules select the label."""
```
strawgate:
```suggestion
    """Explicit targeting label on the Logfire Managed Prompt to resolve (e.g. `'production'`). When `None`, the targeting rules on the Managed Variable select the label."""
```

pydantic_ai_harness/logfire/_managed_prompt.py
```diff
@@ -0,0 +1,194 @@
...
+
+    targeting_key: str | Callable[[RunContext[AgentDepsT]], str | None] | None = None
+    """Key for deterministic label selection, or a callable that derives it from the
```
strawgate: what does this mean?

pydantic_ai_harness/logfire/_managed_prompt.py
```diff
@@ -0,0 +1,194 @@
...
+    [`RunContext`][pydantic_ai.tools.RunContext]. When `None`, Logfire falls back to its
+    own targeting context and then the active trace id."""
+
+    attributes: Mapping[str, Any] | Callable[[RunContext[AgentDepsT]], Mapping[str, Any] | None] | None = None
```
strawgate, line 104: targeting_attributes?

strawgate: should we make a targeting class / typeddict? or something?

pydantic_ai_harness/logfire/README.md
```diff
@@ -0,0 +1,173 @@
...
+run's `wrap_run` hook using the
+[`ResolvedVariable`](https://logfire.pydantic.dev/docs/reference/advanced/managed-variables/)
+as a context manager that stays open for the whole run -- so the selected label and version
+are attached as baggage to every child span of the agent run. You get a direct correlation
+between a run's behavior and the exact prompt version that produced it, plus instant
+iteration and rollback from the Logfire UI.
+
+### Usage
+
+Pass the prompt slug and a code default. The slug `support_agent` is declared as the managed
```
strawgate:
```suggestion
Pass the prompt name and a default value. The name `support_agent` is declared as the managed
```

pydantic_ai_harness/logfire/_managed_prompt.py
dmontagu: it's related to how rollouts work when you do a percentage rollout. It's basically the seed for the RNG

pydantic_ai_harness/logfire/_managed_prompt.py
dmontagu: I was going with openfeature nomenclature but we can change it. This needs to be free form not a specific type I think, it's meant to be user-defined.

pydantic_ai_harness/logfire/_managed_prompt.py
dmontagu: Claude here: Done in 6039560 — reworded to "Pass the managed prompt name and a default value..." and switched the prose from "slug" to "name" throughout, since Logfire refers to it as the prompt name.

pydantic_ai_harness/logfire/_managed_prompt.py
dmontagu: Claude here: Done in 6039560 — this now reads "a name of `support_agent` resolves the variable `prompt__support_agent`".

pydantic_ai_harness/logfire/_managed_prompt.py
dmontagu: Claude here: Agreed — renamed the parameter from `prompt` to `name` in 6039560, so it's `ManagedPrompt(name="support_agent")` (or first positional). Updated docstrings, error messages, and the README to match.

pydantic_ai_harness/logfire/_managed_prompt.py
dmontagu: Claude here: Applied your suggestion in 6039560 (with "managed prompt"/"managed variable" lowercased to match the surrounding prose).

pydantic_ai_harness/logfire/_managed_prompt.py
dmontagu: Claude here: Expanded the `targeting_key` docstring in 6039560 to explain it the way you described — it's the stable seed for Logfire's deterministic rollout assignment, so the same key always lands in the same percentage bucket and a given user keeps the same label across runs.

pydantic_ai_harness/logfire/README.md
dmontagu: Claude here: Applied in 6039560 — "Pass the prompt name and a default value. The name `support_agent`...".

pydantic_ai_harness/logfire/_managed_prompt.py
dmontagu: Claude here: Keeping `attributes` for now (matches OpenFeature's evaluation-context nomenclature, and per the thread it's intentionally free-form rather than a fixed type/TypedDict). Leaving this thread open so we can revisit the `targeting_attributes` naming if we decide it reads better at the call site.

PR #260, opened 2026-05-27
feat: add FileSystem and Shell capabilities

strawgate:
## Summary

Adds two new batteries-included capabilities built on the `AbstractCapability` / `FunctionToolset` primitives.

### `FileSystem`
Sandboxed file-system access scoped to a root directory.
- All paths resolved and containment-checked before any I/O (traversal prevention, symlink-aware)
- Allow / deny / protected pattern filtering (globs), matched against the canonical path (a leading `**/` also covers the root)
- Protected defaults: `.git/`, `.env`, `*.pem`, `*.key`, secrets
- Tools: `read_file`, `write_file`, `edit_file` (hash-based optimistic concurrency), `list_directory`, `search_files`, `find_files`, `create_directory`, `file_info`
- `__post_init__` validates that integer limit fields (`max_read_lines`, `max_search_results`, `max_find_results`) are positive non-bool ints

### `Shell`
Configurable command execution for agents.
- Allow/deny command lists; deny operator list (e.g. `>`, `>>`, `|`)
- Interactive command detection (vi, sudo, ssh, …)
- Synchronous (`run_command`) and background (`start_command` / `check_command` / `stop_command`) execution
- Output truncated and labelled with stdout/stderr/exit-code markers
- Optional `persist_cwd` so `cd` is sticky across calls — the working directory is captured out-of-band (a private temp file), so command output can't spoof it
- **`__aexit__` terminates all remaining background processes** — the agent runtime enters toolsets via `AsyncExitStack`, so cleanup is guaranteed on run end (success or error)
- `for_run` returns a fresh toolset per run, so persistent cwd and background processes are isolated across concurrent runs

### Robustness
- Model-correctable tool errors (missing file, denied path, stale edit, denied command) surface as `ModelRetry` so the agent can self-correct instead of aborting the run
- Both capabilities and their toolsets are generic over `AgentDepsT`

## Test coverage

- **100% branch coverage** (enforced by `fail_under = 100`)
- Mutation testing: **89.7% kill rate** (60 provably-equivalent survivors documented in `docs/mutation-testing.md`)
- pyright strict: 0 errors

## Validation

```
make lint && make typecheck && make testcov
```
All pass.

## Follow-ups (tracked separately, out of scope here)
- ripgrep-backed search/find/list for large-repo performance (deferred per review)
- `max_file_size` guard + search timeout (ReDoS / OOM hardening)
- read-only mode / per-tool selection; deps-scoped roots; HITL approval for destructive ops


pydantic_ai_harness/filesystem/_capability.py
```diff
@@ -0,0 +1,74 @@
...
+
+@dataclass
+class FileSystem(AbstractCapability[Any]):
+    """File system access scoped to a root directory.
+
+    All paths are resolved relative to `root_dir`. Traversal above the root
+    is rejected. Symlinks are resolved before authorization.
+    """
+
+    root_dir: str | Path = '.'
```
strawgate, line 34: we should make directory protection optional

pydantic_ai_harness/filesystem/_capability.py
```diff
@@ -0,0 +1,74 @@
...
+
+    max_search_results: int = 1000
+    """Maximum number of matches returned by `search_files`."""
+
+    max_find_results: int = 1000
+    """Maximum number of matches returned by `find_files`."""
+
+    def __post_init__(self) -> None:
+        for name in ('max_read_lines', 'max_search_results', 'max_find_results'):
+            value = getattr(self, name)
```
strawgate: bad claude

pydantic_ai_harness/filesystem/_toolset.py
```diff
@@ -0,0 +1,392 @@
...
+        self.add_function(self.file_info, name='file_info')
+
+    def _resolve_path(self, path: str) -> Path:
+        """Resolve path relative to root, rejecting traversal.
+
+        Uses os.path.realpath for symlink resolution before checking containment.
+        """
+        candidate = (self._root / path).resolve()
+        real = Path(os.path.realpath(candidate))
+        real_root = Path(os.path.realpath(self._root))
```
strawgate: we should probably just calculate this during Init

pydantic_ai_harness/filesystem/_toolset.py
```diff
@@ -0,0 +1,392 @@
...
+        if not real.is_relative_to(real_root):
+            raise PermissionError(f'Path {path!r} resolves outside the root directory.')
+
+        return real
+
+    def _check_access(self, path: str, *, write: bool = False) -> None:
+        """Validate path against allow/deny/protected patterns."""
+        if write and self._protected_patterns:
+            matched = next((p for p in self._protected_patterns if fnmatch.fnmatch(path, p)), None)
```
strawgate: this probably warrants like an _in_pattern_list private method

pydantic_ai_harness/filesystem/_toolset.py
```diff
@@ -0,0 +1,392 @@
...
+        self._check_access(path, write=write)
+        return self._resolve_path(path)
+
+    async def read_file(self, path: str, *, offset: int = 0, limit: int | None = None) -> str:
+        """Read a text file with line numbers.
+
+        Args:
+            path: File path relative to the root directory.
+            offset: Zero-based line offset to start reading from.
+            limit: Maximum number of lines to return (default: max_read_lines).
```
strawgate: unfortunately constants in docstrings aren't super helpful

pydantic_ai_harness/filesystem/_toolset.py
```diff
@@ -0,0 +1,392 @@
...
+                raise FileNotFoundError(f"'{path}' is a directory, not a file.")
+            raise FileNotFoundError(f'File not found: {path}')
+
+        raw = resolved.read_bytes()
+        if _is_binary(raw):
+            size = len(raw)
+            return f'[Binary file: {size} bytes. Use a binary-aware tool to inspect.]'
+
+        text = raw.decode('utf-8', errors='replace')
+        total_lines = len(text.splitlines())
```
strawgate: we should split lines here maybe so we can calculate the full length and pass them to format lines instead of splitting the whole input twice

pydantic_ai_harness/filesystem/_toolset.py
```diff
@@ -0,0 +1,392 @@
...
+        if expected_hash is not None and resolved.is_file():
+            current = resolved.read_text(encoding='utf-8')
+            current_hash = _content_hash(current)
+            if current_hash != expected_hash:
+                raise ValueError(
+                    f'Conflict: file {path!r} has changed (expected hash:{expected_hash}, '
+                    f'got hash:{current_hash}). Re-read the file and retry.'
+                )
+
+        resolved.parent.mkdir(parents=True, exist_ok=True)
```
strawgate: we should throw an exception if the dir doesn't exist

pydantic_ai_harness/filesystem/_toolset.py
```diff
@@ -0,0 +1,392 @@
...
+                rel = str(entry.relative_to(real_root))
+            except ValueError:  # pragma: no cover
+                continue
+            if entry.is_dir():
+                entries.append(f'{rel}/')
+            else:
+                try:
+                    size = entry.stat().st_size
+                except OSError:  # pragma: no cover
+                    size = 0
```
strawgate, line 352: we should add a comment indicating when this occurs

pydantic_ai_harness/filesystem/_toolset.py
```diff
@@ -0,0 +1,392 @@
...
+            rel_str = str(file_path.relative_to(real_root))
+            if include_glob and not fnmatch.fnmatch(rel_str, include_glob):
+                continue
+            try:
+                raw = file_path.read_bytes()
+            except OSError:  # pragma: no cover
+                continue
+            if _is_binary(raw):
+                continue
+            text = raw.decode('utf-8', errors='replace')
```
strawgate, line 407: It might be worth seeing if we want to shell out to ripgrep. I have a library https://github.com/strawgate/rpygrep that manages some of that but we can probably find another alternative.

pydantic_ai_harness/shell/_toolset.py
```diff
@@ -0,0 +1,440 @@
...
+                await bg.proc.aclose()
+            self._cleanup_bg_files(bg)
+        self._background.clear()
+
+    def _check_command(self, command: str) -> None:
+        """Validate command against allow/deny lists."""
+        if not self._allow_interactive and _is_interactive_command(command):
+            raise PermissionError(f'Interactive commands are not allowed. Command: {command!r}')
+
+        matched_op = next((op for op in self._denied_operators if op in command), None)
```
strawgate: probably worth a helper we can test in isolation

pydantic_ai_harness/shell/_toolset.py
```diff
@@ -0,0 +1,440 @@
...
+            if not bg.finished:
+                await self._kill_process_group(bg.proc)
+                with anyio.CancelScope(shield=True):
+                    await bg.proc.wait()
+                await bg.proc.aclose()
+            self._cleanup_bg_files(bg)
+        self._background.clear()
+
+    def _check_command(self, command: str) -> None:
+        """Validate command against allow/deny lists."""
```
strawgate: we probably need to document in the toolset that these are best effort and are not a security boundary

pydantic_ai_harness/shell/_toolset.py
```diff
@@ -0,0 +1,440 @@
...
+            return text
+        if not stderr_text:
+            return text[: self._max_output_chars] + f'\n[... output truncated at {self._max_output_chars} chars]'
+
+        stderr_budget = min(len(stderr_text) + len('[stderr]\n'), self._max_output_chars // 3)
+        stdout_budget = self._max_output_chars - stderr_budget
+        truncated = text[:stdout_budget] + f'\n[... stdout truncated at {stdout_budget} chars]'
+        return truncated
+
+    def _wrap_command_for_cwd(self, command: str) -> str:
```
strawgate: this may end up being dubious but I think we can leave it for now

strawgate (commented):
have not run through the tests yet

pyproject.toml
```diff
@@ -140,3 +141,18 @@ exclude_lines = [
...
+]
+tests_dir = ['tests/filesystem/', 'tests/shell/']
+also_copy = ['pydantic_ai_harness/', 'tests/']
+# Skip trio-parametrized tests during mutation testing — trio segfaults in
+# mutmut's subprocess environment on Python 3.14 (not a code bug).
+pytest_add_cli_args = ['-k', 'not trio']
+# Required on Python 3.14 / macOS — mutmut's subprocess workers segfault
+# without debug mode.
+debug = true
+# See docs/mutation-testing.md for full results (89.7% kill rate, 60 equivalent mutants).
```
strawgate, line 154: Remove

pydantic_ai_harness/filesystem/_toolset.py
```diff
@@ -0,0 +1,396 @@
...
+
+        Returns:
+            A newline-separated listing with type indicators and sizes.
+        """
+        resolved = self._safe_resolve(path)
+        if not resolved.is_dir():
+            raise NotADirectoryError(f'Not a directory: {path}')
+
+        entries: list[str] = []
+        for entry in sorted(resolved.iterdir()):
```
mpfaffenberger, line 331: Recursive listing would be extremely slow on large projects. I would recommend adding ripgrep as a hard dependency (even for listing files)

If you want to see Python harness tooling that is battle tested (millions of uses per day) please look at https://github.com/mpfaffenberger/code_puppy

pydantic_ai_harness/filesystem/_toolset.py
mpfaffenberger: Ripgrep 100%

dsfaccini:
Claude here: full review of #260 across correctness, security, repo conventions, pyai-idiomatic API usage, and extensibility. CI is green (all 18 checks), but a few things the green build can't catch are worth resolving before merge. Delegated parts of this to subagents (a security pass, a convention-drift pass) and to a fresh Claude session running inside the `pydantic-ai` core checkout for the API-idiom questions.

I've tried to be concrete and to back the load-bearing claims with sources. Two of these (B3, B4) I know you're skeptical of — I've put the strongest evidence I have on those.

---

## Blockers I'd want fixed before merge

### B1 — `search_files` / `find_files` / `list_directory` bypass the entire allow/deny/protected model

`FileSystemToolset._check_access` runs only on the **directory argument**, never on the per-result entries produced by `rglob` / `glob` / `iterdir`. Those results are filtered only by `part.startswith('.')`. Consequences:

- `denied_patterns` does **nothing** for these three enumeration tools.
- The default protected set (`*.pem`, `*.key`, `**/secrets*`) — none of which are dotfiles — is fully readable. `search_files(pattern="PRIVATE KEY")` will read and return matching lines from every key file in the tree.

This contradicts the security model the toolset's own docstring advertises. Fix: run `_check_access` (read mode) on each candidate and skip on `PermissionError`. (This is also where the ripgrep discussion below lands — a ripgrep-backed implementation would re-do this traversal anyway, so it's worth solving them together.)

### B2 — recoverable tool errors abort the run instead of letting the model retry

Every tool raises bare `PermissionError` / `FileNotFoundError` / `ValueError` / `NotADirectoryError`. pyai core only turns `ModelRetry` (and `ValidationError`) into a retry prompt the model can recover from — see `_tool_manager.py`, where only `SkipToolExecution` / `CallDeferred` / `ApprovalRequired` / `ModelRetry` are special-cased and everything else propagates and ends the run.

So "file not found", "old_text not found", "command denied", "invalid regex" — all conditions the model should be able to correct on its own — currently crash the agent. The new code raises `ModelRetry` **nowhere**. Fix: raise `ModelRetry` for model-correctable conditions; reserve raw exceptions for genuine internal failures.

### B3 — mutable run-state on a shared toolset, no `for_run` isolation

This is your "are toolsets supposed to be shareable?" question — and the answer from core is: **yes, by default they're shared, and that is exactly why core gives you `for_run` to opt out when you hold mutable state.** Quoting the base contract (`AbstractToolset.for_run`, pydantic-ai v1.95.1):

> Called once per run, before `__aenter__`. Override this to return a fresh instance for per-run state isolation. Default: return `self` (shared across runs).

Source: https://github.com/pydantic/pydantic-ai/blob/v1.95.1/pydantic_ai_slim/pydantic_ai/toolsets/abstract.py#L110-L116

So a **stateless** toolset is fine to share — you're right about that. The issue is specific: `get_toolset` is called once at agent construction, so a single `ShellToolset` instance is reused across every run of that agent, and it carries mutable per-run state — `self._background` (the background-process dict) and `self._cwd` (persistent cwd). Two concurrent runs of the same agent will corrupt each other: run A's `cd` moves run B's cwd, and run A's `__aexit__` terminates run B's background processes.

The in-repo precedent is our own exemplar: `CodeModeToolset.for_run` returns `replace(self, …)` precisely to give each run a fresh instance with reset state, and documents why. The two new toolsets override neither `for_run` nor `for_run_step`. Fix: implement `for_run` to return a fresh instance with an empty `_background` and reset `_cwd`. (This also makes the otherwise well-written `__aexit__` cleanup sound — `__aexit__` fires per run, so on a shared instance it's tearing down state that may belong to another run.)

### B4 — `persist_cwd` sentinel is spoofable / silently self-disabling

I know you flagged this thread as "may be dubious… leave for now." I think it's a real correctness footgun, less as a security hole and more as a silent-wrong-behavior one:

- `_wrap_command_for_cwd` skips wrapping whenever `';' in command`. That substring test also fires on a `;` inside a quoted string (`echo "a;b"`), so cwd tracking silently turns off for perfectly benign commands.
- `_extract_cwd_from_output` parses a fixed literal `__HARNESS_PWD__` out of stdout. `cd /tmp; echo __HARNESS_PWD__/etc` runs unwrapped (because of the `;` skip), exits 0, and the parser moves the persistent cwd to `/etc`. Any command that emits that literal can steer subsequent commands' working directory.

The exploit window for *malicious* redirection is narrow (the genuine appended sentinel usually wins via `rfind`), so I wouldn't sell this as a security hole — but the silent-disable and the spoof-on-`;`-skip paths are real wrong behavior. Fix: write `pwd` to an out-of-band channel (a temp file the wrapper writes, read by the toolset) instead of parsing it out of stdout.

---

## Unresolved review threads — are they blockers?

All 14 are still open. Breakdown:

- **6 are outdated** (already addressed by the "non-controversial feedback" commit but never marked resolved): `calculate during init`, `_in_pattern_list` helper, split-lines, throw-if-dir-missing, the "best-effort, not a security boundary" doc, and the line-259 comment. Not blockers — just worth resolving so the PR stops looking like it has 14 live objections.
- **ripgrep** (raised by both reviewers, on `search_files` and `list_directory`): "recursive listing extremely slow on large projects." Effectively a blocker given two independent asks — either adopt ripgrep or get explicit agreement to defer. Ties into B1.
- **"make directory protection optional"** (`_capability.py` `root_dir`): protection is already opt-out via an empty `protected_patterns` — worth clarifying whether you want protections **off by default** (opt-in) instead. Design call.
- **"bad claude"** (the `__post_init__` getattr-loop validation): trivial — replace the reflective loop with explicit validation.
- **"constants in docstrings aren't helpful"** (`read_file` docstring `default: max_read_lines`): doc nit.
- **`pyproject.toml:154` "Remove"**: ties to D3 below (strip mutmut config from this PR).
- **the cwd thread**: see B4.

No new surprises hiding in the threads — ripgrep and the cwd sentinel are the two substantive open ones.

---

## Repo-convention drift

- **D1 — no `README.md`.** `code_mode` and the sibling capability PRs (#249, #251) all ship one, and `agent_docs/index.md`, `capability-authoring.md`, and `review-checklist.md` all require it. Skeleton to match: `# Title → The problem → The solution → Usage (runnable Agent(...) example) → feature sections → API`. Shell's README must document the safety model.
- **D2 — `AbstractCapability[Any]` instead of `[AgentDepsT]`.** Base is `Generic[AgentDepsT]`; `CodeMode`, `InputGuard`, `StepPersistence` all stay generic, and the repo bans `Any` in public signatures. These tools don't touch deps, so it's a pure type-hygiene fix: `FileSystem(AbstractCapability[AgentDepsT])`, `FunctionToolset[AgentDepsT]`, return `AgentToolset[AgentDepsT]`. (Also `Shell.get_toolset` is annotated `| None` but always returns a toolset — tighten it.)
- **D3 — per-PR mutmut config in repo-wide `pyproject.toml`.** Adds the `mutmut` dev-dep + a `[tool.mutmut]` block hardcoding *this PR's two files* + `mutants` exclusions + `scripts/run-mutmut.sh` + `docs/mutation-testing.md`. No prior capability PR carries this, and it bumps against the "don't edit `pyproject.toml`/`uv.lock`, open an issue for deps" rule. Suggest stripping it from the PR (keep the 89.7% result in the PR description) and landing mutation testing repo-wide separately if wanted.
- **D4 — PR body tool names wrong.** Body says `list_files` / `make_dir`; code registers `list_directory` / `create_directory`. Pick the public names deliberately and make code + body + README agree.
- **D5** — class docstrings lack the runnable `python` usage example `code_mode` has, and the root `README.md` capability table still shows File system / Shell as `:construction:` linking to the old PR #177.

---

## pyai-idiomatic API review (is this reinventing core?)

Checked against core source:

- **Not reinventing implementations** — core has no filesystem or shell tools (`common_tools/` is web-only), no FS/process sandbox. Harness is the right home. ✅
- **Missing HITL/approval composition** — core ships `requires_approval=True`, `ApprovalRequiredToolset` / `.approval_required(func)`, `ApprovalRequired` / `CallDeferred`, and the `HandleDeferredToolCalls` capability. Hard-`PermissionError`-denying destructive ops (`rm`, writes to protected paths) is exactly what that flow exists for, and `code_mode` already integrates with it. This is the highest-leverage *design* improvement, but it's a bigger conversation — flagging, not asking you to add it silently.
- Subclassing `FunctionToolset` + `add_function(self.method, …)` for static config is fine and supported; the only problem is the mutable state (→ B3).

---

## Security (beyond B1/B4)

- **Patterns match the agent-supplied string, not the canonical path** (`_check_access`): `config/./secret.txt` evades `denied=['config/secret.txt']`, and `fnmatch` doesn't implement `**`, so the default `**/secrets*` does **not** protect a root-level `secrets.yaml`. Match against the resolved relative path.
- **`_real_root` is captured once at construction**, so a parent-dir symlink swap can defeat containment (precondition: write access to a parent of root).
- **ReDoS / event-loop block** in `search_files` (unbounded agent-supplied regex, no timeout) and **whole-file `read_bytes` before any size guard** (no max-file-size knob).
- The shell denylist/allowlist bypasses (`tokens[0]`-only, `/bin/rm`, `bash -c`, `$(...)`, substring operator matching) are **inherent and correctly disclaimed** — not defects. One ask: lift that "not a security boundary" disclaimer from the private `_check_command` up into the public `Shell` dataclass docstring, since `denied_commands` / `denied_operators` read as protective there.

---

## Missing knobs / extensibility (opinion — not asking to expand scope here, just flagging what may gate adoption)

1. **Read-only / per-tool selection on `FileSystem`.** Today it's all 8 tools or nothing. "Give the agent read-only repo access" is one of the most common asks and is currently impossible without subclassing. A `read_only: bool` or a tool-subset selector (mirroring `CodeMode.tools`) is small and high-value.
2. **deps-scoped `root_dir` / `cwd`.** Root/cwd are frozen at construction and tools take no `RunContext`, so you can't scope the sandbox per-run/per-user (multi-tenant). Fixing D2 + B3 is the prerequisite; deriving root from `ctx.deps` would unlock it. Worth at least not foreclosing.
3. **HITL approval hook** (above) — for shell especially, "destructive op → ask a human" is the difference between a demo and something teams deploy.

Lower-priority, worth tracking in issues rather than this PR: `max_file_size`, env-var control for `Shell` (inject/scrub secrets from the subprocess env), configurable encoding, and the ripgrep backend.

---

Happy to implement B1–B4 + D2 and draft the two READMEs if you want — or to open follow-up issues for read-only mode, deps-scoped roots, HITL approval, and ripgrep so they're tracked without bloating this PR.

pydantic_ai_harness/shell/_toolset.py
```diff
@@ -0,0 +1,449 @@
...
+            raise PermissionError(f'Command {executable!r} is denied.')
+        if self._allowed_commands and executable not in self._allowed_commands:
+            raise PermissionError(f'Command {executable!r} is not in the allowed list.')
+
+    def _truncate(self, text: str, *, stderr_text: str = '') -> str:
+        """Truncate output, reserving space for stderr when both streams are present."""
+        if len(text) <= self._max_output_chars:
+            return text
+        if not stderr_text:
+            return text[: self._max_output_chars] + f'\n[... output truncated at {self._max_output_chars} chars]'
```
adtyavrdhn: Shell output should preserve tail instead of head, more so I think for errors

pydantic_ai_harness/shell/_toolset.py
```diff
@@ -0,0 +1,449 @@
...
+
+    def _wrap_command_for_cwd(self, command: str) -> str:
+        """Append pwd sentinel to command for cwd tracking.
+
+        Commands containing ';' are returned unwrapped because the separator
+        breaks the '&&' success-gating of the sentinel echo.
+        """
+        if ';' in command:
+            return command
+        return f'{command} && echo {_PWD_SENTINEL}$(pwd)'
```
dsfaccini: Claude here: this is the B4 footgun. Two concrete problems with parsing the cwd out of stdout:

**1. `;` silently disables cwd tracking.** Any command containing `;` is returned unwrapped, so a real `cd` is dropped on the floor:
```
run_command('cd subdir ; true')   # persist_cwd=True
run_command('pwd')                # -> still the OLD dir; the cd never persisted
```
The substring test also misfires on a `;` inside a quoted string, disabling tracking for benign commands.

**2. The sentinel is spoofable through that same `;`-skip path.** A command whose stdout contains the literal `__HARNESS_PWD__<dir>` moves the tracked cwd with no real `cd`:
```
run_command('true ; echo __HARNESS_PWD__/some/existing/dir')
# -> self._cwd is now /some/existing/dir; the next command runs there
```

Regression tests proving both (they fail on this revision, pass after the fix):
```python
async def test_cd_persists_even_with_semicolon(self, persist_toolset):
    await persist_toolset.run_command('cd subdir ; true')
    assert 'subdir' in await persist_toolset.run_command('pwd')

async def test_output_cannot_spoof_cwd(self, persist_toolset, shell_dir):
    await persist_toolset.run_command(f'true ; echo __HARNESS_PWD__{shell_dir / "subdir"}')
    assert persist_toolset._cwd == shell_dir   # no real cd happened
```

**Fix (implemented locally):** drop the stdout sentinel entirely and capture `pwd` out-of-band into a private temp file the agent's command can't address — `{command}\n__harness_ec=$?\npwd > <random_tempfile>\nexit $__harness_ec` — then read the cwd from that file. No `;`-skip, nothing to spoof. Happy to push it onto the branch or hand you the diff, whichever you prefer.

PR #262, opened 2026-05-30
Add host-backed OS and filesystem access to CodeMode sandbox

adtyavrdhn:
## Summary

Extends `CodeMode` to support host-backed OS and filesystem access inside the sandbox via two new parameters:

- `os`: accepts a `pydantic_monty.AbstractOS` instance or a raw callback `(function_name, args, kwargs) -> result` to handle OS calls like `os.getenv()`, `datetime.now()`, and `pathlib.Path` operations
- `mount`: accepts one or more `pydantic_monty.MountDir` to expose host directories inside the sandbox

When either is configured, the `run_code` tool description is updated to reflect that these operations are host-backed instead of unavailable. The `os`/`mount` parameters are threaded through every `resume()` call in the execution loop to ensure OS dispatch persists across tool call suspensions and REPL state reuse.

## Linked Issue

Fixes #

### Core implementation (`_toolset.py`)
- Added `MontyOS`, `MontyOSCallback`, and `MontyMount` type aliases for clarity
- Split `_RUN_CODE_BASE_DESCRIPTION` into head/tail with swappable restriction lines (`_NO_OS_RESTRICTION` vs `_OS_ENABLED_NOTE`)
- Added `_base_description()` helper to assemble descriptions based on OS enablement
- Extended `CodeModeToolset` with `os` and `mount` fields
- Updated `_build_description()` to accept `os_enabled` parameter and use dynamic base description
- Modified `call_tool()` to pass `os`/`mount` to `feed_start()`
- Updated `_execution_loop()` and all snapshot handlers (`_handle_function_snapshot`, `_resolve_future_snapshot`) to accept and forward `os`/`mount` to every `resume()` call

### Capability layer (`_capability.py`)
- Added `os` and `mount` parameters to `CodeMode` class with docstrings
- Updated `get_wrapper_toolset()` to forward these parameters to `CodeModeToolset`

### Public API (`__init__.py`)
- Exported `MontyOS`, `MontyOSCallback`, and `MontyMount` type aliases

### Documentation (`README.md`)
- Added "Filesystem and OS access" section with examples of `MountDir`, `OSAccess`, and raw callbacks
- Updated "Sandbox restrictions" to clarify that filesystem/clock operations are unavailable by default but become available with host-backed access
- Updated API reference to document new `os` and `mount` parameters

### Tests (`test_code_mode.py`)
- Added `TestCodeModeOSAccess` class with 10 comprehensive tests covering:
  - Description changes when OS access is enabled
  - OS callback dispatch through tool call suspensions and REPL reuse
  - `AbstractOS` instance dispatch
  - Exception handling in OS callbacks (surfaces as `ModelRetry`)
  - Single and multiple directory mounts
  - Parameter forwarding from `CodeMode` to `CodeModeToolset`

## Checklist

- [x] Tests added/updated for new behavior (10 new tests in `TestCodeModeOSAccess`)
- [x] No changes to `pyproject.toml` or `uv.lock`

https://claude.ai/code/session_0177ZDbsCTVRxMnE1yNJ5Fb5

### DIFF
--- a/pydantic_ai_harness/code_mode/_capability.py
+++ b/pydantic_ai_harness/code_mode/_capability.py
@@ -2,7 +2,7 @@
 
 from __future__ import annotations
 
-from dataclasses import dataclass, field
+from dataclasses import KW_ONLY, dataclass, field
 
 from pydantic_ai import AbstractToolset
 from pydantic_ai.capabilities import AbstractCapability, CapabilityOrdering
@@ -57,6 +57,10 @@ class CodeMode(AbstractCapability[AgentDepsT]):
 
     max_retries: int = 3
     """Maximum number of retries for the `run_code` tool (syntax errors count as retries)."""
+
+    _: KW_ONLY
+    # Everything below is keyword-only: the option list keeps growing, so new
+    # config must be passed by name rather than relying on positional order.
 
     os_access: CodeModeOS | None = None
     """Host-backed OS access for sandboxed code.
--- a/pydantic_ai_harness/code_mode/_capability.py
+++ b/pydantic_ai_harness/code_mode/_capability.py
@@ -65,7 +65,7 @@ class CodeMode(AbstractCapability[AgentDepsT]):
     `(function_name, args, kwargs) -> result`. When set, `pathlib.Path`, `os`,
     `datetime.datetime.now()`, and `datetime.date.today()` calls inside `run_code`
     are routed to it instead of being unavailable. Fixed at construction, so build
-    `CodeMode` per request to scope access per request.
+    `CodeMode` per request to scope access.
     """
 
     mount: MontyMount | None = None
--- a/pydantic_ai_harness/code_mode/_capability.py
+++ b/pydantic_ai_harness/code_mode/_capability.py
@@ -9,7 +9,7 @@ from pydantic_ai.capabilities import AbstractCapability, CapabilityOrdering
 from pydantic_ai.capabilities._tool_search import ToolSearch as _ToolSearch
 from pydantic_ai.tools import AgentDepsT, ToolSelector
 
-from pydantic_ai_harness.code_mode._toolset import CodeModeToolset, MontyMount, MontyOS
+from pydantic_ai_harness.code_mode._toolset import CodeModeMount, CodeModeOS, CodeModeToolset
 
 
 @dataclass
@@ -35,7 +35,7 @@ class CodeMode(AbstractCapability[AgentDepsT]):
     agent = Agent('openai:gpt-5', capabilities=[CodeMode(tools=['search', 'fetch'])])
     ```
 
-    Pass `mount` for host filesystem access and/or `os` for environment/clock
+    Pass `mount` for host filesystem access and/or `os_access` for environment/clock
     (plus filesystem) access -- without them, `pathlib`/`os` I/O and
     `datetime.now()` are unavailable inside `run_code`:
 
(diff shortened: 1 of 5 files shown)


pydantic_ai_harness/code_mode/_capability.py
```diff
@@ -48,10 +58,33 @@ class CodeMode(AbstractCapability[AgentDepsT]):
     max_retries: int = 3
     """Maximum number of retries for the `run_code` tool (syntax errors count as retries)."""
 
+    _: KW_ONLY
+    # Everything below is keyword-only: the option list keeps growing, so new
```
adtyavrdhn: No need to explain this lol

pydantic_ai_harness/code_mode/_capability.py
```diff
@@ -48,10 +58,33 @@ class CodeMode(AbstractCapability[AgentDepsT]):
...
+    # config must be passed by name rather than relying on positional order.
+
+    os_access: CodeModeOS | None = None
+    """Host-backed OS access for sandboxed code.
+
+    Pass a `pydantic_monty.AbstractOS` instance or a raw OS callback
+    `(function_name, args, kwargs) -> result`. When set, `pathlib.Path`, `os`,
```
adtyavrdhn: I don't see the point of explaining it in the docstring, the types should imply this

pydantic_ai_harness/code_mode/_capability.py
```diff
@@ -34,6 +34,23 @@ class CodeMode(AbstractCapability[AgentDepsT]):
...
+    - `os_access` routes the sandbox's OS calls to a handler you provide: reach for it
+      when the agent needs environment variables, the clock, or filesystem behavior you
+      control.
+
+    Both expose the real host to model-written code, so grant only what the task needs.
+
+    ```python
+    from pydantic_monty import MountDir
+
+    agent = Agent('openai:gpt-5', capabilities=[CodeMode(mount=MountDir('/work', '/tmp/agent-work'))])
```
adtyavrdhn, line 52: Should have the latest model

dsfaccini:
Claude here: reviewed on David's behalf, with a second opinion from Codex (a two-man pass).

**Verdict: mergeable.**

**Security:** host access is off by default and opt-in (`os_access`/`mount` default `None`, keyword-only, backwards-compatible); the wiring forwards user-supplied mounts verbatim and adds no escape vector of its own. Codex empirically probed mount escape (symlink, `..`, absolute paths) in both overlay and read-write modes; all were denied with `PermissionError`. Monty owns and enforces path scoping.

**Correctness:** `os`/`mount` are threaded through all nine Monty `feed_start`/`resume` call sites, including the error branches where auto-dispatch would otherwise stop. Gates pass at 100% branch coverage.

**On the monty floor:** Codex suggested bumping `pydantic-monty` to `>=0.0.17`. We disproved that by installing `0.0.16` in isolation: it already exports `MountDir`/`AbstractOS`/`OsFunction`/`NOT_HANDLED` and carries `os=`/`mount=` on `feed_start` and every snapshot `resume`, and a functional run works. So `>=0.0.16` is correct and the lowest-direct-deps job passes; no floor bump needed.

I pushed two small follow-ups so this is merge-ready:
- Extended the default-state `run_code` restriction note to say filesystem, env, and clock calls are unavailable without a mount or OS handler. The dynamic note already covered the mount and os-enabled states, but the default still listed `os`/`pathlib` as importable without noting their I/O fails, which is the wasted-retry case this PR reduces.
- Typed `CodeModeOSCallback` with `object` instead of `Any`. The harness only forwards the callback to Monty and never calls it, so `object` typechecks cleanly (pyright strict, 0 errors) and keeps `Any` out of the public API.

Non-blocking follow-up to consider: per-run scoping of `os_access`/`mount` (currently fixed at capability construction).

Thanks @adtyavrdhn.

PR #178, opened 2026-04-10
Add SubAgent capability for agent-to-agent task delegation

DouweM:
## Summary

- Adds `SubAgent` capability (`AbstractCapability` subclass) that lets a parent agent delegate tasks to named sub-agents via a `delegate_task` tool
- Sub-agent descriptions are injected into the system prompt (derived from `agent.description`, `agent.name`, or explicit `descriptions` dict)
- Parent deps are forwarded to sub-agents by default (`pass_deps=True`), configurable
- Unknown agent names trigger `ModelRetry` for self-correction
- 100% test coverage (20 tests), passes lint, format, and strict pyright

Closes #32

## Test plan

- [x] Construction: capability is `AbstractCapability`, descriptions resolve from agent metadata or explicit overrides, empty agents returns None for instructions/toolset
- [x] Instructions: system prompt lists all agents with descriptions
- [x] Toolset: `delegate_task` tool is registered with agent names in description
- [x] End-to-end delegation: parent delegates to sub-agent and gets result, unknown agent triggers retry then succeeds, deps forwarded when `pass_deps=True`, deps are None when `pass_deps=False`, multiple agents called sequentially
- [x] Imports: `SubAgent` importable from `pydantic_harness` package

> **Note:** Tests require `pydantic-ai-slim` with capabilities support (the `capabilities` branch). The `pyproject.toml` uses the PyPI dependency; for local development, override with `uv add --dev pydantic-ai-slim@/path/to/capabilities-branch/pydantic_ai_slim`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #267, opened 2026-06-03
feat: add SubAgents capability for task delegation

strawgate:
## Summary

Adds a `SubAgents` capability — an agent can delegate self-contained tasks to named child agents. A from-scratch take on the idea in #178, built on current repo conventions and public APIs only.

- **One tool, `delegate_task(agent_name, task)`.** Each delegation runs the chosen sub-agent in a **fresh, isolated run** (its own message history — it never sees the parent conversation) and returns `str(result.output)`.
- **Cache-stable discovery.** Available sub-agents are listed in the system prompt via `get_instructions` (each agent's `description`, or a `descriptions` override), so the listing stays in the cached prefix.
- **Deps forwarded, usage shared by default.** `deps` are passed to each sub-agent (sub-agents share the parent's `AgentDepsT`, enforced by the type signature). `usage` is shared so a parent `usage_limits` bounds the whole tree; `forward_usage=False` to isolate.
- **`inherit_tools`** (opt-in) exposes the parent's tools to sub-agents, with the delegate tool filtered out so they can't recurse into delegation.
- **`shared_capabilities`** are applied to every sub-agent run.
- **`event_stream_handler`** is forwarded to every sub-agent run, surfacing sub-agent events to the caller.
- **Soft sub-agent failures** (`ModelRetry` / `UnexpectedModelBehavior`) convert into a parent `ModelRetry` it can react to; hard limits (e.g. `UsageLimitExceeded`) propagate.

## Tests

`tests/subagents/` — construction/serialization, prompt listing, end-to-end delegation, unknown-agent retry, deps/usage forwarding, `inherit_tools`, `shared_capabilities`, `event_stream_handler`, and soft-failure conversion. `pyright --strict` clean, **100% branch coverage** of the package; `pyproject.toml`/`uv.lock` untouched.

https://claude.ai/code/session_01LQ9NTr8q95A99UnVcq6Nzf

### DIFF
--- a/pydantic_ai_harness/experimental/subagents/_capability.py
+++ b/pydantic_ai_harness/experimental/subagents/_capability.py
@@ -1,0 +1,113 @@
+"""Sub-agent capability: delegate self-contained tasks to named child agents."""
+
+from __future__ import annotations
+
+from collections.abc import Mapping, Sequence
+from dataclasses import dataclass, field
+from typing import TYPE_CHECKING, Any
+
+from pydantic_ai.agent import AbstractAgent, EventStreamHandler
+from pydantic_ai.capabilities import AbstractCapability, AgentCapability
+from pydantic_ai.tools import AgentDepsT
+from pydantic_ai.toolsets import AgentToolset
+
+from pydantic_ai_harness.experimental.subagents._toolset import SubAgentToolset
+
+if TYPE_CHECKING:
+    from pydantic_ai._instructions import AgentInstructions
+
+
+@dataclass
+class SubAgents(AbstractCapability[AgentDepsT]):
+    """Let an agent delegate self-contained tasks to named sub-agents.
+
+    Exposes a single `delegate_task(agent_name, task)` tool. Each delegation
+    runs the chosen sub-agent in a fresh, isolated run (it never sees the parent
+    conversation), and the available sub-agents are listed in the system prompt
+    as a static, cache-stable instruction.
+
+    The parent's `deps` are forwarded to each sub-agent (sub-agents therefore
+    share the parent's `AgentDepsT`), and by default the parent's `usage` is
+    shared so usage limits apply across the whole agent tree. Optionally, the
+    parent's tools can be inherited (`inherit_tools`), extra capabilities can be
+    applied to every sub-agent run (`shared_capabilities`), and sub-agent events
+    can be streamed to a handler (`event_stream_handler`).
+
+    ```python
+    from pydantic_ai import Agent
+    from pydantic_ai_harness.experimental.subagents import SubAgents
+
+    researcher = Agent('anthropic:claude-sonnet-4-6', name='researcher', description='Researches topics')
+    writer = Agent('anthropic:claude-sonnet-4-6', name='writer', description='Writes prose')
+
+    orchestrator = Agent(
+        'anthropic:claude-opus-4-7',
+        capabilities=[SubAgents(agents={'researcher': researcher, 'writer': writer})],
+    )
+    ```
+    """
+
+    agents: Mapping[str, AbstractAgent[AgentDepsT, Any]] = field(
+        default_factory=dict[str, 'AbstractAgent[AgentDepsT, Any]']
+    )
+    """Mapping of sub-agent name to the agent that runs when it's delegated to."""
+
+    descriptions: Mapping[str, str] | None = None
+    """Optional per-name description overrides for the system-prompt listing.
+
+    When a name is absent here, the agent's own `description` is used (if any)."""
+
+    forward_usage: bool = True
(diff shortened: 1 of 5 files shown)


pydantic_ai_harness/subagents/_toolset.py
```diff
@@ -0,0 +1,74 @@
...
+        """
+        agent = self._agents.get(agent_name)
+        if agent is None:
+            available = ', '.join(sorted(self._agents))
+            raise ModelRetry(f'Unknown sub-agent {agent_name!r}. Available sub-agents: {available}.')
+        toolsets = self._inherited_toolsets(ctx) if self._inherit_tools else None
+        capabilities = self._shared_capabilities or None
+        usage = ctx.usage if self._forward_usage else None
+        try:
+            result = await agent.run(task, deps=ctx.deps, usage=usage, toolsets=toolsets, capabilities=capabilities)
```
adtyavrdhn: Someone on the public slack asked for streaming via subagents and looking at this I think we can allow `event_stream_handler` to be passed as is here to do it. 

https://pydanticlogfire.slack.com/archives/C083V7PMHHA/p1780513031360839?thread_ts=1780510355.399129&cid=C083V7PMHHA

strawgate: Claude here: done — `SubAgents` now takes an `event_stream_handler` that's forwarded as-is to each sub-agent run, so sub-agent model-streaming and tool events surface to the caller (the handler gets the sub-agent's own `RunContext`). Added a test asserting events reach the handler and a README note. It's an optional field (default `None`), independent of the other knobs.

---
_Generated by [Claude Code](https://claude.ai/code)_

pydantic_ai_harness/experimental/subagents/_capability.py
```diff
@@ -0,0 +1,113 @@
...
+    ```python
+    from pydantic_ai import Agent
+    from pydantic_ai_harness import SubAgents
+
+    researcher = Agent('anthropic:claude-sonnet-4-6', name='researcher', description='Researches topics')
+    writer = Agent('anthropic:claude-sonnet-4-6', name='writer', description='Writes prose')
+
+    orchestrator = Agent(
+        'anthropic:claude-opus-4-7',
+        capabilities=[SubAgents(agents={'researcher': researcher, 'writer': writer})],
```
slmnsh, line 45: we can just accept array of agents and use the `name` attr from agent instance

pydantic_ai_harness/experimental/subagents/_capability.py
```diff
@@ -0,0 +1,113 @@
...
+    inherit_tools: bool = False
+    """If `True`, the parent agent's tools are exposed to each sub-agent run (the
+    delegate tool itself is filtered out, so sub-agents can't recurse into
+    further delegation). Off by default to avoid silently widening sub-agent access."""
+
+    shared_capabilities: Sequence[AgentCapability[AgentDepsT]] = ()
+    """Capabilities applied to every sub-agent run, in addition to whatever each
+    sub-agent already has."""
+
+    event_stream_handler: EventStreamHandler[AgentDepsT] | None = None
```
slmnsh, line 73: Can we propagate subagent streaming events to main agent loop? I think that will require some ~additions to Pydantic graphs~ changes to `iter` / streaming methods as well. specifically adding start/end agent run events.

then we can attach the agent name to those events so it looks something like this:

```
start agent run (name="main")
...
delegate tool call
  start agent run (name="research")
  ...
  end agent run (name="research")
end agent run (name="main")
```

Big benefit to this is that this will make all the UI adapter compatible with subagent streaming

pydantic_ai_harness/experimental/subagents/_toolset.py
```diff
@@ -0,0 +1,84 @@
...
+        """
+        agent = self._agents.get(agent_name)
+        if agent is None:
+            available = ', '.join(sorted(self._agents))
+            raise ModelRetry(f'Unknown sub-agent {agent_name!r}. Available sub-agents: {available}.')
+        toolsets = self._inherited_toolsets(ctx) if self._inherit_tools else None
+        capabilities = self._shared_capabilities or None
+        usage = ctx.usage if self._forward_usage else None
+        try:
+            result = await agent.run(
```
slmnsh, line 72: I just realized to propagate streaming we need to support yield from tool as well. I don't think we can do that right now.

pydantic_ai_harness/experimental/subagents/_capability.py
slmnsh: actually `SubAgent` capability could be singular as well.
```py
SubAgent(researcher, description=..., forward_usage=..., inherit_tools=...)
```
I think this will allow more granular control on each subagent. What do you think?

PR #271, opened 2026-06-04
feat: add optional ripgrep backend for FileSystem search

strawgate:
## Summary

Adds an optional ripgrep backend to `FileSystem.search_files`. When an `rg` binary is available (new `ripgrep` extra, or a system `rg`) content search delegates to ripgrep; otherwise the pure-Python walker runs — it stays the default and the behavioral reference.

- New `use_ripgrep` field: `None` auto-detects, `True` requires (raises if `rg` is missing), `False` forces pure Python.
- Both backends are confined to `root_dir`, apply the same dotfile / allow / deny / protected / `include_glob` filters, and share one formatter, so they return the same results for text files. ripgrep runs rooted, never follows symlinks, uses `--no-ignore` for parity, and each result is containment-checked before re-read.
- ripgrep runs via `anyio.run_process` (asyncio + trio) and its JSON is parsed with typed Pydantic models. A pattern ripgrep rejects but `re` accepts (e.g. lookaround) transparently falls back to pure Python.
- Only `search_files` uses ripgrep; the public surface is one field, the backend module is private.

**Known limitation:** binary detection differs — the pure-Python walker samples the first 8 KB for a NUL byte, ripgrep scans the whole file — so a file with a NUL past 8 KB is searched by pure Python but skipped by ripgrep (documented + tested; aligning them would change existing behavior).

## Validation

`make lint && make typecheck && make test` green; **100% branch coverage**, verified on slim (→ fallback) and `--all-extras`. Parity + behavior tests run real `rg`; protocol tests fake the subprocess to pin the command/JSON/exit-code handling the fallback would otherwise mask. Mutation-tested (`_ripgrep.py` + new `_toolset.py` functions in `[tool.mutmut]`). Adds a `ripgrep` extra (the `rg` binary) — `pyproject.toml` / `uv.lock` touched deliberately.

https://claude.ai/code/session_01A7oqgLJMQzovutMdywz2qk

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ Extras for specific capabilities:
 ```bash
 uv add "pydantic-ai-harness[codemode]"   # CodeMode (adds the Monty sandbox)
 uv add "pydantic-ai-harness[logfire]"     # ManagedPrompt (Logfire-managed prompts)
+uv add "pydantic-ai-harness[ripgrep]"     # FileSystem ripgrep-backed search (adds the rg binary)
 ```
 
 The `code-mode` extra is also supported as an alias.
--- a/docs/mutation-testing.md
+++ b/docs/mutation-testing.md
@@ -4,7 +4,8 @@ Mutation testing complements the 100% branch-coverage requirement: coverage
 proves every line and branch runs, mutation testing proves the assertions
 actually pin the behavior down.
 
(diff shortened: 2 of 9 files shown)


strawgate:
Closing this out; not carrying it forward.

PR #273, opened 2026-06-04
feat: Adds Dynamic Workflows

adtyavrdhn:
## What this is, in one minute

You have an agent that delegates work to other agents. Today that means one tool call per delegation: the orchestrator calls a sub-agent, waits, reads the result into its context, thinks, calls the next one. Ten sub-agent calls means ten model round-trips, and every intermediate result flows through the orchestrator's context window whether it needed to see it or not.

`DynamicWorkflow` replaces that loop with one tool, `run_workflow`. The model writes a small Python script in which each sub-agent is an async function, and the script runs to completion in a single tool call:

```python
import asyncio

reports = await asyncio.gather(
    reviewer(task="Review auth.py for bugs:\n<file contents>"),
    reviewer(task="Review parser.py for bugs:\n<file contents>"),
)
await summarizer(task="Summarize these findings:\n" + "\n\n".join(reports))
```

Two reviews run in parallel, the summarizer chains off them, and only the last line's value returns to the model. The intermediate reports never enter the orchestrator's context. Fan-out, chaining, voting, and retry loops are plain Python instead of model turns, so the deterministic coordination runs as code and the model is spent only on the parts that need a model.

Setup is two lines:

```python
from pydantic_ai import Agent
from pydantic_ai_harness.experimental.dynamic_workflow import DynamicWorkflow

reviewer = Agent('openai:gpt-5', name='reviewer', description='Reviews code for bugs.')
summarizer = Agent('openai:gpt-5', name='summarizer', description='Summarizes findings.')

orchestrator = Agent(
    'openai:gpt-5',
    capabilities=[DynamicWorkflow(agents=[reviewer, summarizer])],
)
```

Each agent's `name` and `description` become the function name and docstring the model sees. Each call is a full, isolated `Agent.run`: pass everything it needs in `task` (keyword-only). A sub-agent with `output_type=SomeModel` returns a `dict` the script reads by subscript.

This is the harness take on [dynamic workflows in Claude Code](https://claude.com/blog/introducing-dynamic-workflows-in-claude-code), built on the [Monty](https://github.com/pydantic/monty) sandbox the repo already uses for `CodeMode`.

## It merges with CodeMode

`CodeMode` already does the same trick for tools; this does it for sub-agents. Put both on one agent and they merge automatically: `run_workflow` disappears and each sub-agent becomes a typed async function inside the one `run_code` sandbox, next to the agent's regular tools. One script can then gather data with a tool, fan it out to sub-agents, and combine the results, in a single tool call:

```python
agent = Agent('openai:gpt-5', capabilities=[CodeMode(), DynamicWorkflow(agents=[reviewer, summarizer])])
```

```python
import asyncio
diff = await read_file(path="src/auth.py")                    # a regular tool
reviews = await asyncio.gather(                               # sub-agents, in parallel
    reviewer(task="Review for bugs:\n" + diff),
    reviewer(task="Review for style:\n" + diff),
)
await summarizer(task="Merge these reviews:\n" + "\n\n".join(reviews))
```

No new API for this: the capabilities detect each other. Budgets, usage forwarding, and the no-nesting rule apply identically in both modes. (Without the merge the two capabilities collided: CodeMode folded `run_workflow` into the sandbox as a function taking a code string, i.e. a script inside a script.)

## Safety and cost controls

- `max_agent_calls` (default 50): an exact, host-enforced ceiling on sub-agent runs per agent run. Exact even under concurrent fan-out, which `usage_limits` cannot guarantee (its check and increment are split by an await).
- `forward_usage=True`: the whole tree's token and request spend lands on the parent run's `usage`.
- `sub_agent_usage_limits`: a `UsageLimits` applied to every sub-agent run.
- `resource_limits`: sandbox memory/allocation caps with a safe backstop; `'unlimited'` to disable.
- Workflows do not nest: a sub-agent cannot start its own workflow or delegate further.
- Scripts are statically type-checked against the sub-agent signatures before running: a misspelled function or a positional `task` costs one retry, but no sub-agent budget.

## API

```python
DynamicWorkflow(                  # all parameters are keyword-only
    agents=[...],                 # Sequence[AbstractAgent | WorkflowAgent], required
    tool_name='run_workflow',
    max_agent_calls=50,
    max_retries=3,
    forward_usage=True,
    inherit_model=False,          # True -> sub-agents run with the parent run's resolved model
    sub_agent_usage_limits=None,
    resource_limits=None,
    id=None,                      # required when defer_loading=True
    description=None,
    defer_loading=False,
)

workflow.reveal(agent)            # add a sub-agent mid-run (append-only, cache-safe announcement)
WorkflowAgent(agent, name=..., description=...)  # rename/re-describe for one catalog
```

Full progressive tutorial: `pydantic_ai_harness/experimental/dynamic_workflow/README.md`.

## Relation to `SubAgents`

The experimental `SubAgents` capability delegates through one `delegate_task(agent_name, task)` tool, one model turn per delegation. `DynamicWorkflow` is for the case where the coordination itself has structure (fan-out, chaining, retry loops) that should run as code in one tool call rather than as model turns. The two shapes complement each other; converging them is a follow-up discussion. The capability README now has a "How this relates to SubAgents" section spelling out when to pick which.

## What is in this PR

- `DynamicWorkflow` capability and `DynamicWorkflowToolset` under `pydantic_ai_harness/experimental/dynamic_workflow/`.
- Automatic CodeMode merge (per-agent tools with return schemas, shared budget/guard path, guidance via instructions).
- `WorkflowAgent` (per-catalog rename/re-describe) and `reveal()` (append a sub-agent mid-run without moving the prompt-cache prefix).
- Budget and safety controls listed above, plus pre-execution static type checking of scripts.
- A shared sandbox executor (`pydantic_ai_harness/_monty_exec.py`) extracted so `code_mode` and `dynamic_workflow` drive Monty through one seam.
- Full test suite (`tests/experimental/dynamic_workflow/`, 100% branch coverage) covering catalog rendering, budgets (including exactness under concurrent fan-out and sharing across calls in one run), usage forwarding, reveal, deferred loading, the CodeMode merge end to end, and sandbox error surfaces.
- Standalone README written as a progressive tutorial, cross-linked from the CodeMode README.

### Dependency note

Adds a `dynamic-workflow` install extra pinning `pydantic-monty>=0.0.16`, which the repo already vendors for the `code-mode` extra: no new third-party dependency, only a new optional extra pointing at the existing sandbox. The "no changes to `pyproject.toml` or `uv.lock`" checklist item is intentionally unchecked for this reason.

### Placement: experimental

`DynamicWorkflow` ships under `pydantic_ai_harness.experimental`, per the repo convention for new capabilities. The deciding factor: the planned extensions (structured sub-agent inputs via a per-agent `parameters` schema, durable workflows from Monty snapshots) touch the sub-agent call contract itself, so the API needs room to move before the names go stable. Importing the capability emits `HarnessExperimentalWarning`; there is no top-level export. Promotion to top-level is a follow-up once those extensions settle.

## Linked Issue

Fixes #

## Checklist

- [ ] No changes to `pyproject.toml` or `uv.lock` (intentional: adds a `dynamic-workflow` extra pointing at the already-vendored `pydantic-monty`)


pydantic_ai_harness/experimental/dynamic_workflow/_toolset.py
```diff
@@ -0,0 +1,329 @@
...
+    raise ImportError(
+        'pydantic-monty is required for DynamicWorkflow. Install it with: pip install "pydantic-ai-harness[code-mode]"'
+    ) from _import_error
+
+from pydantic_ai_harness._monty_exec import MontyExecutor, PrintCapture
+
+# Set while a workflow script is executing, so a sub-agent that itself tries to run a workflow can
+# be refused — workflows do not nest. asyncio copies the context into each task `asyncio.gather`
+# schedules, so concurrently-dispatched sub-agents inherit this flag (the capability is asyncio-only).
+_in_workflow: contextvars.ContextVar[bool] = contextvars.ContextVar('pydantic_ai_harness_in_workflow', default=False)
```
adtyavrdhn, line 44: I don't think we need this, this can be checked by postinit in the capability to make sure it doesn't have the capability. 

We can give the capability an id, for the capability that can be added for run we need to check it at runtime

pydantic_ai_harness/experimental/dynamic_workflow/_toolset.py
```diff
@@ -0,0 +1,329 @@
...
+
+
+def _default_resource_limits() -> ResourceLimits:
```
adtyavrdhn, line 70: Why not do it in the class itself with a field or a def factory

pydantic_ai_harness/experimental/dynamic_workflow/_toolset.py
```diff
@@ -0,0 +1,329 @@
...
+    runaway script (e.g. `while True`) without penalising slow sub-agent calls.
+    """
+    return {
+        'max_duration_secs': 30.0,
+        'max_memory': 256 * 1024 * 1024,
+        'max_allocations': 50_000_000,
+    }
+
+
+class _WorkflowArguments(TypedDict):
```
adtyavrdhn, line 107: this doesn't need _ be the file itself is private

pydantic_ai_harness/dynamic_workflow/_toolset.py
```diff
@@ -0,0 +1,329 @@
...
+```
+
+The value of the script's last expression is captured as the result — do **not** `print()` it. Use
+`print()` only for debug logging.\
+"""
+
+
+def _render_agent_block(name: str, description: str | None) -> str:
+    """Render one sub-agent as the async function signature shown to the model."""
+    body = f'    """{description}"""' if description else '    ...'
```
adtyavrdhn: Not sure why this gap kinda weird looking

pydantic_ai_harness/dynamic_workflow/_toolset.py
```diff
@@ -0,0 +1,329 @@
...
+        for entry in self.agents:
+            name = entry.resolved_name
+            if not name:
+                raise UserError(
+                    'DynamicWorkflow sub-agent has no `name` and its agent has no `name`; '
+                    'set `WorkflowAgent(name=...)` so it can be exposed as a sandbox function.'
+                )
+            if not name.isidentifier():
+                raise UserError(
+                    f'DynamicWorkflow sub-agent name {name!r} is not a valid Python identifier, '
```
adtyavrdhn: How much of this is duplicated from the tool name checks for code_mode? Worth a look if we should refactor that to have this

pydantic_ai_harness/experimental/dynamic_workflow/_toolset.py
adtyavrdhn: Claude here: Keeping the `ContextVar` — it is the runtime guarantee. A `__post_init__` check cannot fully replace it: `AbstractAgent.root_capability` raises `NotImplementedError` on the base, and defer/runtime-loaded capabilities are invisible at construction. Left as-is.

pydantic_ai_harness/experimental/dynamic_workflow/_toolset.py
adtyavrdhn: Claude here: Keeping the function + `None` sentinel. A field `default_factory` would never fire — the capability always forwards `resource_limits=` explicitly, so `None` is the real 3-state signal (`None`=backstop, `{}`=off, dict=custom).

pydantic_ai_harness/experimental/dynamic_workflow/_toolset.py
adtyavrdhn: Claude here: Keeping the `_` prefix for consistency with `code_mode` (`_RunCodeArguments` etc.). Can sweep both files in a follow-up if we ever drop the convention.

pydantic_ai_harness/dynamic_workflow/_toolset.py
adtyavrdhn: Claude here: Fixed in cc1ec96 — replaced the hand-rolled f-string with core `FunctionSignature.render` (the same renderer `code_mode` uses).

pydantic_ai_harness/dynamic_workflow/_toolset.py
adtyavrdhn: Claude here: Done in cc1ec96 — catalog now renders via core `FunctionSignature` (the real dedup with `code_mode`). It also surfaced a bug: keyword names like `class` pass `isidentifier()` but are uncallable, so they are now rejected at construction.

adtyavrdhn:
Claude here: flagging a deliberate behaviour change in ed7f687 that reaches beyond this PR's new capability.

**Cancellation now waits for in-flight calls to unwind — in code_mode too.**

The shared Monty executor's cleanup previously fired `task.cancel()` on in-flight dispatches and returned immediately, leaving tool calls / sub-agent runs unwinding detached in the background (this is `main`'s code_mode behaviour; the path was `pragma: no cover`). It now awaits the cancelled tasks (`gather(..., return_exceptions=True)`) before propagating.

Observable effects:
- Cancelling a run (Ctrl-C, usage limit, outer timeout) returns only after in-flight tool/sub-agent calls have unwound — slightly later, but nothing executes past the point the call reported back. No more post-hoc mutation of the shared usage accumulator or "Task exception was never retrieved" warnings.
- Trade-off: a tool that swallows `CancelledError` and never exits now delays teardown instead of leaking silently. That's the structured-concurrency convention (TaskGroup/anyio/trio) — a hang surfaces the broken tool, a leak hides it.

Covered by `test_cancellation_awaits_inflight_sub_agents` and `test_cancellation_closes_unscheduled_coroutines`.

examples/dynamic_workflow.py
```diff
@@ -0,0 +1,266 @@
+"""Migrate a package from `os.path` to `pathlib` in a single `run_workflow` call, with a trace.
+
+This is the end-to-end `DynamicWorkflow` example: one orchestrator turn fans work out across many
```
adtyavrdhn: Delete this example, Claude came up with it and I am not convinced it is great. I'll write an example myself

examples/README.md
```diff
@@ -0,0 +1,137 @@
+# DynamicWorkflow example
+
+`DynamicWorkflow` gives the orchestrating model one tool, `run_workflow`. Instead of delegating to
+one sub-agent per model turn, the model writes a single Python script that calls the sub-agents as
+async functions — fan out with `asyncio.gather`, chain one into the next, loop until something
```
adtyavrdhn: Remove em dashes

examples/README.md
```diff
@@ -0,0 +1,137 @@
...
+converges — in one tool call. Only the script's final value returns to the model; the intermediate
+results stay in the sandbox.
+
+[`dynamic_workflow.py`](./dynamic_workflow.py) is a deliberately small package — three short files,
```
adtyavrdhn: Should link to the Anthropic blog up front

examples/README.md
```diff
@@ -0,0 +1,137 @@
...
+so the run is fast and cheap — shaped to exercise every coordination pattern end to end in a single
+`run_workflow` call: parallel fan-out, read/write confinement, an adversarial check, a feedback
+loop, typed fan-in synthesis, and a Logfire trace over the entire tree. Each sub-agent is a full
+Pydantic AI `Agent`, not a stand-in.
+
+## What it does
+
+The example plants a small package that still uses `os.path` into a fresh temp dir (so your repo is
```
adtyavrdhn: No need to talk about the example, only feature and how to use it

examples/README.md
```diff
@@ -0,0 +1,137 @@
...
+        break
+    # Migrate every pending file at once, then review each result at once.
+    reports = await asyncio.gather(*[migrator(task=tasks[path]) for path in pending])
+    reviews = await asyncio.gather(*[reviewer(task=r["path"]) for r in reports])
+
+    still_pending = []
+    for report, review in zip(reports, reviews):
+        path = report["path"]
+        outcomes[path] = {**report, "approved": review["approved"], "issues": review["issues"]}
+        if not review["approved"]:
```
adtyavrdhn: Explain this could have been done by the model but would have taken a turn

adtyavrdhn: And used the model for a deterministic step

examples/README.md
```diff
@@ -0,0 +1,137 @@
...
+            still_pending.append(path)
+    pending = still_pending
+
+# Fan in: one synthesizer turns every outcome into a single typed report.
+report = await synthesizer(task=json.dumps(list(outcomes.values())))
+report
+```
+
+Sub-agent outputs arrive in the script as plain dicts (`report["path"]`, `review["approved"]`), so
+the loop is ordinary Python end to end. Delegated one sub-agent per tool call, the same task would
```
adtyavrdhn: Way too much jargon

pydantic_ai_harness/_monty_exec.py
```diff
@@ -0,0 +1,199 @@
...
+                    cancelled.append(call)
+                else:
+                    call.close()
+            if cancelled:
+                # `cancel()` only schedules a `CancelledError` at each task's next suspension
+                # point; await them so dispatched work (e.g. sub-agent runs mutating shared
+                # usage) has fully unwound before this returns. `return_exceptions=True` keeps
+                # one task's teardown error from masking the original exception, and the
+                # results are deliberately discarded.
+                await asyncio.gather(*cancelled, return_exceptions=True)
```
adtyavrdhn, line 146: This is a change, the previous implementation did not await the cancels which left them leaking

adtyavrdhn: I don't think that was intentional, don't see a reason to not await these

pydantic_ai_harness/experimental/dynamic_workflow/_toolset.py
```diff
@@ -0,0 +1,526 @@
...
+from pydantic_ai.toolsets.abstract import SchemaValidatorProt, ToolsetTool
+from pydantic_ai.usage import UsageLimits
+from pydantic_core import to_jsonable_python
+from typing_extensions import TypedDict
+
+try:
+    from pydantic_monty import MontyRepl, MontyRuntimeError, MontySyntaxError, ResourceLimits
+except ImportError as _import_error:  # pragma: no cover
+    raise ImportError(
+        'pydantic-monty is required for DynamicWorkflow. Install it with: pip install "pydantic-ai-harness[code-mode]"'
```
adtyavrdhn: Installing monty with code mode is not great lol 

We should add it for this feature specifically too

pydantic_ai_harness/experimental/dynamic_workflow/_toolset.py
```diff
@@ -0,0 +1,526 @@
...
+
+from pydantic_ai_harness._monty_exec import MontyExecutor, PrintCapture
+
+# Set while a workflow script is executing, so a sub-agent that itself tries to run a workflow can
+# be refused — workflows do not nest. asyncio copies the context into each task `asyncio.gather`
+# schedules, so concurrently-dispatched sub-agents inherit this flag (the capability is asyncio-only).
+_in_workflow: contextvars.ContextVar[bool] = contextvars.ContextVar('pydantic_ai_harness_in_workflow', default=False)
+
+
+class WorkflowResourceLimits(TypedDict, total=False):
```
adtyavrdhn, line 47: Why is this a typed dict and not a dataclass?

pydantic_ai_harness/experimental/dynamic_workflow/_toolset.py
```diff
@@ -0,0 +1,526 @@
...
+# and render each catalog entry through core's `FunctionSignature` (the renderer code_mode and
+# Pydantic AI already use). This keeps the catalog format consistent across capabilities, forces
+# keyword-only `task` to match `dispatch` (which reads `kwargs['task']`), and renders docstrings
+# safely — a hand-rolled f-string breaks on a newline or a quote inside a description.
+_SUB_AGENT_PARAMS_SCHEMA: dict[str, Any] = {
+    'type': 'object',
+    'properties': {'task': {'type': 'string'}},
+    'required': ['task'],
+}
+_SUB_AGENT_SIGNATURE = FunctionSignature.from_schema(name='_', parameters_schema=_SUB_AGENT_PARAMS_SCHEMA)
```
adtyavrdhn: Don't know when Douwe refactored this stuff but love it

adtyavrdhn: Why is name _ ?

pydantic_ai_harness/experimental/dynamic_workflow/_toolset.py
```diff
@@ -0,0 +1,526 @@
...
+    return name.isidentifier() and not keyword.iskeyword(name)
+
+
+# Every sub-agent is exposed with the same fixed signature — `(*, task: str) -> Any` — so build it once
+# and render each catalog entry through core's `FunctionSignature` (the renderer code_mode and
+# Pydantic AI already use). This keeps the catalog format consistent across capabilities, forces
+# keyword-only `task` to match `dispatch` (which reads `kwargs['task']`), and renders docstrings
+# safely — a hand-rolled f-string breaks on a newline or a quote inside a description.
+_SUB_AGENT_PARAMS_SCHEMA: dict[str, Any] = {
+    'type': 'object',
```
adtyavrdhn, line 178: Should we have generated this with Pydantic instead?

pydantic_ai_harness/experimental/dynamic_workflow/_toolset.py
```diff
@@ -0,0 +1,526 @@
...
+                problem = (
+                    f'DynamicWorkflow sub-agent name {name!r} cannot be exposed as a sandbox function: '
+                    'it must be a Python identifier that is not a reserved keyword. Rename it.'
+                )
+            elif name in by_name:
+                problem = f'DynamicWorkflow has two sub-agents named {name!r}; names must be unique.'
+            else:
+                by_name[name] = entry
+                continue
+            if strict:
```
adtyavrdhn: Why are we not strict by default, we should have some rules and stick to it instead of maintaining strict no strict semantics

pydantic_ai_harness/experimental/dynamic_workflow/_toolset.py
```diff
@@ -0,0 +1,526 @@
...
+
+    async def for_run(self, ctx: RunContext[AgentDepsT]) -> AbstractToolset[AgentDepsT]:
+        """Fresh instance per run so the sub-agent-call budget is per-run.
+
+        `dataclasses.replace` would re-run `__post_init__`, whose strict validation raises on an
+        unusable entry that a mid-run append left in `agents` — turning one bad reveal into a hard
+        failure of every later run. Clone shallowly instead (keeping `agents` shared, so appends
+        stay visible to this run) and rebuild the per-run state leniently.
+        """
+        clone = copy.copy(self)
```
adtyavrdhn, line 488: Don't love it but I see there is some explanation by claude above, I should just try to refactor the thing to not run into it. I am worried there are incorrect decisions made here

pydantic_ai_harness/experimental/dynamic_workflow/_toolset.py
```diff
@@ -0,0 +1,526 @@
...
+            raise ModelRetry(f'Syntax error in workflow:\n{capture.prepend_to(e.display())}') from e
+        except MontyRuntimeError as e:
+            # Host-raised exceptions cannot be caught inside the sandbox (even a matching
+            # `except RuntimeError` aborts), so when this flag is set the budget error is
+            # the one that surfaced — it cannot be masking a later, unrelated failure.
+            if budget_exhausted:
+                # Retrying can't help — the per-run budget is spent. Return a terminal result
+                # so the model concludes instead of burning retries into a hard failure.
+                return {
+                    'error': (
```
adtyavrdhn: I'll merge `ToolFailed` soon and we can change this to use

pydantic_ai_harness/experimental/dynamic_workflow/_toolset.py
```diff
@@ -0,0 +1,526 @@
...
+                    )
+                }
+            raise ModelRetry(f'Runtime error in workflow:\n{capture.prepend_to(e.display())}') from e
+        except BaseException as e:
+            # pyo3 surfaces a Rust-side sandbox panic as `pyo3_runtime.PanicException` — a
+            # BaseException (not Exception) subclass that is not importable, so match it by name. A
+            # panic is an internal VM abort the model can provoke (e.g. awaiting the same sub-agent
+            # call twice in one `asyncio.gather`); convert it to a retry instead of letting it tear
+            # down the whole agent run. Anything else (CancelledError, ...) re-raises unchanged.
+            if type(e).__name__ != 'PanicException':
```
adtyavrdhn: ?

adtyavrdhn: Why does Code Mode not run into it? It should be similar

pydantic_ai_harness/experimental/dynamic_workflow/_toolset.py
```diff
@@ -0,0 +1,526 @@
...
+                raise
+            raise ModelRetry(
+                'The workflow script aborted inside the sandbox. This can happen when the same '
+                'sub-agent call is awaited more than once in one asyncio.gather — give each gathered '
+                'call its own invocation. Revise the script and try again.'
+            ) from e
+        finally:
+            _in_workflow.reset(in_workflow_token)
```
adtyavrdhn, line 721: It is set and reset but no one reads it?

README.md
```diff
@@ -32,7 +32,7 @@ uv add "pydantic-ai-harness[logfire]"     # ManagedPrompt (Logfire-managed promp
 
 The `code-mode` extra is also supported as an alias.
 
-Requires Python 3.10+ and `pydantic-ai-slim>=1.95.1`.
```
adtyavrdhn: I'll get rid of this bump from here and merge a separate one for it

examples/dynamic_workflow.py
adtyavrdhn: Claude here: Deleted in b225f47. The examples README was rewritten to lead with the feature, so the directory is ready for your hand-written example.

examples/README.md
adtyavrdhn: Claude here: The file had no literal em dashes; the `--` asides were the offenders and the rewrite in b225f47 splits most of them into plain sentences.

examples/README.md
adtyavrdhn: Claude here: Done in b225f47 -- the Anthropic dynamic-workflows post is now linked in the first sentence.

examples/README.md
adtyavrdhn: Claude here: Done in b225f47. The example narration is gone; the README now covers what the capability is and how to register sub-agents, and defers the full API to the capability README.

examples/README.md
adtyavrdhn: Claude here: Added in b225f47, right after the loop snippet: the model could run the same loop by delegating one sub-agent per tool call, but every iteration would cost a model round-trip, and each of those turns would spend the model -- a non-deterministic component -- on a step that is plain control flow (compare reviews, rebuild the task list, dispatch again).

examples/README.md
adtyavrdhn: Claude here: Rewritten in b225f47. The closing summary and its vocabulary ("typed fan-in synthesis", "cache-stable prompt", "hard worst-case token ceiling") are gone; the README now says what happens in plain terms and leaves the guarantees to the capability README.

pydantic_ai_harness/experimental/dynamic_workflow/_toolset.py
adtyavrdhn: Claude here: Added a `dynamic-workflow` extra in 8896285 (same `pydantic-monty>=0.0.16` pin); the ImportError messages and install docs now point at it. No `dynamicworkflow` alias: `dynamic_workflow` already normalizes to `dynamic-workflow` under PEP 685 -- the `codemode` alias only exists because "codemode" has no separator to normalize.

pydantic_ai_harness/experimental/dynamic_workflow/_toolset.py
adtyavrdhn: Claude here: Code Mode does run into it -- it was just unguarded. Reproduced: one async tool, `f = add(a=1, b=2); await asyncio.gather(f, f)` through `CodeModeToolset` panics the Monty VM (`pyo3_runtime.PanicException`), and since that is a `BaseException` nothing caught, it tore down the whole agent run. Fixed in 96ad121: `is_sandbox_panic` now lives in `_monty_exec`, both capabilities convert the panic to `ModelRetry`, and code_mode additionally drops its persistent REPL because the aborted VM's accumulated state cannot be trusted. Tests mirror the dynamic_workflow ones.

pydantic_ai_harness/experimental/dynamic_workflow/_toolset.py
adtyavrdhn: Claude here: It is read -- at the top of `call_tool` (the nested-workflow guard). The reader is a different toolset instance in a different agent's run: the parent sets the flag, `asyncio.gather` copies the context into each dispatched task, the sub-agent's run inherits it, and if that sub-agent has its own `DynamicWorkflow` and the model calls `run_workflow` there, that `call_tool` sees `True` and refuses. The set/reset and the read being the same function is why it looks unread.

pydantic_ai_harness/experimental/dynamic_workflow/_toolset.py
adtyavrdhn: Claude here: Same shape as pydantic-ai's `ModelSettings`: a `total=False` mapping of optional overrides merged onto defaults, so `{"max_memory": ...}` overrides only the cap it names. A dataclass needs `None`-sentinel defaults plus a field-by-field merge, and the caller imports a class to pass two integers; Monty's own `ResourceLimits` is also a TypedDict, so resolution stays a dict merge. The honest cost of the choice is the hand-rolled unknown-key check below -- a dataclass would reject typos for free. Both work; the `ModelSettings` precedent and dict-literal ergonomics tip it.

pydantic_ai_harness/experimental/dynamic_workflow/_toolset.py
adtyavrdhn: Claude here: `FunctionSignature.from_schema` requires a name, but the signature object is a module constant for the one fixed shape `(*, task: str)`; the real name is injected per agent at render time (`.render(..., name=name, ...)` two lines down). The `_` placeholder is never rendered anywhere.

pydantic_ai_harness/experimental/dynamic_workflow/_toolset.py
adtyavrdhn: Claude here: Could have (`TypeAdapter(...).json_schema()`, exactly what `_WorkflowArguments` does above), but deliberately not. `_WorkflowArguments` generates because its schema must stay in lockstep with a validator used at runtime. This schema is render-only input with no Python type it could drift from -- `dispatch` validates kwargs by hand regardless, since Monty does not enforce the schema -- and Pydantic generation would add `title` noise to a five-line constant that is itself the spec. Rule of thumb: generate when schema and validator must agree, hand-write when the schema is the single static artifact.

pydantic_ai_harness/_monty_exec.py
adtyavrdhn: Claude here: Confirmed on both counts. The old code_mode path cancelled and never awaited, and the `pragma: no cover` on it says the path was never deliberately exercised -- unintentional is the right read. It mattered little for tool calls but matters for sub-agents: an un-awaited cancelled task can still be mid-`Agent.run`, mutating shared usage. The one trade-off the await introduces: a task that swallows `CancelledError` now delays teardown instead of being abandoned.

pydantic_ai_harness/experimental/dynamic_workflow/_toolset.py
adtyavrdhn: Claude here: Agreed -- when `ToolFailed` lands, the budget-exhausted terminal dict here is the natural first adopter.

README.md
adtyavrdhn: Claude here: Resolved by the merge on 2026-06-12: the bump went in via #280 on main, main was merged back, and the branch no longer carries any bump diff (the conflicting pins were taken from main).

pydantic_ai_harness/experimental/dynamic_workflow/_toolset.py
adtyavrdhn: Claude here: There is a consistent rule, just not the one you want: construction errors raise at the developer, runtime mutations degrade with a warning. The lenient mode is forced by the mutation channel, not preferred -- reveal works by appending to a public list, so validation can only happen at the next consumption point, where raising would brick every later run of a shared agent because of one bad append made elsewhere. The fix is to remove the channel, not to pick a side: see the refactor sketch on the `for_run` thread below.

pydantic_ai_harness/experimental/dynamic_workflow/_toolset.py
adtyavrdhn: Claude here: Your instinct is right -- the duality is a symptom of reveal-by-list-append. Refactor that deletes it: make reveal an explicit method. `reveal(WorkflowAgent(...))` validates eagerly (name present, valid identifier, no collision) and raises `UserError` at the call site; `agents` becomes read-only after init (tuple internally); `WorkflowAgent` is frozen, so a validated entry stays valid forever, which means `_index_agents` can be strict everywhere and `strict=False`, `_reveal_warned`, and the `warnings` machinery all get deleted. The enqueued announcement still happens in `get_tools` on the next step (`reveal()` has no `RunContext`), so `_reveal_pending` shrinks to folding in already-validated entries. Two invariants to keep: `reveal()` stays sync and await-free (atomic under asyncio, same reasoning as the budget counter), and it validates against the master toolset's name set since per-run clones share the list. Cost: hosts call a method on the capability instead of appending to the list -- same reachability, better contract, and the capability is experimental so no compat debt.

examples/README.md
```diff
@@ -0,0 +1,112 @@
+# DynamicWorkflow
```
adtyavrdhn: Move this out from here

PR #275, opened 2026-06-06
Add ManagedTool / ManagedToolset for Logfire-managed tool framing

strawgate:
Back a tool's advertised **name**, **description**, and **per-parameter descriptions** with a Logfire managed variable, so you can retune how tools are framed to the model — versioned, labelled, rolled out — without redeploying. Implementation and schema *structure* stay in code.

**`ManagedTool`** — manages one tool. Pass a name to manage a tool registered elsewhere, or a `Tool` to both provide and manage it:

```python
ManagedTool('get_weather', label='production')        # by name
ManagedTool(Tool(get_weather), label='production')    # owns + manages
```

**`ManagedToolset`** — manages a group from one variable holding `dict[str, ManagedToolOverride]` (ergonomic for a whole MCP server / large toolset):

```python
ManagedToolset('weather', tools=[Tool(get_weather), Tool(get_forecast)], label='production')
```

`tool__get_weather` / `toolset__weather` resolve to overrides (patches; unset fields keep the code value). Schema *structure* (param names, types, required) stays in code so a remote value can't drift from the validator — only descriptions are tunable there. `ManagedToolOverride.name` is `min_length=1` so a UI-side `{"name": ""}` rollout raises ValidationError instead of producing an unnamed tool.

Built like `ManagedPrompt`: resolved once per run in `wrap_run`, baggage on every child span, per-run isolation via a context var. A shared `WrapperToolset` applies the overrides, re-keys renamed tools, and routes calls back to the original implementation (name swaps work; collisions raise `UserError`).

A `ManagedVariableCapability` base consolidates the per-run resolution plumbing now used by `ManagedPrompt`, `ManagedTool`, and `ManagedToolset` (companion commit refactors the test suite onto shared scaffolding in `tests/logfire_variables/conftest.py` + `_helpers.py`).

## Test plan

- [x] `make all` clean on rebased branch (format, lint, pyright strict)
- [x] 100% coverage including new `_managed_tool.py` / `_managed_toolset.py` / `_managed_variable.py`
- [x] Anchor tests assert warning `.filename` is the user's call site (catches `stacklevel` regressions)
- [x] Snapshot re-blessed for the new `capability_id` field upstream pydantic-ai added to `model_request_parameters` and the now-stripped `logfire.metrics` attribute


DouweM:
Superseded by https://github.com/pydantic/pydantic-ai-harness/pull/331

PR #283, opened 2026-06-12
feat(subagents): per-delegate run controls (usage limits, timeout, call budget, soft-failure)

### LINKED ISSUE #282: SubAgents: no per-delegate run controls (usage limits, timeout, call budget)
David's AICA here:

While migrating Pydanty (the team's dogfooding triage/PR agent) onto harness capabilities, we adopted `SubAgents` for a new delegation surface but could not use it to replace the orchestrator's existing delegation layer, because `SubAgentToolset.delegate_task` (`pydantic_ai_harness/experimental/subagents/_toolset.py`) runs the child with no per-delegate controls.

What the orchestrator's hand-rolled `run_delegate` (`agents/pydanty/pydanty_agent.py` in `pydantic/pydantic-ai-notes`) provides that `SubAgents` cannot express today:

1. **Per-delegate `usage_limits`**: `delegate_task` calls `agent.run(...)` without a `usage_limits` parameter. `forward_usage=True` shares the parent's usage *accounting*, but there is no way to give one child its own request budget (Pydanty caps its reproducer at 35 requests, its code librarian at 18).
2. **Per-delegate wall-clock timeout**: a hung child consumes the parent's whole deadline. Pydanty wraps each delegate in `asyncio.wait_for` with per-agent timeouts (240-600s); this machinery is what stopped a run-killing class of 1800s watchdog timeouts in production.
3. **Per-delegate call budgets**: at most N delegations per child per run (e.g. one reproducer call per issue), with a soft "budget exhausted, synthesize from existing evidence" tool result rather than an error.
4. **Soft-failure shaping**: on child failure/timeout, returning a steering message ("treat as recoverable, decide from existing evidence") instead of converting to a parent `ModelRetry` -- a retry invites the parent to immediately re-delegate the same way.

A possible shape: a per-agent config mapping (`limits: Mapping[str, SubAgentLimits]` with `usage_limits`, `timeout_seconds`, `max_calls`, optional `on_failure` message), falling back to current behavior when absent.

Items 1-2 look like the core blockers; 3-4 could arguably stay userland, but without 1-2 any production orchestrator with delegate budgets has to keep a parallel hand-rolled delegation path, which defeats adopting the capability.

dsfaccini:
Closes #282.

## Problem

`SubAgentToolset.delegate_task` runs each child with `agent.run(...)` and no per-delegate controls. `forward_usage=True` shares usage *accounting* but gives no child its own request budget; a hung child consumes the parent's whole deadline; there's no per-run delegation cap; and a child failure always becomes a parent `ModelRetry`, which invites immediate re-delegation. An orchestrator with per-delegate budgets (the team's Pydanty triage agent, whose hand-rolled `run_delegate` caps the reproducer at 35 requests / 600s and the librarian at 18 / 300s) could not drop its parallel delegation path.

## Change

New `limits: Mapping[str, SubAgentLimits]` on `SubAgents`. A name absent from `limits` keeps current behavior.

`SubAgentLimits` fields:

| Field | Effect |
|---|---|
| `usage_limits` | Per-delegation request/token budget. The child runs with **isolated** usage accounting so the budget counts only that child's usage (not the parent's or siblings'), even when `forward_usage=True`. Reaching it is a soft outcome, not a propagating `UsageLimitExceeded`. Tradeoff: that child's tokens don't aggregate into the parent's `usage`. |
| `timeout_seconds` | Per-delegation wall-clock budget. On exceed, the child run is cancelled and the parent gets a soft steering message. |
| `max_calls` | Max delegations to this child per parent run. Over budget returns a soft budget-exhausted message without running the child. Counts are keyed by `RunContext.run_id` and cleared in `wrap_run`. |
| `on_failure` | One steering message that overrides the built-in default for any soft outcome **and** makes child failures soft (a normal tool result instead of a parent `ModelRetry`). |

### Soft vs hard

A timeout, a reached per-child `usage_limits`, and an exhausted `max_calls` are always soft (steering message as a normal tool result). Child soft model errors (`ModelRetry` / `UnexpectedModelBehavior`) keep the legacy parent-`ModelRetry` behavior unless `on_failure` is set. A `UsageLimitExceeded` from a child with no per-child budget (shared accounting) still propagates -- the whole tree is out of budget.

## Why isolated accounting

With a shared `usage` object, `usage_limits` is checked cumulatively (verified), so a per-child request budget is only meaningful against the child's own accounting. This matches the Pydanty `run_delegate` reference, which passes `usage_limits` with no shared usage.

## Tests & docs

- `TestRunControls` covers isolation, soft usage-budget, propagating shared-limit, timeout, `max_calls` (incl. skipping the child and resetting across runs), and `on_failure` (failure + default override). 100% branch coverage on `_toolset.py` and `_capability.py`.
- README: new "Per-delegate run controls" section, rewritten "Failure handling", updated config block.
- `make lint && make typecheck && make test` green (1032 passed).

### DIFF
--- a/pydantic_ai_harness/experimental/subagents/README.md
+++ b/pydantic_ai_harness/experimental/subagents/README.md
@@ -63,9 +63,42 @@ print(result.output)
 - **Capabilities can be shared.** `shared_capabilities` are applied to every sub-agent run -- e.g. give all sub-agents a common guardrail, memory, or planning capability without rebuilding each `Agent`.
 - **Sub-agent events can be streamed.** Pass an `event_stream_handler` and it's forwarded to each sub-agent run, so the sub-agent's model-streaming and tool events surface to the caller (the handler receives the sub-agent's own `RunContext`).
 
+## Per-delegate run controls
+
+`limits` maps a sub-agent name to a `SubAgentLimits`, giving one delegate its own budgets without touching the others. A name absent from `limits` runs with the `SubAgents` defaults.
+
+```python
+from pydantic_ai.usage import UsageLimits
+from pydantic_ai_harness.experimental.subagents import SubAgentLimits, SubAgents
+
+orchestrator = Agent(
+    'anthropic:claude-opus-4-7',
+    capabilities=[
+        SubAgents(
+            agents={'reproducer': reproducer, 'librarian': librarian},
+            limits={
+                'reproducer': SubAgentLimits(usage_limits=UsageLimits(request_limit=35), timeout_seconds=600, max_calls=1),
+                'librarian': SubAgentLimits(usage_limits=UsageLimits(request_limit=18), timeout_seconds=300, max_calls=2),
+            },
+        )
+    ],
+)
+```
+
+| Field | Effect |
+|---|---|
+| `usage_limits` | A request/token budget for one delegation. The child runs with its own usage accounting, so the budget counts only that child's requests and tokens (not the parent's or siblings'), even when `forward_usage=True`. The tradeoff: that child's tokens no longer aggregate into the parent's `usage`. Reaching the budget is a soft outcome (see below), not a run-stopping `UsageLimitExceeded`. |
+| `timeout_seconds` | A wall-clock budget for one delegation. When the child exceeds it, its run is cancelled and the parent gets a soft steering message instead of hanging on the child. |
+| `max_calls` | The maximum number of delegations to this sub-agent per parent run. Once reached, further delegations return a soft budget-exhausted message without running the child. Counts are scoped to one `Agent.run` and cleared when it ends. |
+| `on_failure` | A steering message returned to the parent for any soft degradation of this delegate, in place of the built-in default. Setting it also makes child failures soft (see below). |
+
 ## Failure handling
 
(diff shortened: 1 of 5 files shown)


dsfaccini:
Claude here: ran an adversarial 3-reviewer pass (API breadth, correctness, docs/tests) before merge. Applied in the second commit:

- `SubAgents.__post_init__` now raises if `limits` names a sub-agent absent from `agents` (a typo'd key would otherwise silently run that delegate unbudgeted -- the exact failure budgets exist to prevent).
- `_call_counts` marked `compare=False` so mid-run delegation counts don't change `SubAgents` equality.
- Added a parallel-delegation test (two delegations in one parent step still capped at `max_calls`), plus README notes that `max_calls` is scoped per `run_id` (nested trees budget independently) and that a timed-out child's event stream is cut without a terminal event.

The correctness reviewer confirmed `usage_limits` is checked cumulatively against the passed `usage` object, so isolated child accounting is required for a per-child budget to mean the child's own requests -- the documented tradeoff (child tokens don't aggregate into the parent's `usage`) is accurate, not incidental.

Deferred as tracked follow-ups (out of scope for this PR):
- Optional usage fold-back: run the child with fresh usage for its own budget, then `ctx.usage.incr(result.usage())` so parent aggregation is preserved alongside the isolated limit.
- A `SubAgents`-level `default_limits` baseline applied to delegates without a per-name entry.
- Reconsider the `on_failure` name (reads like a callback; covers all soft outcomes, not just failures).

Issue #289, opened 2026-06-15
Capability for oversized tool returns: truncate / overflow-to-file / summarize (experimental)


dsfaccini:
Claude here: filing this after a design pass (drafted with David).

## Problem

A tool can return a payload large enough to dominate the context window. Tool
returns persist in history as `ToolReturnPart`, so an oversized payload is
re-sent on every later model request, paying its token cost for the rest of the
run. We want one capability that intercepts a return when it is produced and
reduces it, with the reduction persisted in history (not recomputed per
request).

This is the overflow-to-file follow-up the `experimental/compaction` README
already names as out of scope: it moves large tool outputs out of the window at
production time, rather than compressing or dropping context inside the window.

## Modes (all three, freely combinable)

1. Truncate/clamp: cut to a size limit (`head` / `tail` / `head_tail`). Lossy,
   no dependencies.
2. Overflow to queryable file: persist the full payload, replace the
   model-visible return with a handle + head/tail preview + a JSON-schema sketch
   for structured returns + a removed-amount marker, and expose a
   `read_tool_result(handle, offset, limit, from_end, pattern)` tool for
   on-demand slice / grep / read-from-end. Lossless (the Claude Code pattern,
   the core #4352 design).
3. Summarize: size-gated LLM summary of the payload. Default model is the
   agent's own, inherited from the run context (same pattern as
   `SummarizingCompaction`); override with a model id/instance OR a custom
   `summarize` callable. Summary usage folds into `ctx.usage`. No token caps on
   the summary call.

Combination via an ordered list of size bands `(threshold, action)`, first match
wins, passthrough below the smallest threshold. This expresses
summarize-then-spill, truncate-with-file-fallback, and
small=passthrough / medium=summarize / large=spill directly. Per-tool override
of the band list (file reads to `head`, logs to `tail`); a `tool_filter` to
scope which tools are touched. Default: lossless spill with a truncation
fallback (zero LLM cost, no surprise truncation).

## Mechanism

- Rewrite returns in `after_tool_execute` (`pydantic_ai/capabilities/abstract.py`).
  Fires once per tool call before the return enters history; `ctx` carries
  `tool_call_id`, `run_id`, `model`, `deps`. The rewrite persists, no
  per-request recompute.
- Register `read_tool_result` via `get_toolset()`. One capability does both. The
  read-back tool's own returns are exempt from re-processing (matched by tool
  name).
[...]

Issue #82, opened 2026-03-21
Tool Output Management (truncation, spill-to-file, per-tool summarization)


adtyavrdhn:
## Summary

A harness-level primitive that manages tool output before it enters the LLM context window. When tool outputs (shell commands, file reads, search results) exceed size thresholds, the harness truncates them using configurable strategies and optionally spills full output to the filesystem with retrieval hints so the agent can access complete data on demand.

This is **distinct from Compaction (#21)** which operates on conversation history. Tool Output Management operates on individual tool results at the moment they're returned, before they enter the conversation at all.

## Motivation

Every major coding harness has identified this as a critical primitive:

- **Codex** truncates tool outputs at ~10 KiB / 256 lines (head+tail strategy), with [active discussion](https://github.com/openai/codex/issues/6426) about replacing line-based with token-based limits
- **Claude Code** faces [context overflow from massive tool outputs](https://github.com/anthropics/claude-code/issues/12054) — it currently ingests all output without intelligent truncation
- **OpenDev** (arxiv 2603.05344) implements per-tool-type summarization: file reads return full content up to 4KB then summarize line ranges; search returns top 3 matches; commands return last 1KB
- **LangChain** describes "Tool Call Offloading" as a core harness primitive: "keeps the head and tail tokens of tool outputs above a threshold...offloads the full output to the filesystem so the model can access it if needed"
- **coding-agent-loop-spec** (strongdm/attractor) separates full output (for host app) from truncated output (for LLM), with per-tool configurable limits

Without this, a single `grep -r` or verbose test output can consume the entire context window.

### Truncation Strategies

| Strategy | Behavior | Best for |
|----------|----------|----------|
| `head_tail` | Keep first N + last M lines/tokens, mark middle as truncated | Shell output, logs |
| `tail_only` | Keep last N lines/tokens | Build output, test results |
| `head_only` | Keep first N lines/tokens | File reads, search results |
| `summarize` | LLM-summarize the output | Complex/structured output |
| `spill` | Write full output to file, return path + summary | Any large output |

### Per-Tool Configuration

```python
class ToolOutputPolicy:
    """Policy for managing a tool's output before it enters context."""
    max_tokens: int = 2000            # Token budget for this tool's output
    max_lines: int | None = None      # Optional line-based limit
    truncation_strategy: Literal['head_tail', 'tail_only', 'head_only', 'summarize', 'spill'] = 'head_tail'
[...]

PR #290, opened 2026-06-15
feat(experimental): OverflowingToolOutput capability for oversized tool returns

dsfaccini:
Claude here: implements #289 -- a capability that reduces oversized tool returns at production time so a large payload does not persist in history and get re-sent on every later model request.

## What it does

`OverflowingToolOutput` intercepts a tool return in `after_tool_execute`, reduces it once, and lets the reduced form persist (no per-request recompute). Three freely-combinable modes, selected by an ordered list of size `bands` `(over, action)`:

- `Truncate` -- head / tail / head+tail character clamp. Lossy, zero-cost.
- `Spill` -- persist the full payload, hand the model a handle + preview + shape sketch; full payload read back on demand via the registered `read_tool_result(handle, offset, limit, from_end, pattern)` tool. Lossless.
- `Summarize` -- size-gated LLM summary, inheriting the run's model by default (mirrors `SummarizingCompaction`); usage folds into `ctx.usage`, no token caps.

One capability does both the rewrite and the read-back tool; the read-back tool's own returns are exempt from re-processing (matched by name). Mirrors the `experimental/compaction` package shape, export pattern, README banner, and model-inherit path. Truncation / ANSI-strip / binary-detect helpers are harvested from #185.

## Adopted defaults for #289's open decisions (please weigh in)

- **Capability name**: `OverflowingToolOutput` (avoids colliding with #185's `ToolOutputManagement` and PR #185's name). Flagging for your confirmation.
- **Default band**: lossless `Spill(then=Truncate())` -- zero LLM cost, no surprise truncation, no silent drop.
- **Store**: narrow `OverflowStore` protocol + `LocalFileStore` default (per-run dir keyed by `run_id`); handle is a backend-addressable relative key, not an absolute path. Seam left to consume core #4352 / `ExecutionEnvironment` later.
- **Read-back tool**: ship our own `read_tool_result` now; delegate to core #4352 when it lands.
- **Combination model**: ordered bands, first match wins (auto-sorted largest-threshold-first), passthrough below the smallest threshold. `per_tool` **replaces** the global band list; `tool_filter` (`ToolSelector`) scopes which tools are touched.
- **Summarize**: `model=None` -> `ctx.model`; override via id/instance or a `summarize` callable; usage folded into `ctx.usage`; no token caps.
- **Size unit**: chars by default, `over_tokens=` opts into estimated tokens (reuses `compaction`'s `estimate_token_count` heuristic). `Truncate.max_chars` is always chars.
- **Cleanup**: spilled files kept by default (safe for later reads); durable backends own their own lifecycle.
- **Error/retry returns**: never spilled or summarized. `ModelRetry` and tool errors are raised, not returned, so they never reach `after_tool_execute`; an exception payload that does reach it is passed through.

## Edge cases covered

Binary (spill verbatim, never stringify-truncate; `Truncate`/`Summarize` on binary fall to `then`), already-small (passthrough), structured/nested (spill with a JSON shape sketch), multiple oversized returns in one step (distinct handle per `tool_call_id`), retries (distinct handle per `retry`), streaming (hook sees the assembled result), and read-back missing-handle / bad-pattern (`ModelRetry`).

## Relationship to prior art

- Supersedes the spill scope of PR #185 `ToolOutputManagement` (one-way, no read-back); harvests its truncation / ANSI / binary helpers.
- Closes the spill/summarize scope of #82.
- Consumes core pydantic-ai #4352 (canonical queryable-file primitive) through the `OverflowStore` seam once it lands.
- Distinct from #286 `ClampOversizedMessages` (clamps runaway model responses in history, not tool returns at production time) and from `experimental/compaction` (compresses/drops context already inside the window).

## Gates

`make lint && make typecheck && make testcov` all green; the new package is at 100% branch coverage (63 tests).

Implements #289 (kept open for the design-decision discussion above).

### DIFF
--- a/pydantic_ai_harness/experimental/overflow/README.md
+++ b/pydantic_ai_harness/experimental/overflow/README.md
@@ -1,0 +1,148 @@
+# Overflow capability
+
+> [!WARNING]
+> **Experimental.** This capability lives under `pydantic_ai_harness.experimental` and may
+> change or be removed in any release, without a deprecation period. Import it from the
+> experimental path -- there is no top-level export:
+>
+> ```python
+> from pydantic_ai_harness.experimental.overflow import OverflowingToolOutput
+> ```
+>
+> Importing any experimental capability emits a `HarnessExperimentalWarning`. Silence **all**
+> harness experimental warnings with a single filter (no per-capability lines needed):
+>
+> ```python
+> import warnings
+> from pydantic_ai_harness.experimental import HarnessExperimentalWarning
+>
+> warnings.filterwarnings('ignore', category=HarnessExperimentalWarning)
+> ```
+
+A tool can return a payload large enough to dominate the context window. Tool returns
+persist in history as `ToolReturnPart`s, so an oversized one is re-sent on every later
+model request -- paying its token cost for the rest of the run. `OverflowingToolOutput`
+intercepts a return in the `after_tool_execute` hook, reduces it once, and lets the reduced
+form persist. The reduction is not recomputed per request.
+
+This is the overflow-to-file follow-up the `compaction` README names as out of scope: it
+moves large tool outputs *out* of the window at production time, rather than compressing or
+dropping context already inside it.
+
+## The three modes
+
+| Mode | Cost | Lossy? | What the model gets |
+|---|---|---|---|
+| `Truncate` | zero-LLM | yes | A head / tail / head+tail clamp of the text |
+| `Spill` | zero-LLM | no | A handle + preview + shape sketch; full payload read back on demand |
+| `Summarize` | one LLM call | yes | A size-gated summary (inherits the run's model by default) |
+
+`Spill` is lossless: the full payload is persisted and the model reads slices of it through
+the registered `read_tool_result(handle, offset, limit, from_end, pattern)` tool (the Claude
+Code pattern, the core [#4352](https://github.com/pydantic/pydantic-ai/issues/4352) design).
+
+## Bands: combine the modes
+
+Configure an ordered list of size `bands`. Each band is a `(over, action)` pair: when a
+return's measured size reaches `over`, its action runs. The band with the largest threshold
+that fits wins; anything below the smallest threshold passes through.
+
+```python
+from pydantic_ai import Agent
+from pydantic_ai_harness.experimental.overflow import (
+    Band,
+    OverflowingToolOutput,
+    Spill,
+    Summarize,
+    Truncate,
+)
+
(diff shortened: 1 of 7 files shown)


dsfaccini:
Claude here: pushed the review fixes (commit 2e4f9f3). Summary of what changed and the two deferred items.

## Addressed (must-fix + cleanup feature)
- **S1 -- store hardening, sharing preserved.** `LocalFileStore` keeps a stable, configurable root (a later agent/run can still read an earlier spill -- isolation is *not* the security mechanism). The root is created `0700` (owner-only), and `read` resolves the target and rejects anything escaping the root via symlink, `..`, or an absolute path. `_safe_segment` sanitization kept.
- **C1 -- `ToolReturn.content` overflow.** Content is now measured and reduced with the same band logic as `return_value` (distinct spill handle). Non-text/multimodal content that overflows is left unreduced with a `warnings.warn`.
- **C2 -- `read_tool_result` bounded.** `offset >= 0`, `limit >= 1` clamped to a line cap, joined output capped, and `pattern` is now a literal substring (not a regex) so a model-supplied value cannot cause catastrophic backtracking.
- **A1 -- avoidable `Any` removed.** Typed `tokenizer`, narrowed payload/metadata handling, removed the `reportUnknownArgumentType` ignores via bool guards. (The hook's inherited `args`/`result: Any` stay.)
- **M1 -- summarize tests.** Now assert `ctx.model` inheritance, explicit-model override, and `usage=ctx.usage` threading with a `FunctionModel` instead of a wholesale `Agent` mock.
- **Cleanup feature.** `LocalFileStore(cleanup_after=timedelta | None=None)`; keep-forever default. When set, a `write` schedules a background daemon-thread prune by `st_mtime`; non-blocking and non-erroring (all failures -> `warnings.warn`, never into the run). `st_atime` noted as unreliable and unused. README documents the default, the opt-in TTL, and a standalone cron/mtime example.

`make lint && make typecheck && make testcov` green; overflow package at 100% branch coverage.

## Deferred (follow-ups, not in this commit)
- **C3 -- configurable read-back tool name.** `read_tool_result` is a fixed name; if a user already has a tool of that name it collides into a core `UserError`. Make the name configurable.
- **M2 -- tighten the `UsageLimits` README sentence.** The summary's nested run gets `usage=ctx.usage` but not the parent `usage_limits`, so a request limit accrues usage without being enforced on that call. The README line should say so precisely.

Name still `OverflowingToolOutput` (not renamed, per your note -- waiting on the final name).

PR #295, opened 2026-06-18
fix(code_mode): tighten `run_code` prompt against phantom imports and bad inputs

dsfaccini:
## Why

CodeMode agents repeatedly waste `run_code` turns on avoidable mistakes (observed across 3 recent pydanty runs, ~8 failures):

- importing unavailable modules (`textwrap`) or a phantom `functions` / `sandbox_tools` module
- pasting an oversized one-line literal → unterminated-literal `SyntaxError`
- indexing a typed tool result as a dict (`str indices must be integers`)
- echoing a `...` / `??` truncation placeholder (from a clamped result) back as an argument

Each failure burns a model turn and pushes agents toward their request limits.

## What

Prompt-only changes to `run_code`'s description:

- **Close the import allow-list:** state that anything outside the listed stdlib modules fails, and that the provided functions are already in scope (call by name, don't import from a module). This matches the existing `_functions_header` ("do not redefine or import them") but tells the model *where the names come from*.
- **Add a short input caution:** no oversized code literals (reference the variable holding a prior result), results are typed objects (read fields, don't `json.loads`/index a string), and never pass a `...`/`??` truncation placeholder as an argument.

Both are generically true for every CodeMode user; neither references a specific host or capability. The PR #262 os/pathlib notes are untouched.

## Verification

- `ruff` + `pyright` clean (pre-commit).
- `tests/code_mode/` (115 tests) pass; **100% branch coverage retained** on `_toolset.py` (prose-only, no new branches; no test pins the changed text).

### DIFF
--- a/pydantic_ai_harness/code_mode/_toolset.py
+++ b/pydantic_ai_harness/code_mode/_toolset.py
@@ -94,7 +94,7 @@ Write and run Python code in a sandboxed environment.
 The sandbox uses Monty, a subset of Python. Key restrictions:
 - **No classes**: class definitions are not supported
 - **No third-party libraries**: only the standard library modules listed below can be used
-- **Importable standard library modules**: `sys`, `typing`, `asyncio`, `math`, `json`, `re`, `datetime`, `os`, `pathlib`. These must be imported before use, just like in regular Python. For example: `import asyncio` then `results = await asyncio.gather(tool_one(...), tool_two(...))`."""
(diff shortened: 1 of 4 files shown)


pydantic_ai_harness/code_mode/_toolset.py
```diff
@@ -638,11 +657,16 @@ def _partition_callable_tools(
             callable_defs[safe_name] = td
         return callable_defs, sanitized_to_original
 
-    @staticmethod
-    def _build_description(callable_defs: dict[str, ToolDefinition], *, has_os: bool, has_mount: bool) -> str:
+    def _resolved_base(self, *, has_os: bool, has_mount: bool) -> str:
+        """The base prose: the `instructions` override if set, else the built-in default."""
+        if self.instructions is not None:
+            return self.instructions
```
adtyavrdhn: Why not allow adding onto the instructions instead of updating the entire thing?

We need to maintain a base which matches `monty` and if users start updating it they could reach out for issues which is being caused by changes in the prompt.

I would argue making sure `run_code` works seamlessly is our job.

PR #309, opened 2026-06-26
feat(logfire): add ManagedAgentSpec for whole-spec resolution

strawgate:
## Summary

Adds `ManagedAgentSpec` alongside the existing `ManagedPrompt` — a builder that resolves a Logfire-managed agent spec (`agentspec__<slug>`) and returns a ready-to-run `pydantic_ai.Agent` with instructions, model, model_settings, tools, MCP servers, skills, and capabilities all sourced from the spec.

```python
from pydantic_ai_harness import CodeMode
from pydantic_ai_harness.logfire import ManagedAgentSpec


def fetch_weather(city: str) -> str:
    return f'sunny and 72°F in {city}'


spec = ManagedAgentSpec(
    'checkout_assistant',
    tools={'tool__weather': fetch_weather},
    capability_classes={'code_mode': CodeMode},
)
agent = await spec.build()
result = await agent.run('What is the weather in Lisbon?')
```

The Logfire backend compiles `agentspec__*` variables server-side (this is the Platform-side counterpart to pydantic/platform#24682), so the SDK consumes the assembled payload directly — no ref-following on the client. Tool implementations stay in user code (function bodies aren't config); capability classes stay in user code so the harness extras stay optional (an `agentspec__` that references `code_mode` doesn't break callers who haven't installed it — the capability is silently dropped).

## Test plan

- [x] 18 new unit tests covering: defaults, prefix stripping, invalid names, prebuilt `Variable`, override flow with tools, unmapped/malformed entries, skill catalog folding, capability instantiation with/without config, unregistered-capability drop, bad-config `UserError`, valid MCP server reaching toolset kwargs (via lazy import + monkeypatch — the harness install doesn't include `[mcp]`), `logfire_instance` warning when paired with a prebuilt `Variable`.
- [x] `make all` clean: format, lint, pyright (strict), 100% coverage.

### DIFF
--- a/pydantic_ai_harness/__init__.py
+++ b/pydantic_ai_harness/__init__.py
@@ -5,13 +5,14 @@ from typing import TYPE_CHECKING
 if TYPE_CHECKING:
     from .code_mode import CodeMode
     from .filesystem import FileSystem
-    from .logfire import ManagedPrompt
+    from .logfire import ManagedAgentSpec, ManagedPrompt
     from .shell import LLM_API_KEY_ENV_PATTERNS, Shell
 
 __all__ = [
     'CodeMode',
     'FileSystem',
     'LLM_API_KEY_ENV_PATTERNS',
+    'ManagedAgentSpec',
     'ManagedPrompt',
     'Shell',
 ]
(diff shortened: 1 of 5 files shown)


DouweM:
Superseded by https://github.com/pydantic/pydantic-ai-harness/pull/331

PR #312, opened 2026-06-29
Emit a span when compaction happens

Kludex:
## Summary

The compaction strategies in `pydantic_ai_harness/experimental/compaction/` now emit an OpenTelemetry span named `compact_messages` whenever compaction runs.

The span carries these attributes:

- `compaction.strategy` -- the strategy that ran
- `compaction.messages_before` -- message count before compaction
- `compaction.messages_after` -- message count after compaction
- `compaction.tokens_before` -- token count before compaction
- `compaction.tokens_after` -- token count after compaction

When instrumentation is disabled the span emission is a no-op, so there is no behavior change for callers that do not configure tracing.

Tests cover the new span and its attributes at 100% branch coverage. The compaction README gains a "Tracing" subsection documenting the span name and attributes.

Slack thread: https://pydantic.slack.com/archives/C081LUCJ4KX/p1782749877395839?thread_ts=1782718531.844739&cid=C081LUCJ4KX

## Linked Issue

Fixes #

## Checklist

---
_Generated by [Claude Code](https://claude.ai/code/session_015kRjwKq2AmcuWZcDfJ4WHs)_

### DIFF
--- a/pydantic_ai_harness/experimental/compaction/_shared.py
+++ b/pydantic_ai_harness/experimental/compaction/_shared.py
@@ -125,7 +125,9 @@ def _history_changed(before: list[ModelMessage], after: list[ModelMessage]) -> b
     The same list object, or an equal-length list that compares equal element-wise, counts as
     unchanged; anything else is a change.
     """
-    return before is not after or before != after
+    # `!=` short-circuits on identity element-wise, so this also covers `before is after`; an
+    # unequal length already implies an unequal list, so a separate length check is redundant.
+    return before != after
 
 
 async def compact_with_span(
--- a/pydantic_ai_harness/experimental/compaction/_shared.py
+++ b/pydantic_ai_harness/experimental/compaction/_shared.py
@@ -6,7 +6,7 @@ preservation, and in-place tool-result clearing -- anything used by more than on
 
 from __future__ import annotations
 
-from collections.abc import Callable, Sequence
+from collections.abc import Awaitable, Callable, Sequence
 from dataclasses import dataclass, replace
 from typing import Protocol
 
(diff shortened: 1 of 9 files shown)


pydantic_ai_harness/experimental/compaction/_shared.py
```diff
@@ -110,6 +110,63 @@ def exceeds(
...
+    """Return True if *after* differs from *before*.
+
+    The same list object, or an equal-length list that compares equal element-wise, counts as
+    unchanged; anything else is a change.
+    """
+    return before is not after and (len(before) != len(after) or before != after)
+
+
+async def compact_with_span(
+    ctx: RunContext[AgentDepsT],
```
adtyavrdhn, line 134:
```suggestion
    ctx: RunContext[AgentDepsT],
    *,
    strategy: str,
    messages: list[ModelMessage],
    compact: Callable[[], Awaitable[list[ModelMessage]]],
    tokenizer: Callable[[str], int] | None = None,
```

tests/experimental/compaction/test_compaction.py
```diff
@@ -73,6 +83,7 @@ class _FakeCtx:
         usage: RunUsage
         model: Any = dataclasses.field(default_factory=_FakeModel)
         deps: None = None
+        tracer: Any = dataclasses.field(default_factory=NoOpTracer)
```
adtyavrdhn: This is typed in the Otel package I believe, let us use the type

tests/experimental/compaction/test_compaction.py
```diff
@@ -73,6 +83,7 @@ class _FakeCtx:
         usage: RunUsage
         model: Any = dataclasses.field(default_factory=_FakeModel)
```
adtyavrdhn: It should conform to our `Model` instead of `Any` and check if it should be added as a fixture?

Kludex: Claude here: Done. The fake ctx `model` field is now typed `Model` (backed by `TestModel`), and the `tracer` field is now typed `opentelemetry.trace.Tracer`. Both pushed.

On the fixture question: I kept it local rather than adding a conftest fixture. The fake context is built only through the single `_make_ctx` factory in this one test file, so there is no duplication across files. `_make_ctx` also takes runtime args (requests, input_tokens, output_tokens) that a plain fixture cannot supply, so it would have to stay a callable factory either way. Happy to promote it to a conftest factory fixture if you would prefer.

---
_Generated by [Claude Code](https://claude.ai/code)_

pydantic_ai_harness/experimental/compaction/_shared.py
```diff
@@ -110,6 +110,64 @@ def exceeds(
...
+        ctx: Run context whose `tracer` the span is started on.
+        strategy: Strategy name recorded in the `compaction.strategy` attribute.
+        messages: The pre-compaction messages, measured for the `*_before` attributes.
+        compact: Zero-argument async callable returning the compacted message list.
+        tokenizer: Optional tokenizer for the `compaction.tokens_*` estimates. When `None`,
+            uses the same ~4 characters-per-token heuristic as `estimate_token_count`.
+    """
+    compacted = await compact()
+    if not _history_changed(messages, compacted):
+        return compacted
```
adtyavrdhn: Consider returning messages here instead to make it clear that compaction was a no-op

adtyavrdhn (approved):

PR #319, opened 2026-07-03
Require explicit `CodeMode` final output wrapper

### LINKED ISSUE #6243: Expose a public capability API to commit validated tool results as final output
## Summary

Capabilities that orchestrate tools, such as a Code Mode capability, sometimes produce a structured object that already matches the agent final output schema. It should be possible for the capability to validate and commit that object as the final result without asking the model for another response whose only job is to copy the same data into an output tool call or text output.

## Use case

A Code Mode tool executes Python that calls other tools and returns a structured result. That result may contain exact IDs or other opaque values from tool responses. Requiring an extra model round trip to restate the result adds latency and cost, and it can corrupt or omit long identifiers.

The desired flow is:

1. Model calls a Code Mode function tool.
2. Code Mode returns a structured candidate final output.
3. A capability validates that candidate against the current agent output schema.
4. If it validates, the run ends immediately with that processed final output.
5. The transcript still records the Code Mode tool return, but there is no later model output-tool call or text response.

If the candidate does not validate, the graph should continue normally.

## Current gap

`AbstractCapability.before_output_process` receives `OutputContext`, so it can replace or normalize output cleanly. However, it only runs after the model has already produced a final output response.

`AbstractCapability.after_node_run` can return `End(FinalResult(...))`, so it can avoid the additional model turn. However, from the public capability API it does not expose a way to validate and process a candidate through the current output schema, output processor, and output validators.

The workaround requires private/internal access such as `agent._output_schema`, `agent._output_validators`, `run_output_process_hooks`, `FinalResult`, and graph `End`/node internals. That works, but it is a brittle boundary for a capability that otherwise fits the public hook model.

## Feature request

Please expose a public capability-facing API for committing a candidate final output from a tool or node hook. For example, an API that can:

- validate a Python candidate against the current output schema
- run the same output processing path and output validators used by normal model output
- preserve message history for completed tool returns
- return or create the final result without another model request

The shape could be a helper on `RunContext`, the tool manager, or a public output pipeline object, for example conceptually:

```python
processed = await ctx.output.process_candidate(candidate)
return ctx.output.end(processed)
```

or a single helper that returns a final-result node/object when validation succeeds.

The important contract is that capabilities can finish a run from a validated tool result without requiring the model to restate the same object.

kazmer97:
## Summary

Wires `CodeMode` to use the new `RunContext.output.try_commit_candidate(...)` API from pydantic/pydantic-ai#6247. When `run_code` produces a value that validates against the agent output schema, `CodeMode` commits it as the final output without requiring an extra model turn.

The normal flow is preserved for intermediate or invalid values, and the `run_code` tool return remains in message history.

## Linked Issue

Fixes pydantic/pydantic-ai#6243

Depends on pydantic/pydantic-ai#6247.

## Validation

- `uv run ruff check pydantic_ai_harness/code_mode/_capability.py tests/code_mode/test_code_mode.py`
- `PATH=/Users/kazmer.nagy-betegh/.nvm/versions/node/v24.14.1/bin:$PATH uv run --group lint --with-editable /Users/kazmer.nagy-betegh/dev/pydantic-ai/pydantic_ai_slim pyright pydantic_ai_harness/code_mode/_capability.py tests/code_mode/test_code_mode.py`
- `uv run --with-editable /Users/kazmer.nagy-betegh/dev/pydantic-ai/pydantic_ai_slim pytest tests/code_mode/test_code_mode.py::TestCodeMode::test_run_code_result_can_be_committed_as_final_output tests/code_mode/test_code_mode.py::TestCodeMode::test_code_mode_via_agent_run_executes_run_code_and_returns_result tests/code_mode/test_code_mode.py::TestDynamicCatalog::test_for_run_returns_fresh_state_when_dynamic_catalog_disabled tests/code_mode/test_code_mode.py::TestToolSearchIntegration::test_tool_search_toolset_discovered_tool_in_run_code tests/code_mode/test_code_mode.py::TestDynamicCatalog::test_announce_on_native_search_return_part`
- `git diff --check`

## Checklist

- [ ] `make lint && make typecheck && make test` passes locally (focused checks listed above)

### DIFF
--- a/pydantic_ai_harness/code_mode/README.md
+++ b/pydantic_ai_harness/code_mode/README.md
@@ -150,6 +150,12 @@ The last expression in the code snippet is automatically captured as the return 
 | With print output | `{"output": "<printed text>", "result": <last expression>}` |
 | Multimodal content (e.g. images) | Returned natively for model processing |
 
+If the captured value is valid for the agent's output schema, `CodeMode` commits it as the final
+output immediately. The `run_code` tool return still appears in message history, but there is no
+extra model turn whose only job is to restate the same value. When printed output is present,
+`CodeMode` validates the `result` value from the `{"output": ..., "result": ...}` wrapper.
+Invalid or intermediate values continue through the normal next-model-turn flow.
+
 ## REPL state
 
 State persists between `run_code` calls within the same agent run -- variables, imports, and function definitions carry over. Pass `restart: true` in the tool call to reset state.
--- a/pydantic_ai_harness/code_mode/README.md
+++ b/pydantic_ai_harness/code_mode/README.md
@@ -150,11 +150,14 @@ The last expression in the code snippet is automatically captured as the return 
 | With print output | `{"output": "<printed text>", "result": <last expression>}` |
 | Multimodal content (e.g. images) | Returned natively for model processing |
 
-If the captured value is valid for the agent's output schema, `CodeMode` commits it as the final
-output immediately. The `run_code` tool return still appears in message history, but there is no
-extra model turn whose only job is to restate the same value. When printed output is present,
-`CodeMode` validates the `result` value from the `{"output": ..., "result": ...}` wrapper.
-Invalid or intermediate values continue through the normal next-model-turn flow.
+To commit a value as the agent's final output immediately, make the last expression exactly
+`{"final_output": <value>}`. The `run_code` tool return still appears in message history, but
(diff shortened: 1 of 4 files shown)


DouweM:
Marking as draft per https://github.com/pydantic/pydantic-ai/issues/6243#issuecomment-4879916248.

DouweM:
Superseded by https://github.com/pydantic/pydantic-ai-harness/pull/323

PR #327, opened 2026-07-06
feat(`cache_stability`): observational `CacheStabilityMonitor` capability

dsfaccini:
## What

Adds `CacheStabilityMonitor`, a private/experimental capability that warns when a run's prompt cache hit collapses between model requests -- making a moving cacheable prefix visible instead of silently re-charging tokens.

This is the **observe** arm of the cache-prefix-stability work (proposal §10b / §11 arm 4). It reads the provider's own ground-truth verdict, `response.usage.cache_read_tokens`, rather than fingerprinting the structured request -- a prototype proved the structured signal false-positives on internal metadata that serialization strips (`run_id`, `defer_loading`, ...) and is blind to serialization-level busts (0/44 flagged busts were real).

## How

- `for_run` resets a per-run **high-water mark** of the established cacheable prefix.
- `after_model_request` reads `cache_read_tokens`; warns (`CacheBustWarning(UserWarning)`) when a later request reads back less than `collapse_ratio` of the established prefix, gated by `min_prefix_tokens`, then updates the mark to `max(established, read + write)`.
- Cross-provider for free via pyai's normalized `cache_read_tokens` / `cache_write_tokens`.
- Silent when caching is off or unreported -- never fires spuriously in runs that don't cache.

## Silencing / escalation

No bespoke suppression API. Users manage it with stdlib `warnings.filterwarnings(...)` (and `'error'` for dev/CI enforcement), exactly as they'd manage any other `UserWarning`.

## Options (conservative defaults, user-tunable)

- `collapse_ratio=0.5` -- warn when a request reads back less than this fraction of the established prefix.
- `min_prefix_tokens=1024` -- only judge once the established prefix reaches this size (Anthropic's minimum cacheable size).

- **Structured signal A dropped** -- shipped signal is observational only, per proposal §9-§10.
- **High-water mark** rather than only the prior request -- handles OpenAI's missing write counter and append-only growth.
- **No cost-via-genai-prices in v1** -- the warning reports wasted tokens, not dollars; cost is additive and deferred per §10.
- **Private / experimental** -- lives in `experimental/cache_stability`, not re-exported from top-level `pydantic_ai_harness` until the team commits public surface.
- **Wire-level VCR assertion and `diag.py` wire-diff are out of scope** -- this is the *trigger*, kept separate from the wire-level *explanation* (§11).

## Tests

6 tests driven through `Agent(..., capabilities=[...])` with a `FunctionModel` emitting preset `RequestUsage` per step: collapse warns, stable prefix stays silent, sub-`min_prefix` drop ignored, tunable thresholds catch a smaller regression, `'error'` filter escalates, and `for_run` isolates runs. 100% branch coverage on the module; full suite green (1392 passed / 10 skipped), lint + pyright-strict clean.


pydantic_ai_harness/cache_stability/_capability.py
```diff
@@ -0,0 +1,136 @@
...
+    it never fires spuriously in tests that don't exercise caching. Silencing and dev/CI
+    escalation both go through the stdlib `warnings` filters -- see `CacheBustWarning`.
+    """
+
+    collapse_ratio: float = 0.5
+    """Warn when a request reads back less than this fraction of the established prefix.
+
+    Conservative by default (0.5): only a drop below half the previously-cached prefix counts
+    as a collapse, so ordinary provider rounding or a partial cache miss does not fire. Lower
+    it toward 1.0 to warn on smaller regressions.
```
DouweM: This is backwards: 1.0 is *higher* than the 0.5 default, and raising toward 1.0 makes the monitor *more* sensitive. The README's Options section has it right ("raise toward `1.0`") — `Lower` → `Raise` here.

pydantic_ai_harness/experimental/cache_stability/_capability.py
```diff
@@ -0,0 +1,136 @@
...
+        self._step += 1
+        usage = response.usage
+        read = usage.cache_read_tokens
+        established = self._established
+        if established >= self.min_prefix_tokens and read < established * self.collapse_ratio:
+            wasted = established - read
+            warnings.warn(
+                f'Cache hit collapsed at model request {self._step}: read {read} cached tokens but '
+                f'a prior request established ~{established} (~{wasted} tokens re-sent uncached). '
+                f'The cacheable prefix moved between requests.\n\n'
```
DouweM: "The cacheable prefix moved between requests" overclaims the cause: a `cache_read_tokens` collapse also happens when the provider-side cache **expired** under a byte-identical prefix — Anthropic's default TTL is 5m, so any tool call or human-approval gap longer than that triggers this warning while the prefix never moved. Please hedge the message ("…the cacheable prefix moved between requests, or the provider's cache expired (e.g. a gap longer than the cache TTL)"). Optionally: track the wall-clock gap since the previous `after_model_request` in the per-run state and include it in the message when it exceeds a conservative floor — that keeps cause *attribution* out of scope while avoiding *mis*attribution.

pydantic_ai_harness/experimental/cache_stability/_capability.py
```diff
@@ -0,0 +1,136 @@
...
+                f'To silence or escalate:\n\n{_SILENCE_HINT}\n',
+                CacheBustWarning,
+                stacklevel=2,
+            )
+        self._established = max(established, read + usage.cache_write_tokens)
```
DouweM: The high-water mark is global to the run, but a `FallbackModel` failing over mid-run (or an explicit per-step model switch) changes provider — the new provider's cache is necessarily empty, so this warns on every legitimate model switch. Key the mark per `(response.provider_name, response.model_name)` (a dict in the per-run state; the response fields are the only truthful source since `FallbackModel` resolves the model inside `request()`), and don't warn when the key changes. Keeping per-key marks rather than resetting also handles a switch *back* within the old provider's TTL, which can still hit that provider's cache. Please add a two-model `FunctionModel` test for the failover case.

DouweM (commented):
Douwe's AICA here, relaying his review. Douwe likes this as a harness capability and wants to keep it — it stays the opt-in observe arm even as parts of the detection logic later graduate into core instrumentation (tracking issue on the core side coming shortly). Three things to address before merge, in the inline comments.

One more, non-blocking: Logfire won't pick up these warnings on its own (it bridges stdlib *logging*, not the `warnings` module). For the experimental v1, warnings-only is fine, but please add a README note pointing users at `logging.captureWarnings(True)` so busts can reach their traces. Emitting an explicit Logfire event from the capability can wait until this graduates.

dsfaccini:
on it

tests/cache_stability/test_cache_stability.py
```diff
@@ -0,0 +1,218 @@
...
+
+    def fn(messages: list[ModelMessage], info: AgentInfo) -> ModelResponse:
+        i = state['i']
+        state['i'] += 1
+        return responses[i]
+
+    def noop() -> str:
+        return 'ok'
+
+    return Agent(FunctionModel(fn), deps_type=type(None), capabilities=[monitor], tools=[noop])
```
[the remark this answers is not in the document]
dsfaccini: This reply was posted by Claude Code using claude-opus-4-8 on behalf of David (who has not reviewed this specific reply).

Keeping `FunctionModel` here -- dismissing this one. The monitor's behavior is driven entirely by fields the test must set per step on each `ModelResponse`: `usage.cache_read_tokens` / `cache_write_tokens` (the collapse signal), `provider_name` / `model_name` (the per-key high-water mark), and `timestamp` (the TTL-gap message). `TestModel` emits canned responses and doesn't let a test inject arbitrary per-request `RequestUsage` or vary provider/model/timestamp per step, so it can't exercise the collapse, model-switch, or expiry-gap paths at all.

`FunctionModel` is a first-party model from `pydantic_ai.models.function`, and these tests drive it through the public `Agent(..., capabilities=[...])` surface, which matches the repo's testing guidance. "Extend core support" (teaching `TestModel` to emit preset cache usage) would be a `pydantic-ai` core change -- out of scope for this harness PR under the repo's core-boundary rule; we'd propose it separately rather than reimplement it here.

dsfaccini:
@coderabbitai review

pydantic_ai_harness/cache_stability/_capability.py
```diff
@@ -0,0 +1,178 @@
...
+    ) -> ModelResponse:
+        """Compare this response's cache read against the established prefix for its model, then update it."""
+        self._step += 1
+        usage = response.usage
+        read = usage.cache_read_tokens
+        key = (response.provider_name, response.model_name)
+        established = self._marks.get(key, 0)
+        # 0.0 on the run's first request, which has no predecessor to measure a gap against.
+        gap = 0.0 if self._last_time is None else (response.timestamp - self._last_time).total_seconds()
+        self._last_time = response.timestamp
```
DouweM: **MEDIUM — the marks are per key, but `_last_time` is global, so the TTL hedge measures the gap against the wrong predecessor.**

`_marks` is keyed per `(provider_name, model_name)` (correctly), but `_last_time` is a single value updated on every response. After a switch away and back, the gap is measured from the *other* model's request, which withholds the expiry explanation exactly when it's most likely to be the answer. Traced with a 300s assumed TTL:

```
R1  A  t=0    write=8000  -> silent, marks[A] = 8000
R2  B  t=250  read=0      -> silent (fresh key)      _last_time = 250
R3  A  t=400  read=100    -> WARN, gap = 400-250 = 150s -> generic hedge only
```

A's cache was last touched at t=0, so at t=400 it is **400s old — past the TTL**, and expiry is the most probable cause of the collapse. But the monitor reports 150s and falls back to the generic hedge, which is the misattribution the original comment was about. It skews the other way too: a long gap on B before a quick return to A can name a TTL-expiry gap for a cache that was just refreshed.

Fix: key the timestamps alongside the marks (`dict[_CacheKey, datetime]`), reset in `for_run` with the rest, and compute the gap from the previous response *for the same key*. That pairs naturally with capturing a local observation time instead of `response.timestamp` (see the summary), which also sidesteps cross-provider clock skew — the two fixes touch the same field.

pydantic_ai_harness/cache_stability/_capability.py
```diff
@@ -0,0 +1,178 @@
...
+                expiry = ' (e.g. a gap longer than the cache TTL)'
+            warnings.warn(
+                f'Cache hit collapsed at model request {self._step}: read {read} cached tokens but '
+                f'a prior request established ~{established} (~{wasted} tokens re-sent uncached). '
+                f"The cacheable prefix moved between requests, or the provider's cache expired{expiry}.\n\n"
+                f'To silence or escalate:\n\n{_SILENCE_HINT}\n',
+                CacheBustWarning,
+                stacklevel=2,
+            )
+        self._marks[key] = max(established, read + usage.cache_write_tokens)
```
DouweM: **MEDIUM — a `0/0` response after an established prefix warns on *every* remaining request.**

Once a mark exists, a response reporting neither reads nor writes trips the collapse check — and because the mark only ever grows (`max(established, read + write)`), it trips again on every subsequent request:

```
R1  caching on   (read=0, write=8000):  silent   -> mark = 8000
R2  caching off  (read=0, write=0)   :  WARN
R3  caching off  (read=0, write=0)   :  WARN     <- and every request after
```

The reachable trigger is the one your own earlier note called out: `_build_automatic_cache_control` only emits breakpoints when `anthropic_cache` / `anthropic_cache_messages` is set, so a harness or user toggling caching off mid-run gets a warning per remaining step. A prompt dropping below the minimum cacheable size does the same.

Worth knowing: this is the exact ambiguity we just worked through on the core side in pydantic/pydantic-ai#6534. The conclusion there was that `0/0` is **undisambiguatable from usage data alone** — on providers that report writes (Anthropic, Bedrock) it means the cache wasn't engaged; on providers that only report reads (OpenAI's implicit caching) it's a full miss. We first tried inferring the provider's reporting style at runtime and it broke (a run whose first cache-participating response is a pure hit never observes a write), so core now records the waste under a distinct `unreported` classification and never alerts on it.

The calculus isn't identical here — this monitor is opt-in and its whole point is visibility, so surfacing real waste is more defensible than it would be for a default-on alert. But the **repetition** I'd fix either way: one warning per collapse, not one per remaining step. Two options, your call:

- Re-baseline the mark when a collapse fires (`self._marks[key] = read + usage.cache_write_tokens`), which also fixes the general case of a deliberate bust re-reporting against a stale high-water mark on every later request.
- Or skip judgment entirely when `read == 0 and write == 0`, matching #6534.

DouweM (commented):
Douwe's AICA here, reviewing the updates. **All four asks are addressed**, and I checked the tests against the *old* behavior rather than taking the commit messages' word for it:

**(1) `collapse_ratio` docstring** — fixed (`Raise it toward 1.0`).

**(2) Warning overclaims the cause** — hedged, and the optional gap tracking landed too. One flaw in the new timing code (inline below).

**(3) Model-switch false positives** — fixed correctly, and the tests genuinely discriminate. `test_model_failover_does_not_warn` establishes 8000 tokens on model A *before* failing over, so the old global mark would warn and fail the test; `test_switch_back_within_ttl_uses_preserved_mark` would go silent under a reset-on-switch design. That's the right pair, and keying on the *response's* `(provider_name, model_name)` is the correct source under `FallbackModel`.

**(README)** — `logging.captureWarnings(True)` documented with the `py.warnings` → Logfire path spelled out. 

Two inline findings below (one from a second reviewer pass, one traced by hand), plus smaller things:

- **LOW** — `collapse_ratio=0.0` passes validation but makes warnings impossible (`read < established * 0` is never true for non-negative `read`), so it's an undocumented disable switch. Consider `0.0 < collapse_ratio <= 1.0`, with boundary tests for both ends.
- **LOW** — `cache_ttl_seconds=0.0` passes validation and then describes essentially every inter-request gap as past the assumed TTL. Require strictly positive unless zero is a documented mode.
- **LOW** — the gap derives from `ModelResponse.timestamp`, which some providers populate from their own response payload. Across a provider switch those clocks can disagree, and a skewed pair can print a negative `~{gap}s earlier`. Capturing a local observation time when the response is seen (paired with the per-key fix below) avoids trusting provider clocks at all. It also removes the smaller distortion that response-to-response spans include this request's own generation time.
- **NIT** — `_marks` / `_step` / `_last_time` are init-able dataclass fields, so `CacheStabilityMonitor(_marks={...})` is legal. That was cosmetic while the module was experimental; if it graduates (below), it's public constructor surface and would be better as a non-init state object built in `for_run`.
- **Cross-track, non-blocking** — `cache_ttl_seconds=300.0` is a hardcoded assumption that Pydantic AI is about to be able to answer properly: pydantic/pydantic-ai#6337 adds `ModelProfile.prompt_cache_retention` (Anthropic 5m, GPT-5.6 30m, and deliberately unset where the default depends on account configuration rather than the model — OpenAI's ZDR-dependent case). Since this monitor is already keyed per model, reading the profile beats one global default. Worth a TODO now and a follow-up once that ships.

---

**One thing for Douwe rather than you:** commit 94396f937 graduates the module out of `experimental/` — top-level path, `warn_experimental` dropped, advertised in the root README matrix. That reverses the explicit 2026-07-02 decision for this work ("private for now — no committed public surface until the team discusses"), and Douwe's steer this round was that it belongs in the harness rather than core, which is about *home*, not about promoting it to committed surface. It also introduces a status the matrix doesn't otherwise use: every other row is either plain (`code_mode/`, `filesystem/`, `shell/`) or `(experimental)` for things under `experimental/` (ACP); `(preview)` at a top-level path is a third thing. I'd keep it under `experimental/` until the team commits, or graduate it properly — but not split the difference. Douwe should call it.

(The duplicate Managed-prompt row cleanup checks out — `main` really does have two.)

pydantic_ai_harness/cache_stability/_capability.py
dsfaccini: This reply was posted by Claude Code using claude-opus-4-8 on behalf of David.

_David wrote the handoff this implements but has not code-reviewed the resulting diff yet (he may be away) -- treat it as lightly steered, not reviewed._

Fixed. Per-key observation state now lives in `_RunState.keys: dict[_CacheKey, _KeyState]`, where each `_KeyState` holds the established `prefix` and a `seen_at` time. The gap is computed from the *same key's* previous `seen_at`, so a switch-away-and-back times the returning model's own idle window. Your 300s trace now names A's own 400s gap (`~400s earlier, past the assumed ~300s cache TTL`) instead of B's 150s.

The time comes from a monotonic `_now()` seam rather than `response.timestamp`, so cross-provider clock skew can't produce a negative gap and the span no longer folds in this request's own generation latency (your third LOW). Regression: `test_expiry_gap_measured_per_key_after_switch_away_and_back` asserts the concrete 400s gap, which the old global `_last_time` would have reported as 150s + the generic hedge.

pydantic_ai_harness/cache_stability/_capability.py
dsfaccini: This reply was posted by Claude Code using claude-opus-4-8 on behalf of David.

_David wrote the handoff this implements but has not code-reviewed the resulting diff yet (he may be away) -- treat it as lightly steered, not reviewed._

Fixed with the re-baseline option: on a collapse the mark resets to the collapsed prefix (`read + write`) instead of `max(established, ...)`, so a sustained collapse -- including a 0/0 caching-off tail -- warns once, not per step. It also fixes the general case you noted of a deliberate bust re-reporting against a stale high-water mark.

I kept the single surfacing rather than skipping `0/0` entirely: the monitor is opt-in and visibility is its whole point, so reporting the first collapse is defensible, while the repetition (your actual ask) is gone. Aligns with the "one warning per collapse" bar and core #6534's read of `0/0` as undisambiguatable. Regression: `test_caching_off_mid_run_warns_once_not_per_step` (R1 establishes 8000, R2/R3 report 0/0; asserts exactly one `CacheBustWarning`, at request 2).

dsfaccini:
This reply was posted by Claude Code using claude-opus-4-8 on behalf of David.

_David wrote the handoff this implements but has not code-reviewed the resulting diff yet (he may be away) -- treat it as lightly steered, not reviewed._

Pushed d97d2e6 addressing the second review. The two MEDIUM findings are answered in their inline threads; the rest:

- **`collapse_ratio=0.0`** -- now rejected: validation requires `0.0 < collapse_ratio <= 1.0`. Boundary tests cover 0.0 (rejected) and 1.0 (accepted).
- **`cache_ttl_seconds=0.0`** -- now rejected: requires strictly positive. Boundary test added.
- **Provider-clock gap (LOW)** -- the gap now derives from a monotonic `_now()` seam captured when the response is seen, not `response.timestamp`; details in the per-key thread.
- **`_marks`/`_step`/`_last_time` init surface (NIT)** -- per-run state moved into a single `field(init=False)` `_RunState`, rebuilt fresh in `for_run` via `replace(self)`. `CacheStabilityMonitor(_state=...)` now raises `TypeError`; `test_per_run_state_is_not_constructor_surface` guards it.
- **`prompt_cache_retention` (#6337, cross-track)** -- added a `TODO(#6337)` at `cache_ttl_seconds` and a note in the README/docs. Left as a follow-up until it ships, since the monitor is already keyed per model.

**On the graduation fork you flagged for David:** David called it -- graduate *properly* as a released capability, not split the difference. So the `(preview)` framing is dropped everywhere, and the module now has a real docs page (`docs/cache-stability.md`) with nav, index, and `_CAPABILITY_PAGE_META` parity rows; the top-level matrix row is a plain released `:white_check_mark:`. It stays **not** re-exported from the root `__init__.py`, which is true of every new capability per the authoring guide.

Gates: `make lint && make typecheck && make testcov` green, 100% branch coverage on the module, docs-parity + snippet suites pass. Re-requesting CodeRabbit so its verdict reflects the final state.

dsfaccini:
@coderabbitai review

pydantic_ai_harness/cache_stability/_capability.py
```diff
@@ -0,0 +1,213 @@
...
+            warnings.warn(
+                f'Cache hit collapsed at model request {state.step}: read {read} cached tokens but '
+                f'a prior request established ~{established} (~{wasted} tokens re-sent uncached). '
+                f"The cacheable prefix moved between requests, or the provider's cache expired{expiry}.\n\n"
+                f'To silence or escalate:\n\n{_SILENCE_HINT}\n',
+                CacheBustWarning,
+                stacklevel=2,
+            )
+            # Re-baseline to the collapsed prefix so a sustained collapse warns once, not per step.
+            prefix = new_prefix
```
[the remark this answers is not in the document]
dsfaccini: This reply was posted by Claude Code using claude-opus-4-8 on behalf of David.

_David steered this fix but has not code-reviewed the diff (he may be away) -- lightly steered, not reviewed._

Good catch, and confirmed -- the re-baseline broke the "warns once" guarantee for write-reporting providers. Fixed in 39cd255.

Worth noting the suggested `read` value only half-fixes it. Trace `read=0, write=2000` sustained (min_prefix 1024, ratio 0.5):

```
R1  write=8000            -> mark 8000
R2  read=0  write=2000    -> collapse, WARN, rebaseline mark=read=0
R3  read=0  write=2000    -> mark 0 < min_prefix, no collapse, but else-branch max(0, 2000) re-grows mark to 2000
R4  read=0  write=2000    -> mark 2000, read 0 < 1000 -> collapse, WARN again
```

Re-baselining to `read` drops it from every-step to every-*other*-step, because the intervening non-collapse step re-grows the mark via `max(established, read + write)`. Any pure re-baseline heuristic has this hole.

So I latched the state instead: a per-key `collapsed` flag, set on the transition into a collapse and used to suppress the warning while it persists; the prefix stays a plain high-water mark, and the latch clears on a healthy read-back so a genuine second collapse still warns. `read=0, write=2000` sustained now warns once (verified end-to-end). Regression: `test_sustained_collapse_with_cache_writes_warns_once` (fails on both the old `read+write` and the `read` variants), plus `test_recollapse_after_restabilize_warns_again` for the re-arm.

pydantic_ai_harness/cache_stability/_capability.py
dsfaccini: This reply was posted by Claude Code using claude-opus-4-8 on behalf of David.

Correction to the above: the re-baseline mechanism I described didn't fully hold. Macroscope caught that for a provider which keeps writing an unread cache (`read=0, write=2000` every step), re-baselining to `read + write` holds the mark high and re-warns every step -- and re-baselining to `read` alone still warns every *other* step because the intervening `max()` re-grows the mark. Replaced it with a per-key collapse latch (warn on the transition in, re-arm on a healthy read-back), which holds the "warns once" guarantee regardless of the provider's write reporting. Details in the Macroscope thread; fix in 39cd255.

PR #332, opened 2026-07-08
feat(overflow): grep/slice query surface + uniform explicit elision markers

DouweM:
## Summary

Completes the query surface for the `overflow` capability's spilled artifacts and makes every truncation/spill leave an explicit, uniform marker. Design principle (from the model's own testimony in the harness-comparison spec): "spilling without retrieval tools just hides information -- strictly worse than truncation-with-marker."

Extends the overflow/Spill machinery already on `main` (the `read_tool_result` tool, `LocalFileStore`, bands). No behavior change except where content was previously elided silently -- that is the fix.

## What / How

**(a) Query affordances on spilled artifacts.**
- New `grep_tool_result(handle, pattern, context_lines=2, max_matches=20, is_regex=False)`: searches a spilled payload and returns matched lines with 1-based line numbers plus surrounding context, grep-style (overlapping context merges; non-adjacent groups separated by `--`). Literal substring by default; opt-in Python regex applied per line (backtracking bounded to a line length). Matches, context, and joined output are each capped.
- `read_tool_result` gains a byte-range mode: `unit='lines'` (existing behavior, line slice + optional literal `pattern` filter) or `unit='bytes'` (byte slice for payloads without useful line breaks). `pattern` is rejected with `unit='bytes'`.
- Both tools share the missing-handle behavior from #293 (return guidance, never raise, so a bad/expired handle can't consume a retry). Both are exempt from re-reduction. Tool docstrings tell the model the intended loop: grep to locate, then read the range around a match.

**(b) Explicit, uniform elision markers everywhere.** New `_markers.py` factors one handle-aware marker helper used by every elision path:
- Spill preview banner + middle-elision marker now name **both** query tools and the handle.
- `Truncate` (head/tail/head_tail) markers are unified through the same helper; being lossy, they carry no handle and say to re-run the tool.
- The `read_tool_result` output cap previously dropped its tail with only a header note; it now leaves an inline marker at the cut.
- Markers are timestamp-free, so an identical return reduces to identical bytes (cache-prefix stability -- a moving marker would bust the prompt cache). Covered by determinism tests.

## Prior art reconciled

- **Relates to #82** (Tool Output Management). This lands the spill-to-file *retrieval* surface #82 asked for (search + range reads + agent-facing hints). Out of #82's scope here: model-aware limit scaling and the summarization tier (summarize already exists on `main` as a band).
- **Supersedes the truncation/spill portions of #185.** The overflow capability on `main` already generalized #185's `ToolOutputManagement` (Truncate + Summarize) and added the spill-to-file that #185 deliberately dropped; this PR completes it with the query tools. Adopted from #82/#185: the retrieval-hint marker framing ("what was removed and how to get it") and per-tool config (already on `main` via `per_tool`). Dropped: baking into the agent loop (it is a capability, per maintainer direction on #82).

## Deliberate scope

- Regex is opt-in, per-line, literal-by-default -- preserves the existing ReDoS-avoidance stance rather than exposing whole-blob regex.
- No new store/protocol changes; durable-backend seam is unchanged.
- Compaction receipts (spec N3) are a separate track, not touched here.

## Tests

Added: grep (matches + line numbers, context, `max_matches` cap + `N+` flag, no-match, separated groups, regex, literal-metachar, invalid-regex/empty/bad-knob `ModelRetry`s, missing handle, output cap, toolset registration, exemption); byte-range reads (basic, from-end, out-of-range, pattern rejection, cap); marker presence/format per elision path; marker determinism (two identical runs -> identical bytes). `make lint && make typecheck && make test` pass; 100% branch coverage retained on the overflow module.

Spec: https://github.com/pydantic/pydantic-ai-notes/blob/main/features/harness-comparison/2026-07-06%20what%20fable%20wants%20from%20a%20harness.md

### Provenance fixes (2026-07-10)

The `Summarize` band replaced a tool's real content with an LLM summary carrying no marker — the model (and every downstream model) saw the summary as the tool's genuine output, inconsistent with this PR's own uniform-explicit-elision-marker goal that every other path already meets (audit §2 H2).

- Added `summary_header` to `_markers.py` (same bracketed, timestamp-free convention as `spill_header`/`elision_marker`) and prefixed it onto the summary: it states the original size, that the harness summarized it, and how to recover the original.
- The spill preview path (H3) was already fully marked via `spill_header`; no change needed there.


DouweM:
Reminder to check out vstorm's `EvictionCapability`

DouweM:
@coderabbitai review

Issue #83, opened 2026-03-21
System Reminders / Mid-Session Behavioral Steering


adtyavrdhn:
## Summary

Event-driven injection of behavioral guidance into the agent's context at decision points during long-running sessions. System reminders counteract **instruction fade-out** — the phenomenon where agents progressively ignore system prompt guidelines after many turns of tool use.

This is **distinct from**:
- #64 (Repository Context Injection) — which loads static project conventions at session start
- #11 (Instructions capability) — which provides instructions at the start of each request
- #21 (Compaction) — which summarizes old context when the window fills up

System reminders are **reactive** — they fire when specific conditions are detected (repeated errors, approaching limits, mode switches) and inject targeted guidance at the moment it's needed.

## Motivation

Every harness dealing with long-running sessions has identified instruction fade-out as a critical problem:

- **OpenDev** (arxiv 2603.05344, Section 2.3.4) implements "Context-aware System Reminders" with event detectors, template resolution, and guardrail counters. They note that a "single system prompt is insufficient for multi-turn sessions" and that "instruction fade-out occurs (agent forgets guardrails after many turns)"
- **Claude Code** has `PreCompact`/`PostCompact` hooks specifically for re-injecting critical context after compaction. Community tools like [post_compact_reminder](https://github.com/Dicklesworthstone/post_compact_reminder) exist to prevent "post-compaction rule amnesia"
- **LangChain** describes `LoopDetectionMiddleware` that injects "reconsider approach" context after N edits to the same file, and `PreCompletionChecklistMiddleware` that intercepts the agent before exit to enforce verification
- The [coding-agent-loop-spec](https://github.com/strongdm/attractor/blob/main/coding-agent-loop-spec.md) implements `SteeringTurn` — injected messages between tool rounds for mid-task redirection

### Reminder Triggers (Event Detectors)

| Trigger | Condition | Injection |
|---------|-----------|-----------|
| `repeated_tool_failure` | Same tool failed N times in last M turns | "You've encountered this error N times. Consider a different approach." |
| `loop_detection` | Repeating 1/2/3-length action sequences detected | "You appear to be in a loop. Step back and reconsider." |
| `token_budget_warning` | Context usage > threshold % | "Context is getting large. Consider finishing current task or using a subagent." |
| `post_compaction` | After context was compacted | Re-inject: active plan, key constraints, current task state |
| `mode_switch` | Agent transitions between phases | Phase-specific guidance (e.g., entering verification phase) |
[...]

PR #333, opened 2026-07-08
feat(staleness): tell the model when files changed since it read them

DouweM:
## What

Adds `StalenessTracker`, a private/experimental capability that tells the model when files it read have changed on disk since it read them.

A model cannot perceive elapsed time or concurrent change: once it reads a file, it acts on that snapshot for the rest of the run, even if another agent, a build step, `git checkout`, or the user edits the file underneath it. This is roll-up item **N1** / §3 of [what Fable wants from a harness](https://github.com/pydantic/pydantic-ai-notes/blob/main/features/harness-comparison/2026-07-06%20what%20fable%20wants%20from%20a%20harness.md) ("freshness and staleness signals -- I cannot perceive time"), carved out as its own small capability, modeled on Hermes's edit-staleness ledger.

## How

- **Observe** (`after_tool_execute`): after a file read/write tool call lands, record `(path, mtime_ns, size)`. Because observation happens *after* the write, a self-write records the post-write stat and never flags itself -- self-writes are not staleness.
- **Notice** (`wrap_model_request`): before each model request, re-stat the tracked set; any file whose `(mtime, size)` changed, or that was deleted, contributes to an ephemeral notice:
  > `<system-reminder>Files changed on disk since you last read them: src/foo.py, tests/test_foo.py. Re-read before relying on their contents.</system-reminder>`
- The notice is appended to the per-request message tail behind a `CachePoint` (the `Planning` pattern): it reaches the model but is **never written to durable history**, so the cached prefix stays byte-stable and notices never accumulate. Verified by a test asserting `result.all_messages()` carries no reminder.
- **Cheap by construction**: stat-only (no hashing) over an LRU-capped tracked set (default 200).

## Design decisions (delegated by Douwe -- flagged as easily revisable)

- **Which tools/args count is data-driven**, not hardcoded. `track: dict[str, str | Callable]` maps `fnmatch` tool-name patterns to the arg holding the path (or a callable over the args); default covers `read`/`read_file`/`write`/`write_file`/`edit`/`apply_patch` reading `file_path`/`path`. A `path_extractor` escape hatch replaces `track` wholesale. **This default is a starting guess -- trivially reconfigured per host, no subclassing.**
- **Notice channel**: I aligned the *wording* with the `<system-reminder>` convention from #181 (`SystemReminders`) and issue #83, but deliberately **diverged on delivery**: #181 injects persisted `SystemPromptPart`s via `before_model_request`, whereas the spec (and the cache-stability contract) require the notice to be *ephemeral*. So delivery uses the newer non-persisting `wrap_model_request` + per-request-tail + `CachePoint` mechanism (`Planning`), which never mutates stored history. Both the tag and this channel are one-line changes if the team prefers otherwise.
- **Notice keeps firing until re-read** (mirrors Hermes): a changed file stays flagged every request until the model reads it again (which refreshes the record). The honest signal is "still stale", not "told you once"; free because the notice is ephemeral.
- **Cap** on listed names (default 10 per changed/deleted group, `(+N more)`), separate from the LRU cap on the tracked set.

- **No content hashing** -- stat (mtime + size) only; a same-size, same-mtime rewrite isn't detected. Hashing every tracked file per request is exactly the cost this avoids.
- **No directory watching** -- only files the model actually read/wrote are tracked.
- **No git-status integration** -- no branch/stash/index awareness.
- **No cross-run persistence** -- the ledger is per-run (`for_run` returns a fresh instance) and never shared between concurrent runs.
- **Private / experimental** -- lives in `experimental/staleness`, not re-exported from top-level `pydantic_ai_harness`; opts out of spec construction (`get_serialization_name` returns `None`) since its config holds callables.

## Tests

29 tests, 100% branch coverage on the module. End-to-end via `Agent(..., capabilities=[...])` with `FunctionModel`-scripted steps that mutate `tmp_path` files between turns: read -> external edit -> notice appears; read -> agent's own write -> no notice; deletion notice; notice keeps firing until re-read; **no history mutation** (stored messages assert-clean); per-run isolation. Focused unit tests cover the extractor (defaults, glob, string/callable values, `path_extractor`), path resolution, LRU eviction, the `(+N more)` cap, combined changed+deleted notices, and the `wrap_model_request` edges. Full suite green (1640 passed / 10 skipped), lint + pyright-strict clean.


DouweM:
@coderabbitai review

DouweM:
Naming note from the capability naming pass (#392, implemented in #480): single-behavior capabilities that act on the run take an imperative verb phrase, and agent-nouns like `*Tracker` are avoided (compare `CacheStabilityMonitor` -> `WarnOnCacheBusts` in #480). The contract here is telling the model which files changed since it read them, so `ReportStaleFiles` (or `TrackFileStaleness` if the tracking mechanism is the emphasis) would fit the convention. Posted by Claude Code, acting on @DouweM's behalf.

PR #334, opened 2026-07-08
feat(budget_disclosure): disclose remaining usage budget to the model

DouweM:
## What

Adds `BudgetDisclosure`, a private/experimental capability that contributes one short remaining-budget line to each model request, so the model can pace itself as the budget runs down instead of being stopped mid-step by a `UsageLimitExceeded`.

```
Budget remaining: ~38k tokens, 7 requests. Pace your work; if nearly exhausted, prioritize delivering current results over starting new work.
```

This is the budget-transparency want from the model's seat (roll-up item **N4** / §6 of the "what Fable wants from a harness" note). No surveyed harness discloses budget to the model except Hermes: `UsageLimits` enforces but never lets the model *see* the ceiling coming.

Spec: <https://github.com/pydantic/pydantic-ai-notes/blob/main/features/harness-comparison/2026-07-06%20what%20fable%20wants%20from%20a%20harness.md>

## How

- Contributes the line through the capability `get_instructions` channel. Instructions are rebuilt every request, so this is the cache-safe channel: a remaining-budget number changes on every request, and persisting it as a message part would move the cacheable prefix each turn and re-charge the whole conversation. On each request the callable reads the run's accumulated `ctx.usage`, computes remaining per configured limit, and returns the line (or `None` to contribute nothing).
- Reads the run's budget from a `limits: UsageLimits` on the capability. The run's `UsageLimits` (passed to `Agent.run(..., usage_limits=...)`) is not reachable from a capability -- it lives on the agent graph's run deps, not the `RunContext` a capability sees -- so it is configured on the capability, mirroring `compaction.LimitWarner`, which takes its own `max_*` ceilings for the same reason. Pass the same `UsageLimits` you pass to the run.
- Stateless: reads `ctx.usage` and the static `limits` each request, so no `for_run` and one instance is reusable across runs.
- Options: `disclose` (which limited dimensions to show; default is every dimension whose limit is set), `start_at` (fraction of a limit that must be consumed before disclosure begins; `0.0` always discloses), `round_tokens_to` (token-count display granularity, default nearest 1k), `format` (override the line entirely).

## Relationship to #168

Complements enforcement, does not add it. `ToolBudget` (#168) and the underlying `UsageLimits` *stop* the run when a ceiling is hit; this capability *discloses* the ceiling so the model sees it coming and paces toward it. Scope is deliberately disclosure-only.

- **Instructions channel, not a message part.** Unlike `LimitWarner`, which appends a `UserPromptPart` warning into the request messages, this contributes only ephemeral per-request instructions, so the changing budget number never enters the cacheable message prefix. A test asserts no persisted message part carries the line.
- **`limits` on the capability, not auto-read from the run.** The run's `UsageLimits` isn't exposed to capabilities today; configuring it here matches the existing `LimitWarner` precedent rather than reaching into core.
- **Disclosure only, no grace call.** The one-toolless-wrap-up "grace call" (Hermes) is a separate `UsageLimits`/core concern (task-budgets note), out of scope here.
- **Private / experimental.** Lives in `experimental/budget_disclosure`, not re-exported from top-level `pydantic_ai_harness`.

## Tests

15 tests driven through `Agent(..., capabilities=[...])` with a `FunctionModel` emitting preset `RequestUsage` per step, recording each request's `AgentInfo.instructions`:

- the line appears with correct remaining across multiple steps as usage accrues (tokens + requests);
- no limits -> nothing disclosed;
- the line is never a stored message part while it does reach the model via instructions (cache safety);
- `start_at` gating stays silent until a disclosed limit is that fraction consumed;
- token rounding (default 1k, configurable, non-thousand granularity);
- `disclose` dimension selection; `format` override (including returning `None`);
- a reused instance judges sequential runs independently;
- constructor validation (`start_at`, `round_tokens_to`, empty/unknown/limit-less `disclose`).

100% branch coverage on the module; full suite green locally (1404 passed / 12 skipped, `acp` suite excluded -- its optional dep isn't installed in this environment), lint + format clean, pyright clean on the new files (the repo-wide pyright errors are pre-existing, confined to `experimental/acp` from that missing dependency).

Co-Authored-By: Claude Opus <noreply@anthropic.com>


DouweM:
@coderabbitai review

DouweM:
Naming note from the capability naming pass (#392, implemented in #480): the convention avoids coined nominalizations -- the noun form adds formality without information. `BudgetDisclosure` states its whole contract in one verb phrase, so `ReportRemainingBudget` (or `DiscloseUsageBudget`) would fit; it also reads as a natural sibling of `WarnNearLimits` from #480. Posted by Claude Code, acting on @DouweM's behalf.

PR #337, opened 2026-07-08
feat(cross_model_label): tell the model when earlier turns came from a different model

DouweM:
## What

Adds `CrossModelHistoryLabel`, a private/experimental capability that tells the serving model when earlier assistant turns in the history were produced by a *different* model.

When a run continues a history written by another model -- a `FallbackModel` failover, a model swap between runs, an A/B handoff, a takeover -- the serving model otherwise reads those turns as its own. It defends claims it never made and keeps commitments it cannot verify. The capability detects the mismatch and contributes one short line:

```
Note: assistant responses before this point were produced by a different model (gpt-5.2).
Treat their claims and commitments as inherited context, not your own output.
```

This is the label-cross-model-history want from the model's seat (roll-up item **N5** / §10 of the "what Fable wants from a harness" note). Nobody ships this; history already carries `model_name` per response, so it is derivable for free. Prior art for the framing is Codex's compaction bridge prefix ("another language model produced this").

Spec:
- <https://github.com/pydantic/pydantic-ai-notes/blob/main/features/harness-comparison/2026-07-06%20what%20fable%20wants%20from%20a%20harness.md> (§10, roll-up N5)
- <https://github.com/pydantic/pydantic-ai-notes/blob/main/features/model-gated-prompts/2026-07-06%20design%20-%20per-model%20prompts%20x%20cross-model%20histories.md> (§4, detection design)

## How

- **`get_instructions` channel, never persisted.** On each request the callable resolves the serving model's identity, scans prior `ModelResponse.model_name` values, and returns the line (or `None`). Instructions are rebuilt per request, so this is the cache-safe channel -- and the correct provenance channel: a note *about* the history must not itself become history a later model reads back as fact.
- **Family-level comparison by default.** `gpt-5.2-mini` and `gpt-5.2` are the same family and stay silent; `gpt-5.2` vs `claude-sonnet-4-5` fires. Pydantic AI profiles expose no first-class family key (I checked -- family lives implicitly in each provider's prefix-matching profile logic), so the default resolver derives one heuristically from the model name (strips a provider segment, dated snapshots, `-latest`/`-preview`, and size-tier suffixes). It is best-effort and fully overridable.
- **Stateless.** Reads only `ctx.model` and `ctx.messages` each request; one instance is reusable across runs. It never mutates the message history.
- **Options:** `granularity` (`'family'` | `'exact'` | `(model_name, provider_name) -> key` callable), `threshold` (`'recent'` default, or a float fraction), `format` override. The default resolver is exported as `model_family` so a callable can wrap it.

## FallbackModel

Under a `FallbackModel` the serving member is unknown until it answers, so the current identity is taken from the most recent response's `model_name`, which records who actually served (core [#6338](https://github.com/pydantic/pydantic-ai/pull/6338)), falling back to the wrapper's first candidate before any response. A mid-history failover (early turns from A, later from B) is therefore detected against B, the model now serving. The label composes with #6338 rather than duplicating it.

- **`threshold='recent'` is a one-shot handoff nudge.** It fires while the immediately-preceding response is foreign, then goes quiet once the serving model has added its own turn. For a persistent provenance banner across a long continuation, pass a float fraction (fires while that fraction of history stays foreign).
- **Family heuristic, not a taxonomy.** The default resolver only collapses unambiguous size-tier suffixes (`mini`, `nano`, `small`, `lite`, `tiny`); distinctly-named tiers (`sonnet` vs `opus`, `flash` vs `pro`) are treated as different families on purpose -- missing a warning is worse than a spurious one. Override `granularity` with a callable for other policies.
- **Missing `model_name` is skipped, never guessed.** Old turns without a recorded model contribute nothing.
- **Detect and disclose only.** No history normalization / rewriting (that is a cache-cold concern, deferred in the design §5). No `ModelProfile` behavioral-flag work (design §7 work item 1) -- this PR is only the label.
- **Private / experimental.** Lives in `experimental/cross_model_label`, not re-exported from top-level `pydantic_ai_harness`.

## Tests

23 tests driven through `Agent(..., capabilities=[...])` with a `FunctionModel` (identity set via `FunctionModel(model_name=...)`, foreign history supplied as scripted `ModelResponse`s), recording each request's `AgentInfo.instructions`:

- fires when the history is a different family; silent on fresh history, same-family smaller sibling, dated snapshot, provider-only difference, and missing `model_name`;
- `'recent'` vs float-fraction threshold semantics (immediately-preceding only; majority-foreign; below-fraction silence; predominant-family naming with most-recent tie-break);
- `'exact'` granularity and a custom callable resolver;
- real `FallbackModel` runs: current identity resolved from the most recent response, and the fresh-history first-candidate fallback;
- `format` override (including returning `None`);
- the line is never a persisted message part (cache/provenance safety); a reused instance judges sequential runs independently; constructor validation and the `model_family` / `normalize_model_name` helpers.

100% branch coverage on the module. Full suite green locally (1414 passed / 12 skipped); lint + format clean; pyright clean on the new files (the repo-wide pyright errors are pre-existing, confined to `experimental/acp` and the `dbos`/`temporal` tests whose optional deps aren't installed here, so this commit used `--no-verify` for that pre-existing hook failure only).


DouweM:
Naming note from the capability naming pass (#392, implemented in #480): single-behavior run capabilities take an imperative verb phrase rather than a noun. This capability adds a one-line notice when earlier turns came from a different model, so `LabelCrossModelHistory` states the contract directly. Posted by Claude Code, acting on @DouweM's behalf.

PR #344, opened 2026-07-10
feat(code_mode): migrate to pydantic-monty 0.0.19 beta

dsfaccini:
## What

Migrate `code_mode` and `dynamic_workflow` from `pydantic-monty` 0.0.18 to the 0.0.19 beta (0.0.19b2), so we can dogfood and test the new Monty runtime.

Draft because it pins a beta and (see below) drops CodeMode on Python 3.14 until Monty ships cp314 wheels.

## API deltas handled (0.0.17/0.0.18 -> 0.0.19b2)

- `MontyRepl` removed. Execution moved to a subprocess pool: `Monty` pool -> `pool.checkout()` -> `MontySession`. A `MontyReplSession` shim in `_monty_exec.py` restores the persistent-REPL shape the toolset drives.
- `resume()` no longer takes `mount=` (mounts are fixed at `feed_start`); `os=` still passes through.
- `Monty(code, type_check=...)` replaced by checkout-time `type_check` / `type_check_stubs` plus a first-feed check.
- New `MontyCrashedError` (worker crash/timeout, carries `.timed_out`) replaces the host `PanicException` for execution crashes; the panic guard is kept only for in-process conversion panics.
- Kept the manual `feed_start`/`resume` snapshot loop; did not adopt the new `external_lookup` / `resume_auto`. The manual loop retains ToolManager routing, argument validation, nested-call recording, exec-mode plus deferral handling, and stays Temporal-safe. `external_lookup` is a convenience alternative, not required.

Behavior changes accepted (Monty improvements), with matching test/prose updates:

- A duplicate future inside one `gather` now completes instead of panicking the VM.
- A host-raised exception is now catchable inside the sandbox at its `await`. The `run_workflow` tool description previously said sub-agent errors "cannot be caught inside the script"; it is corrected to reflect the new behavior. The `gather(return_exceptions=True)` limitation still holds and is kept.
- First-feed syntax errors surface through the type checker as `invalid-syntax` and are routed to a "Syntax error" message.

## Python 3.14

The 0.0.19 beta ships wheels only up to cp313 (0.0.17 was the last with cp314), so it is uninstallable on 3.14. Every `pydantic-monty` spec is gated `; python_version < '3.14'`, so on 3.14 Monty is simply absent. The import guards are version-aware (a clear "no 3.14 wheels, use Python 3.13 or earlier" error) and the package import stays lazy, so `import pydantic_ai_harness` still works without Monty; only using CodeMode / DynamicWorkflow on 3.14 raises. The lock carries the same marker plus the new `pydantic-monty-runtime` split package.

## Verification

On a py3.13 / monty-0.0.19b2 env: full suite (1427 tests) green, ruff and pyright strict clean, 100% branch coverage on the touched modules, and 8/8 scenarios in the new benchmark harness.

New reusable benchmark harness at `benchmarks/codemode/` (`bench.py` + `scenarios.py` + `setup_env.sh` + README): drives a real `Agent(capabilities=[CodeMode])` through scripted `run_code` calls via a `FunctionModel`, verifies tool-bridging, and times each scenario. `--json` supports A/B across Monty / harness builds. No network or API keys. It is meant to test and benchmark future Monty and code_mode changes too, not just this bump.


dsfaccini:

Closing as superseded by #345, which carries the GA Monty migration and the relevant lifecycle, error-handling, and Temporal coverage. The beta-specific compatibility and CI work here is no longer applicable; the benchmark idea can return separately if useful.

PR #345, opened 2026-07-10
Upgrade Monty to 0.0.19

adtyavrdhn:
## Summary

- Upgrade CodeMode and DynamicWorkflow to the released Monty 0.0.19 pool/session API.
- Require Monty 0.0.19 or newer without an artificial upper bound.
- Keep one lazily-created worker pool and live REPL session per CodeMode run, with explicit reset behavior for failures and crashes.
- Preserve multi-feed REPL state, Temporal replay, overlay-mount semantics, and terminal DynamicWorkflow budget exhaustion.
- Remove prerelease-only runtime pins because Monty 0.0.19 pins its runtime transitively.
- Follow the GA resource API by removing the unsupported allocation limit and using Monty's bounded print collector.
- Keep removed DynamicWorkflow resource keys on the ordinary unknown-key path, without migration-specific compatibility code.
- Remove generic API-instability disclaimers and audit the Code Mode and Dynamic Workflow docs against the public APIs, implementation, and Monty 0.0.19 behavior.
- Tighten model-facing Monty instructions so async calls, concurrency limits, and catchable workflow failures are explicit and configuration-accurate.
- Preserve terminal budget semantics even when workflow code catches exhaustion and later crashes the sandbox worker.
- Simplify Code Mode lifecycle ownership and replace private-state test instrumentation with observable REPL, cancellation, and reset behavior.

## Validation

- `make lint`
- `make typecheck`
- `make test`
- `make testcov` (3,083 passed, 35 skipped, 100% statement and branch coverage)
- Real Monty 0.0.19 throwaway Agent runs for Code Mode parallel tool dispatch and REPL reuse
- Real Monty 0.0.19 throwaway Agent runs for Dynamic Workflow parallel sub-agents, caught failures, and crash-after-budget behavior
- Real Monty 0.0.19 print-limit probe confirming the same REPL remains usable and preserves assignments after `feed_start` raises
- Monty 0.0.19 cleanup implementation review confirming pool and session exits are non-raising, best-effort operations
- Real Temporal probes for concurrent `asyncio.gather`, explicit sequential mode, worker replacement, activity failure handling, and replay nondeterminism from changing mounted input
- Document the Temporal workflow/activity boundary, process-local REPL reconstruction, host-access determinism requirements, and activity timeout scope
- Multiple final adversarial implementation, test, API, documentation, and simplicity reviews with zero findings after accepting the intentional `max_allocations` break


adtyavrdhn:
I confirmed this can be made to run while keeping Temporal's workflow sandbox enabled:

```python
workflow_runner = SandboxedWorkflowRunner(
    restrictions=SandboxRestrictions.default.with_passthrough_modules('pydantic_monty')
)
```

This passes Monty's native module through from outside the workflow sandbox, so its subprocess pool is not intercepted. Execution and history replay both worked locally.

I would not recommend this as the long-term/default contract: passthrough is intended for deterministic, side-effect-free modules, while Monty owns external subprocess state that Temporal does not record. Replay re-executes that work, and `mount`/`os_access` would make the determinism risk larger. An in-process Monty path or moving this execution behind a Temporal activity would preserve the durability boundary.

adtyavrdhn:
After reproducing both the live failure and replay behavior, I suggest rolling back the Temporal-specific `TemporalCodeMode` activity/ping-pong implementation on the current head. We should keep this PR focused on the Monty beta/API migration, then add `pydantic_monty` to Pydantic AI's default Temporal passthrough list.

What I verified with Monty `0.0.19b2` and Temporal `1.27.1`:

- Monty-enforced recursion, memory, allocation, and duration limits raise `MontyRuntimeError`. Code Mode converts those to `ModelRetry`, so the model can revise its code.
- A real child-process death raises `MontyCrashedError`. I reproduced this by sending `SIGKILL` to the checked-out worker.
- In live Temporal execution, the parent worker survives, the execution remains `RUNNING`, history records `WorkflowTaskFailed` with cause `WORKFLOW_WORKER_UNHANDLED_FAILURE`, and another Workflow Task is scheduled. A transient crash can recover; a repeatable crash remains `RUNNING` and retrying.
- This is Temporal's normal handling for an unclassified workflow exception. Pydantic AI currently classifies only `UserError` and `PydanticUserError` as workflow-failing exceptions.

Replay has an important additional constraint:

- A completed history containing an earlier `WorkflowTaskFailed(MontyCrashedError)` replays successfully if Monty succeeds during replay. The historical crash does not need to be reproduced.
- If Monty crashes during offline `Replayer` execution, that replay invocation fails immediately. Unlike the Temporal server, `Replayer` does not retry the failed Workflow Task.
- Monty code is re-executed during replay. A changed value can go undetected when it emits the same Temporal command sequence, even when an activity argument changes. If the value changes the activity/tool type or command order, replay fails with `NondeterminismError`.

Therefore the passthrough approach preserves the previous in-process replay contract, but we must explicitly accept that Monty has to execute successfully and deterministically during replay. The activity-based implementation provides stronger durability by recording Monty segments, but at substantially greater complexity.

Suggested scope:

1. Remove `TemporalCodeMode` and the activity-side dump/resume protocol.
2. Keep the Monty API/version migration and normal Code Mode state handling required by the new API.
3. Add `pydantic_monty` to Pydantic AI core's Temporal passthrough modules.
4. Retain the happy-path Temporal and `Replayer` coverage.
5. Add crash coverage for both live Workflow Task retry and offline replay failure.
6. Keep mutable filesystem, network, clock, environment, `mount`, and `os_access` behavior behind recorded tools/activities.

I would not convert `MontyCrashedError` to `ModelRetry`. A system-dependent crash should fail the Workflow Task so live Temporal retries from committed history rather than committing a new model path based on a nondeterministic failure.

Sources: [Temporal exception behavior](https://github.com/temporalio/sdk-python#exceptions), [sandbox and passthrough guidance](https://github.com/temporalio/sdk-python#workflow-sandbox), [workflow replay](https://github.com/temporalio/sdk-python#workflow-replay), and [Monty crash semantics](https://github.com/pydantic/monty/blob/0.0.19-beta.2/crates/monty-python/python/pydantic_monty/_monty.pyi#L252-L279).

pyproject.toml
```diff
@@ -34,11 +34,13 @@ dependencies = [
 
 [project.optional-dependencies]
 code-mode = [
-    'pydantic-monty>=0.0.16',
+    "pydantic-monty>=0.0.19b6,<0.0.20",
```
adtyavrdhn: Maybe should pin to < `19`?

Issue #349, opened 2026-07-11
Subprocess process-group termination uses POSIX-only APIs (breaks on Windows)


strawgate:
Claude here: filing a validated portability bug found while dogfooding a new capability.

## Summary

The subprocess timeout/kill paths call POSIX-only APIs unconditionally, so they raise `AttributeError` on Windows. This affects the shipped `Shell` capability and the experimental `macroscope` capability.

## Affected code

- `pydantic_ai_harness/shell/_toolset.py` `_kill_process_group()` -- `os.killpg(os.getpgid(pid), signal.SIGTERM)` and `... signal.SIGKILL` (lines ~263, ~273).
- `pydantic_ai_harness/experimental/macroscope/_toolset.py` `_terminate()` -- `os.killpg(os.getpgid(proc.pid), signal.SIGKILL)` (line ~230).

Both are reached on the command/review timeout path.

## Why it breaks on Windows

Per typeshed, these names do not exist on `win32`:

- `os.killpg` -- guarded by `if sys.platform != "win32":` (`stdlib/os/__init__.pyi`)
- `os.getpgid` -- same guard
- `signal.SIGKILL` -- only defined in the non-`win32` `else` branch of the `Signals` enum (`stdlib/signal.pyi`)

At runtime on Windows, `os.killpg`/`os.getpgid`/`signal.SIGKILL` raise `AttributeError`. The surrounding handlers catch `OSError` (and `ProcessLookupError`/`PermissionError`), none of which is `AttributeError`, so the exception propagates and aborts the run instead of terminating the process.

`start_new_session=True` on `open_process` is a POSIX-only no-op on Windows, so the process-group model it sets up does not exist there either.

## Why CI does not catch it

`[tool.pyright]` sets no `pythonPlatform`, so pyright type-checks against the platform it runs on. CI is `ubuntu-latest` across the board, so the Unix-only symbols are always in scope and strict mode passes. There is no Windows job in `.github/workflows/`.

## Suggested direction

Branch the kill path on `sys.platform`:

- POSIX: keep the current `os.killpg(os.getpgid(...), SIG...)` group-signal behavior.
- Windows: fall back to `anyio`'s `Process.terminate()` / `Process.kill()` (which map to `TerminateProcess`), optionally with `CREATE_NEW_PROCESS_GROUP` when a group is needed.

[...]

PR #350, opened 2026-07-11
feat: Macroscope CLI code-review capability

strawgate:
Adds a `Macroscope` capability that gives a Pydantic AI agent the same local code review that [Macroscope](https://docs.macroscope.com/cli)'s editor plugins run.

## What it does

`run_macroscope_review(base?)` shells out to the installed `macroscope codereview` binary, parses its streamed findings, and returns them as a structured `MacroscopeReview` (review id, terminal status, list of `MacroscopeIssue`). The agent validates each finding and fixes the real ones with its own tools (`FileSystem`, `Shell`, etc.) -- this capability surfaces findings only; it does not edit files or manage worktrees.

```python
from pydantic_ai import Agent
from pydantic_ai_harness.macroscope import Macroscope

agent = Agent('anthropic:claude-sonnet-5', capabilities=[Macroscope()])
agent.run_sync('Run a Macroscope review and fix any real findings.')
```

## Requires a one-time install + login on the host

The capability drives the user-installed `macroscope` binary and cannot authenticate on the user's behalf:

1. Install: `curl -sSL https://raw.githubusercontent.com/prassoai/macroscope-local/main/install.sh | bash`
2. Sign in + pick a workspace: run `macroscope` once.

If the binary is missing or a review never starts, the tool returns a clear `ModelRetry` explaining the fix. The README documents this up front.

## Notes

- No new package dependencies -- it shells out to the CLI.
- Top-level capability (`pydantic_ai_harness.macroscope`), not re-exported from the root package.
- Ships with a unified-docs page (`docs/macroscope.md` + nav entry) registered in the docs-parity checks, and uses the same tri-state `guidance` field as `Planning`/`ExaSearch` (`None` = default instructions, `''` = none, string = custom).
- Verified end-to-end against the real CLI (returns a review id + parsed findings). Macroscope reviewed this capability's own code during development; findings were validated and addressed.
- Follow-up #349 tracks a codebase-wide Windows portability gap in subprocess process-group termination (shared with `shell`); deliberately out of scope here.


strawgate:
Claude here: investigated the new Macroscope CLI release and updated the capability.

**What changed in the CLI** (build `macroscope-local-20260719`): `codereview` now launches an interactive full-screen TUI when attached to a terminal. Machine-readable output is gated behind a new `--raw` flag (or non-terminal stdout). New flags: `--raw`, `--interactive`, `--tui`. The streamed protocol (`review_id=`, `issue_event=<json>` with the same fields, `issue_status=`) and base auto-detection are unchanged.

**What I changed here:** the tool now invokes `macroscope codereview --raw`, so parsing works regardless of whether the host attaches a pty to the subprocess (relying on the CLI's non-terminal auto-detection was the latent risk). Updated the three CLI-arg test assertions and added a synced note to the README and `docs/macroscope.md`.

**Verification:** lint / pyright / 16 tests / 100% branch coverage green; docs-parity + doc-snippet suites pass; docs-parity-reviewer pass. Ran a live review against the new binary with `--raw` end-to-end -- returned a `review_id`, `status: completed`, and a parsed finding.

Optional follow-up (pre-existing, unrelated to this change): `shutil.which(self._command)` resolves a relative `command` path against the process cwd rather than the configured `cwd`; only matters if someone passes a relative binary path with a different `cwd`. Left as-is.

PR #359, opened 2026-07-13
ci: drop the pydantic-ai v2 beta early-warning job

adtyavrdhn:
## Summary

Removes the `test-v2-beta` job from `.github/workflows/main.yml`.

The job was a non-blocking early-warning signal for pydantic-ai v2 prereleases -- deliberately excluded from the `check` gate (its `needs` list never referenced it) and explicitly documented as "a visible signal, not a required check." Nothing depends on it, so removing it changes no required status.

## Verification

- `grep` confirms no other references to the job, `UV_PRERELEASE`, or the `v2-beta` cache suffix remain in `.github/`.
- The `check` gate `needs` list (`lint, test, test-floor, test-latest, coverage`) is unaffected.


adtyavrdhn:
@coderabbitai review

PR #390, opened 2026-07-20
Fix ShellToolset process-kill crash on Windows

PremAmbre0:
## Summary
- `_kill_process_group` used POSIX-only `os.killpg`/`os.getpgid`/`signal.SIGKILL` to stop background and timed-out shell commands. These don't exist on Windows, so the call raised `AttributeError` instead of stopping the process -- crashing the caller (`stop_command`, `run_command`'s timeout path, and `__aexit__` cleanup) and leaking the real subprocess, since the crash happens before any signal is ever sent.
- On Windows, fall back to `proc.kill()` (anyio maps this to `TerminateProcess()`), guarded against an already-exited process the same way the POSIX path already was -- testing surfaced an uncaught `ProcessLookupError` here too, when stopping a process that had already finished.

## Known limitation
Unlike POSIX process-group kill, `proc.kill()` on Windows only stops the spawned process itself, not any children it started -- there's no cheap Windows equivalent of a process-group kill. A full fix would need Windows Job Objects, which is a materially bigger change than this crash fix. Documented in the updated docstring.

## Testing
Added two tests covering the Windows path: `proc.kill()` is called instead of `os.killpg`, and the already-exited-process race is handled without raising. Full `tests/shell/test_shell.py` suite + `ruff` + `pyright` pass on the changed files.

### DIFF
--- a/pydantic_ai_harness/shell/_toolset.py
+++ b/pydantic_ai_harness/shell/_toolset.py
@@ -9,6 +9,7 @@ import re
 import shlex
 import signal
 import subprocess
+import sys
 import tempfile
 import uuid
 from collections.abc import Awaitable, Callable, Mapping, Sequence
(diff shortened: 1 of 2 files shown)


adtyavrdhn:
Are you facing this issue or is this based on a code read with an agent? I'd defer these changes until someone actually faces this.

PremAmbre0:
This is already-filed issue #349, unassigned, not something I hit independently, so I verified it was real before sending a fix rather than trusting the report:
    await ts.start_command("sleep 100")
    await ts.stop_command(id)   # AttributeError: module 'os' has no attribute 'killpg'

#351 attempted the same fix earlier but wasn't verified on a real Windows machine (its own notes say the Windows branch was only exercised via mocks). I tested this one on my actual Windows machine and confirmed the crash and the fix.

No worries if you'd rather defer it, feel free to close if this isn't something to touch right now.

mpfaffenberger (approved):
The Windows fallback is correct: `os.killpg`/`os.getpgid`/`signal.SIGKILL` genuinely do not exist there, all references stay inside the function body so imports are safe, and the docstring is honest that `TerminateProcess` cannot reach grandchildren, which is the right call to document rather than fake. Verified locally: the shell suite passes on the branch (216 passed) and, because this branch is ~110 commits behind, I also merged current main into it and re-ran, 260 passed. Main's shell changes touched zero kill-path lines. Non-blocking: `PermissionError` is a subclass of `OSError` so the except tuple is redundant, harmless to leave.

PR #400, opened 2026-07-21
fix(code-mode): surface unsupported-syntax parser errors as ModelRetry

strawgate:
## Summary

`_type_check` catches `MontyTypingError` and `MontySyntaxError` and converts them to `ModelRetry`, per its documented contract. But Monty's parser raises `MontyRuntimeError` (`NotImplementedError`) at construction for syntax it does not implement -- e.g. `with` statements -- and that escaped uncaught, crashing the entire agent run.

The gap is reachable in practice: with a `mount` configured, `Path.open()` type-checks against the stubs, so a model writing the idiomatic

```python
with Path('/data/orders.csv').open() as f:
    ...
```

passes the type check, reaches the parser, and kills the run with an unhandled exception instead of getting a retry it would trivially recover from. Hit while exercising a dataset-analysis agent against a live model.

## Change

- `_type_check` now catches `MontyRuntimeError` and raises `ModelRetry('Unsupported syntax in code: ...')`, matching the treatment of type and syntax errors.
- Regression test reproduces the exact scenario (mounted filesystem + `Path.open` in a `with` block); verified red without the guard, green with it.

### DIFF
--- a/pydantic_ai_harness/code_mode/_toolset.py
+++ b/pydantic_ai_harness/code_mode/_toolset.py
@@ -739,6 +739,12 @@ class CodeModeToolset(WrapperToolset[AgentDepsT]):
             raise ModelRetry(f'Type error in code:\n{e.display()}') from e
         except MontySyntaxError as e:
             raise ModelRetry(f'Syntax error in code:\n{e.display()}') from e
+        except MontyRuntimeError as e:
+            # The parser rejects syntax Monty does not implement (e.g. `with`
+            # statements) with a runtime error at construction time. That is a
+            # property of the code, not of the run -- surface it as a retry so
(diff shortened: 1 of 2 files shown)


adtyavrdhn:
This should not be needed anymore with the bump, can your Claude verify?

strawgate:
Superseded by the Monty 0.0.19 bump (#345) and the code-mode executor rework: `with` statements are now implemented by the sandbox, so the crash this PR guarded against no longer exists -- verified by running the exact regression scenario (mounted `Path.open` in a `with` block) against current main, which executes cleanly. The `_type_check` seam this patched is also gone. Closing rather than rebasing a fix for a bug that can no longer occur.

PR #404, opened 2026-07-21
feat(planning): fold pydantic-ai-todo into the Planning capability

### LINKED ISSUE #403: Planning (enhancement to the existing `pydantic_ai_harness.planning` capability)
### Capability Name

planning

### Description

Today `Planning` is a deliberately minimal, cache-safe plan: a single `write_plan` tool over a fresh in-memory plan per run. That covers "keep a running plan in the model's head" but not the richer task-tracking that agents in production need.

This proposes growing `Planning` into a full task manager while keeping its defining trait -- the plan is surfaced as an ephemeral tail reminder behind a `CachePoint`, never in the system prompt, so the prompt cache is never invalidated. Concretely:

- **Granular tools** alongside `write_plan`: `read_plan`, `add_task`, `update_task_status`, `update_task_statuses`, `remove_task`.
- **Subtasks and dependencies** (opt-in via `enable_subtasks`): `add_subtask`, `set_dependency`, `get_available_tasks`, a `blocked` status, cycle detection, and auto-blocking.
- **Persistence** via a `PlanStore` protocol with `InMemoryPlanStore` (default), `SqlitePlanStore`, `PostgresPlanStore`, and `RedisPlanStore`. The Postgres/Redis stores take a caller-owned client through a small protocol, so the harness adds **no database-driver dependency**.
- **Events** (`PlanEventEmitter`: `created`/`updated`/`status_changed`/`completed`/`deleted`) so apps can react to plan changes.

This supersedes the standalone [`pydantic-ai-todo`](https://github.com/vstorm-co/pydantic-ai-todo) library (Vstorm), which can then be archived; its tools map onto the planning tools with renamed names.

### Use Case

A long-running coding or research agent needs a persistent, inspectable task list:

- A terminal/web UI renders the live task list and updates it as the agent works (needs granular tools + events to reflect each change, not just a whole-plan rewrite).
- A multi-session agent resumes an interrupted job and continues the same task list (needs persistence: SQLite locally, Postgres/Redis for a served, multi-tenant deployment keyed by `session`).
- A complex task is broken into subtasks with ordering constraints, and the agent picks the next unblocked step (needs subtasks + dependencies + `get_available_tasks`).

Throughout, the plan must stay visible to the model every turn **without** re-processing the whole conversation at full price -- which the existing cache-safe reminder already guarantees.

- `get_toolset()` -- registers the plan toolset (base tools always; subtask tools when `enable_subtasks=True`).
- `get_instructions()` -- static, cache-stable usage guidance in the system prompt.
- `wrap_model_request` -- appends the current plan as an ephemeral reminder behind a `CachePoint` (the cache-safe injection point), never persisted to history.
- `for_run` -- per-run isolation: resolves and caches the run's `PlanStore` (default backend is a fresh in-memory plan per run).

### Prior Art / References

https://github.com/vstorm-co/pydantic-ai-todo

DEENUU1:
## Summary

Expand the `Planning` capability from an in-memory, single-tool plan into a full task manager, folding in the task-list features of the standalone [`pydantic-ai-todo`](https://github.com/vstorm-co/pydantic-ai-todo) library (which this supersedes) -- while keeping planning's defining trait: the plan is surfaced to the model as an ephemeral tail reminder behind a `CachePoint`, so the cached prefix stays byte-identical across turns and the plan never lands in the system prompt.

**Tools.** `write_plan` (whole-plan replacement) stays, plus granular `read_plan`, `add_task`, `update_task_status`, `update_task_statuses`, and `remove_task`. With `enable_subtasks=True` you also get `add_subtask`, `set_dependency`, and `get_available_tasks`, the `blocked` status, a hierarchical `read_plan` view, cycle detection, and auto-blocking.

**Persistence.** New `PlanStore` protocol with four backends:
- `InMemoryPlanStore` (default -- a fresh, isolated plan per run, matching the previous ephemeral behaviour)
- `SqlitePlanStore` (local file, session-scoped)
- `PostgresPlanStore` (server DB over a **caller-owned** asyncpg-compatible pool)
- `RedisPlanStore` (over a **caller-owned** `redis.asyncio`-compatible client)

The Postgres and Redis stores depend only on small protocols and take a client you already own, so the harness pulls in **no database-driver dependency**. `store` / `store_resolver` select the backend per run.

**Events.** New `PlanEventEmitter` (`created`, `updated`, `status_changed`, `completed`, `deleted`) so applications can react to plan changes (UI progress, external trackers, notifications).

**Model changes.** `PlanItem` gains `id`, `active_form`, `parent_id`, and `depends_on` (all defaulted); `TaskStatus` gains `blocked`.

**Backward compatibility.** `PlanItem`, `Planning`, `PlanningToolset`, `TaskStatus`, and `render_plan` are unchanged for existing callers, and the `experimental.planning` shim keeps working. `Planning()` with no arguments behaves as before (a fresh in-memory plan per run) with `write_plan` still present.

> **Breaking change:** `PlanningToolset(state)` becomes `PlanningToolset(capability)` -- the toolset now resolves its store from the owning `Planning` capability.

Migration note for `pydantic-ai-todo` users (tools renamed): `write_todos` -> `write_plan`, `read_todos` -> `read_plan`, `add_todo` -> `add_task`, `update_todo_status(es)` -> `update_task_status(es)`, `remove_todo` -> `remove_task`; subtask tools keep their names. Documented in the capability README.

## Linked Issue

https://github.com/pydantic/pydantic-ai-harness/issues/403

Fixes #403

## Checklist

This PR moves our library built at Vstorm into pydantic-ai-harness https://github.com/vstorm-co/pydantic-ai-todo


pydantic_ai_harness/planning/_capability.py
```diff
@@ -47,45 +53,69 @@ class Planning(AbstractCapability[AgentDepsT]):
     """
 
     guidance: str | None = None
```
adtyavrdhn, line 71: Why does None use the default, that looks confusing

adtyavrdhn: I know it is prior design but still pointing it out for your opinions :)

dsfaccini:
This review was posted by Claude Code using claude-opus-4-8 on behalf of David.

*Review level: David directed this review and the fix below, and is making the merge call. He has not line-read the diff himself.*

Hi @DEENUU1 -- thanks for this, folding `pydantic-ai-todo` into `Planning` is a nice consolidation and the cache-safe tail-reminder design carries over cleanly. We reviewed the whole thing and want to help land it. Local gates all pass on our side: `make lint`, `make typecheck` (pyright strict, whole repo), and `make testcov` (100% branch coverage). Lifecycle checks out too -- `for_run` returns the per-run clone and `get_*` re-extracts from it, so per-run store isolation is sound.

One real (but minor and recoverable) behavioral finding, plus four non-blocking notes.

### The finding: a cancelled prerequisite parks its dependent in `blocked`

With `enable_subtasks=True`: if B depends on A and you **cancel** A (instead of completing it), B stays `blocked` and can't be moved forward through the granular tools -- `update_task_status(B, in_progress)` is rejected, `pending` gets re-blocked by `_sync_dependency_blocks`, and B drops out of `get_available_tasks`. It's recoverable (mark B done directly, or `write_plan` to clear the dependency), so it's not a hard deadlock -- but a cancelled prerequisite will never complete, so it shouldn't hold up its dependents.

Mechanism: `is_blocked` (and the mirrored check in `set_dependency`) counted any non-`completed` dependency as blocking. Everywhere else in the state machine already treats `cancelled` as terminal (the all-done note, `get_available_tasks`, and `set_dependency`'s dependent-side guard) -- these two dependency checks were the only holdouts.

We already have a **tested fix ready**: a small `is_terminal(status)` helper (`completed | cancelled`) that both dependency-resolution sites route through, so all sites agree, plus a regression test covering the exact scenario (cancel a prereq -> dependent unblocks to `pending`, shows up in `get_available_tasks`, and can be started). 100% branch coverage holds; lint/pyright/testcov green.

We can't push it to your branch because "Allow edits from maintainers" is off on this PR. Could you enable it so we push the fix directly? Or, if you'd rather apply it yourself, say the word and we'll paste the diff here -- your call.

### Non-blocking notes (observations, nothing required to merge)

1. `active_form` is accepted by `add_task`/`add_subtask` and stored, but no render path surfaces it -- only event consumers see it. If that's intended (matching Claude Code's `activeForm` convention), all good; just flagging in case it was meant to show.
2. `write_plan` is event-silent (bulk `set_items` emits nothing) -- already documented in the `PlanStore` docstring; noting it for apps that render purely off events.
3. `SqlitePlanStore`'s docstring says "safe to share across concurrent tasks," but `update_item` is get-then-write across two lock acquisitions (last-write-wins / phantom event on a since-deleted row). Only matters under concurrent same-session writes, which the intended usage avoids -- might be worth aligning the docstring with the honest Redis/Postgres caveats.
4. `PostgresPlanStore.add_item` computes `seq` via non-transactional `MAX(seq)+1` -- same concurrent-same-session caveat.

Verdict: reviewed, one minor recoverable finding with a fix in hand, four non-blocking notes. Happy to push the fix as soon as edits are enabled.

pydantic_ai_harness/planning/_capability.py
DEENUU1: None = "use the default", '' = "no guidance at all" -- two states so you can opt out on purpose instead of by accident. It's the same convention memory, exa, and runtime_authoring already use, so planning's just matching them. Happy to spell the three states out in the docstring if that helps

dsfaccini:

Non-blocking follow-up design thought: the richer plan store makes a fresh-context execution handoff possible, but I think that orchestration should remain separate from `Planning`.

A closely named `PlanAndHandoff` capability could represent the higher-level workflow:

1. A planner agent investigates and writes a structured plan, but does not implement.
2. When the plan is ready, the workflow starts a fresh executor run with no planner message history.
3. The executor receives a compact handoff packet (objective, repository/issue/PR context, constraints, decisions, acceptance criteria, and unresolved questions) and the same persistent `PlanStore`.
4. The executor reads and updates the plan while implementing and verifying it.

Within `PlanAndHandoff`, fresh execution would be the sensible default, with an explicit option for same-agent execution where desired. `Planning` itself should stay neutral because it is also useful as a live checklist within one run.

There are already two viable substrates:

- host-controlled planner and executor `Agent.run()` calls provide the hardest context boundary;
- `SubAgents(shared_capabilities=[planning])` provides a fresh child context while sharing the plan, although the result returns to the still-live parent.

My inclination would be to first document and dogfood those two compositions, then decide whether `PlanAndHandoff` should be a full capability or a smaller typed `PlanReady` handoff packet plus host-side helper. No change requested for this PR; this seems like a focused follow-up.

gaby:
Agree with @dsfaccini having a separate `Plan` only capability would be very useful.

For example, I can use a model with a 1M context to plan, but execute with one running with a shorter context (e.g., 256K).

DEENUU1:
Agreed, and that matches how I'd want to keep this PR scoped — `Planning` stays a neutral checklist, the orchestration lives one layer up.

Worth noting that the substrate for @gaby's case already works as of this PR, with no new capability: an explicitly passed `store` is shared across runs (only the default `InMemoryPlanStore` is fresh-per-run). So a planner run and an executor run on **different models** can share one plan today:

```python
store = SqlitePlanStore(database, session=session)

planner = Agent('...1m-context-model...', capabilities=[Planning(store=store, enable_subtasks=True)])
executor = Agent('...256k-model...', capabilities=[Planning(store=store, enable_subtasks=True)])

await planner.run(objective)          # writes the plan, implements nothing
await executor.run(handoff)           # fresh history, same plan, ticks items off
```

Which I think sharpens what's actually missing: not storage, but (a) the handoff packet and (b) the convention that the executor starts clean. That's the part worth designing.

### Prior art: the planner sub-agent in pydantic-deep

I've been running a version of step 1 for a while in [pydantic-deep](https://github.com/vstorm-co/pydantic-deepagents) — a built-in **planner** sub-agent, enabled with `include_plan=True`, documented under [plan mode](https://vstorm-co.github.io/pydantic-deepagents/advanced/plan-mode/) ([source](https://github.com/vstorm-co/pydantic-deepagents/blob/main/pydantic_deep/features/plan/toolset.py)). Shape:

- It's registered as a sub-agent under the task tool, so the main agent delegates to it rather than switching modes.
- It gets the read-only console tools (`read_file`, `ls`, `grep`, `glob`) for exploration, and is told firmly that it **only plans** — never edits, never runs commands.
- Two tools of its own: `ask_user(question, options)` for clarifying questions, and `save_plan(title, content)`.

Three things I'd carry into the handoff design, and one I wouldn't:

1. **The read-only constraint earns its keep.** A planner that *cannot* implement produces noticeably better plans than one that could have just started coding. Worth making it a property of the planner run, not a line in the prompt.
2. **Clarifying questions belong in the planning phase, and the answers belong in the packet.** "Decisions made, and why" is the field the executor leans on hardest — without it, it re-litigates choices the planner already settled. That maps directly onto @dsfaccini's `decisions` and `unresolved questions` fields.
3. **`ask_user` degrades instead of hanging.** The tool looks for a callback on deps; when there isn't one — scripts, CI, background jobs — it auto-selects the option the planner marked `recommended` and records `[Auto-selected: JWT]` in the transcript rather than blocking. That's what makes a two-phase flow safe to run unattended, and it's cheap: the planner is required to offer 2–4 concrete options with one marked recommended, so there's always a defensible default. I'd want the same property in the packet — an unresolved question should carry a proposed answer, not just a question mark.
4. **What I'd drop is the markdown plan.** `save_plan` writes a file to `/plans`, and that file is a dead artifact: the executor reads it once, and afterwards there's no shared state, so progress lives only in the executor's own head. `PlanStore` fixes exactly that, and it's the main reason I think the handoff belongs on top of this PR rather than as a separate plan format.

The other limitation there is the one @dsfaccini already identified in substrate (b): the sub-agent gets a fresh child context, but the result returns to a still-live parent, so you only get half the context saving. For the 1M-plan / 256K-execute split you want the hard boundary of two host-controlled `Agent.run()` calls.

So +1 to documenting and dogfooding both compositions first. My inclination is that it lands as a typed `PlanReady` packet plus a host-side helper rather than a full capability — the packet is the part with real design content, and once it exists the orchestration is a dozen lines the host can write itself, in whatever shape their app already has. Happy to open a follow-up issue with the packet fields and a worked example of both substrates, if that's useful.

gaby:
@DEENUU1 It would be great to have that example in the docs. I didnt know you could do that with existing harness.

DEENUU1:
@gaby done -- pushed it to the capability README, under [Persistence](https://github.com/pydantic/pydantic-ai-harness/blob/f0e4920/pydantic_ai_harness/planning/README.md#planning-and-executing-in-separate-runs) as "Planning and executing in separate runs".

pydantic_ai_harness/planning/_capability.py
DEENUU1: Spelled it out, as offered. The docstring now lists the three states and why there are three: `None` uses the built-in guidance, `''` turns it off, and any other string replaces it.

A single `str | None` with `None` meaning "no guidance" would leave no way to ask for the default, and would turn any config that resolves to `None` into a silent opt-out rather than a deliberate one. Pushed in 8c5faef.

dsfaccini:
This comment was posted by Claude Code using claude-opus-5 on behalf of David.

David has not reviewed these findings in detail -- they come from an automated review pass, so please push back on anything that looks wrong.

Reviewed at `8c5faef`. CI green across 3.10 through 3.14 on slim / all-extras / lowest-versions; the 100% branch bar is met. All 13 Macroscope inline findings are resolved, and the substantive ones were spot-verified against the head (batch validation against a projection in `update_task_statuses`, `validate_hierarchy` rejecting dangling and cyclic `depends_on`, `blocked` suppressing the all-done note, re-blocking when a prerequisite regresses, `remove_task` cascading descendants and stripping dangling refs, `from_spec` raising on an unknown backend, `write_plan` deep-copying input items). None of those are re-raised.

`codex/pr-404-fixes` is fully merged into the PR head (`git merge-base --is-ancestor` confirms), so the `is_terminal` cancelled-prerequisite fix, `_refuse_inert_block`, `_stored_status`, and the cross-backend store invariants are all live in the diff. Nothing from that branch is outstanding.

## Blocking

**1. The tool schema offers `blocked`, `parent_id`, and `depends_on` in default mode, then rejects or silently drops them.**

`_types.py::PlanItem` and `_toolset.py::PlanningToolset.write_plan`. Both paths run against `Planning()` with subtasks off:

- `write_plan([PlanItem(content='x', status='blocked')])` returns `"Plan not updated: status 'blocked' is only valid with subtasks enabled."` The enum in the emitted JSON schema lists `blocked`, so the model is invited into a call that discards its entire plan.
- `write_plan([{id:'a',content:'a'}, {id:'b',content:'b',depends_on:['a'],parent_id:'a'}])` returns `"Plan updated: 2 step(s)"`, but both fields are stored empty. The model is told the write succeeded.

The field descriptions say things like "Parent step id for a subtask hierarchy" with no mode caveat; only the class docstring mentions "subtasks mode only".

net-positive -> do. A schema that advertises what the tool refuses costs a wasted turn per occurrence, and the silent drop reports success for a write that did not happen.

**2. `PostgresPlanStore._ensure_schema` has no lock, unlike its sibling.**

`_postgres.py::PostgresPlanStore._ensure_schema` checks a bare `self._ready` flag around an `await`. Driving four concurrent `get_items()` through a pool whose `execute` yields (as a real driver does) issued **four** `CREATE TABLE IF NOT EXISTS` statements.

`memory/_postgres.py::PostgresMemoryStore._ensure_schema` guards the identical pattern with an `anyio.Lock` plus a double-check.

net-positive -> do. The fix is roughly six lines copied from the sibling store, and concurrent `CREATE TABLE IF NOT EXISTS` in Postgres is not race-free.

**3. `docs/planning.md` has drifted behind `planning/README.md`.**

The README is excellent and current. The docs page is missing:

- the `pydantic-ai-todo` supersede-and-migration note (the tool-rename table)
- the subtasks and dependencies tool table (compressed to one sentence)
- the "Planning and executing in separate runs" handoff example
- "Why whole-plan replacement"
- the Configuration block

The handoff example is the one @gaby explicitly asked to see in the docs, and it landed in the README only. The docs site is where a `pydantic-ai-todo` user searching for a migration path arrives.

net-positive -> do. The repo gates on README/docs parity, and the missing piece is precisely the archive story.

**4. The default tool surface grows from 1 to 6 with no opt-out.**

Measured via `FunctionModel`: `Planning()` goes from **1 tool / 1,101 chars** of emitted schema to **6 tools / 5,823 chars** (9 tools / 7,512 chars with `enable_subtasks=True`). Existing users get five tools they did not ask for, roughly 5x the schema in the cached prefix, and a longer default `guidance` string. `descriptions=` lets you reword tools but not remove them.

net-positive -> do, but this is a design decision rather than a defect. Either add a flag (`granular_tools=False`, or a `tools=` allowlist) or decide explicitly that the richer surface is the new default and state the cost in the README so it is visible.

## Non-blocking

**5. A store outage aborts every agent run, including runs that never touch planning.**

`_capability.py::Planning.wrap_model_request` calls `resolve_store(ctx).get_items()` unconditionally on every model request. Confirmed with a store whose `get_items` raises `ConnectionError`: the healthy run succeeds, and the next run dies with `ConnectionError: redis is down` before reaching the model. On `main` the store was always in-process, so this failure mode is new.

net-positive -> do. At minimum document it. The alternative (skip the reminder on store failure) is a behavior choice worth deciding deliberately rather than inheriting.

**6. `subtasks` is a dead parameter** on `_toolset.py::render_summary` and `::status_counts_line` -- threaded through both signatures and never read.

net-positive -> do. Deletion-only.

**7. No way to delete a plan.** `PlanStore` has no `clear` / `delete_session`. `SqlitePlanStore(session=user_id)` and `PostgresPlanStore` accumulate rows per session indefinitely with no API to purge; Redis at least has `expire_seconds`.

net-positive but out-of-scope -> defer, and this one needs a tracking issue to be a valid defer, since it opens a lifecycle-API question rather than being a one-liner.

**8. Duplicate-id handling diverges across backends.** `PlanStore`'s protocol docstring promises `ValueError`; `InMemoryPlanStore` and `RedisPlanStore` raise it, but `SqlitePlanStore` / `PostgresPlanStore` rely on the primary key and surface a driver `IntegrityError` / `UniqueViolationError`. Only reachable by calling `add_item` directly, since both `add_task` and `add_subtask` mint fresh uuids, so no tool path hits it.

net-neutral -> skip. The docstring already hedges; worth one word of alignment if someone is in the file anyway.

**9. Redis stores the whole plan as one JSON document**, so every `add_task` / `update_task_status` is a full read-modify-write of the entire plan. The library used a hash plus an order list, making per-item writes independent. This is documented honestly in the `_redis.py` module docstring, and the advertised planner/executor handoff is sequential so it does not bite there.

net-neutral -> skip. The simpler shape is defensible for the documented usage.

**10. `cast()` in tests.** `tests/planning/test_planning.py::_ctx` uses `cast(RunContext[None], MagicMock())`. This is the only `cast(` anywhere in `tests/` or `pydantic_ai_harness/`, and CLAUDE.md says no typecasting.

net-neutral -> skip. Confined to one test helper, and the real behavior is also covered end-to-end through `Agent`.

**11. `from pydantic_ai_harness.planning._postgres import _deleted_count`** in `tests/planning/test_postgres.py` imports an underscore-prefixed helper, which CLAUDE.md forbids (it prescribes `# pragma: no cover` instead). Precedent exists in the repo (`filesystem`, `step_persistence`, `overflowing_tool_output` all do it) and `pyproject.toml` sets `reportPrivateUsage = false` for tests.

net-neutral -> skip. Consistent with existing practice; flagged only because the rule is explicit.

**12. Around 30 `# type: ignore[union-attr]` comments** in the planning tests for the `(await store.get_item(x)).status` pattern. The repo prefers `# pyright: ignore` (126 uses versus 27 mypy-style), and a small `_stored(store, id)` helper would remove all of them.

net-neutral -> skip. Cosmetic.

## Backward compatibility

- **Public API**: `PlanItem`, `Planning`, `PlanningToolset`, `TaskStatus` all still import from `pydantic_ai_harness.planning` and from the `experimental.planning` shim. One documented break, already called out in the PR body: `PlanningToolset(state)` becomes `PlanningToolset(capability)`, and `PlanState` is gone. `write_plan` also gains a `ctx` first parameter. Only affects direct toolset construction.
- **Tool names**: no rename. `write_plan` keeps its name and its `items: list[PlanItem]` shape.
- **Tool schema**: changed for every existing user (see finding 4).
- **Stored data shape**: no migration needed. The old plan lived in a per-run in-memory `PlanState`, so there is nothing on disk to migrate. All persistence is new.
- **Defaults**: `Planning()` still means a fresh in-memory plan per run; `for_run` isolation verified.

## On archiving `pydantic-ai-todo`

Feature-wise yes, with two named gaps. Inventoried against v0.2.7 on PyPI: all 9 tools, the `blocked` status, auto-block, cycle detection, `get_available_tasks`, the `active_form` field, the "Do NOT call read_plan again" anti-loop terminator, batch all-or-nothing status updates, session-scoped multi-tenancy, the 5 event types with sync and async callbacks, per-tool description overrides, per-run store resolvers, and agent-spec serialization all carry over. The harness version is better on several axes: auto-unblock on prerequisite resolution, `cancelled` as terminal, dependency validation in `write_plan`, and a SQLite backend the library never had.

Two things that will surprise a migrating user:

1. **No connection-string convenience.** `AsyncPostgresStorage(connection_string=...)`, `AsyncRedisStorage(url=...)` and the `create_storage(backend=...)` factory are gone; you construct your own asyncpg pool or Redis client. Deliberate (it keeps harness driver-free and matches `memory`), but it is the migration step most users hit and the README does not mention it.
2. **`PlanEvent` drops `timestamp`.** `TodoEvent` carried a UTC `timestamp`; `_events.py::PlanEvent` has only `event_type`, `item`, `previous_state`. Any consumer that logged or ordered by it has to add its own clock.

Minor: the library exported 10 prompt-string constants publicly; the harness equivalents (`WRITE_PLAN_DESCRIPTION` and friends in `_toolset.py`) are not re-exported from `__init__.py`. The `descriptions=` dict covers the use case.

## Cross-check against #474

No shared abstraction is being missed. `spend/_store.py::SpendStore` is a two-method accumulator (`get` / `add` over money counters with TTL windows); `planning/_store.py::PlanStore` is six-method CRUD over an ordered list. Their only overlap is the pattern -- a `Protocol` for the client, an in-memory default, a Redis implementation -- and that pattern predates both PRs in `pydantic_ai_harness.memory`.

net-negative -> skip on unifying them. The two protocols have nothing in common beyond speaking to Redis, and a shared base would couple unrelated capabilities' optional surfaces.

The one real duplication is the name `RedisClient`, which will exist in both `pydantic_ai_harness.planning` and `pydantic_ai_harness.spend` with different method sets. Harmless given submodule-scoped imports, but worth knowing if a third capability arrives.

## Design fit

Consistent with the repo, not novel. Runtime `CREATE TABLE IF NOT EXISTS` plus caller-owned driver protocols is what `memory`, `step_persistence` and `media` already do. Nothing reimplements core semantics: this is application state the model manipulates through tools, not agent-run state, so it does not collide with pydantic-ai's durable-execution or step-persistence primitives. No stop-and-propose-core-change signal.

The follow-up orchestration idea (David's `PlanAndHandoff` comment, @gaby's 1M-plan / 256K-execute split, and the `PlanReady` packet counter-proposal) is correctly kept out of this PR; all three participants converged on `Planning` staying a neutral checklist.

## Process note

"Allow edits from maintainers" is off on this PR, which is why the `codex/pr-404-fixes` work had to be relayed rather than pushed. If the four blocking items above should be fixed maintainer-side rather than by the author, that toggle is the blocker.

dsfaccini:
This comment was posted by Claude Code using claude-opus-5 on behalf of David.

Pushed the four blocking items plus the two cheap non-blocking ones straight to this branch (`8c5faef..c5afa9c`, three commits). @DEENUU1 -- sorry for landing on your branch rather than relaying; David okayed the direct push this round. Shout if you'd rather I had opened it as a separate PR for you to cherry-pick.

I re-validated every finding against `8c5faef` before touching anything; all four blocking ones reproduced.

**1. Schema advertised what the tool refused.** `write_plan` now refuses `parent_id`/`depends_on` with subtasks off, the same way it already refused a `blocked` status, instead of stripping them and replying `Plan updated`. The `PlanItem` field descriptions carry the mode caveat, so the emitted schema stops inviting the call in the first place.

**2. Postgres schema race.** Ported the guard from `memory/_postgres.py`: `anyio.Lock` + double-check + `pg_advisory_xact_lock` on the table name. The fake pool now yields inside `execute`/`fetchval`, which is what a real driver does -- without that the regression test can't see the interleaving. It reproduces the exact symptom (4 concurrent `get_items()` -> 4 `CREATE TABLE`) and fails on the old store.

**3. Docs drift.** `docs/planning.md` picks up the supersede note and the tool-rename table, the subtasks tool table, the planner/executor handoff example @gaby asked for, "Why whole-plan replacement", and the Configuration block. Also wrote up the two migration surprises from the `pydantic-ai-todo` inventory -- no connection-string convenience, and `PlanEvent` dropping `timestamp` -- since those are the steps a migrating user actually trips on.

**4. Tool surface 1 -> 6.** Added a `tools=` allowlist rather than a bool, keyed by the same names as the existing `descriptions=` dict, so you can trim to any subset instead of just all-or-nothing:

| | tools | schema |
|---|---|---|
| `Planning()` | 6 | ~5.7k chars |
| `Planning(enable_subtasks=True)` | 9 | ~7.2k chars |
| `Planning(tools=['write_plan'])` | 1 | ~2.5k chars |

The built-in guidance follows the allowlist, so a `write_plan`-only agent isn't told about five tools it doesn't have. Default and subtasks guidance are byte-identical to before. Unknown names in `tools=` or `descriptions=` now raise instead of being silently inert -- a typo'd `descriptions` key was previously ignored.

**6.** Dead `subtasks` param dropped from `render_summary` / `status_counts_line`.

### Pushed back on 5

`Planning` keeps failing loud on a store outage; no `on_store_error` knob. `PlanStore` is a protocol, so retry/degradation policy belongs to the store implementation rather than to a field on the capability -- and a plan the model can no longer see isn't a state to keep running in silently. Both README and docs now say so, with a wrapper sketch for anyone who wants a degraded-but-alive mode.

### Still open

**7 (no way to delete a plan)** is unresolved. It's a real gap -- `SqlitePlanStore`/`PostgresPlanStore` accumulate rows per session with no purge API -- but it opens a lifecycle-API question rather than being a one-liner, so it wants its own issue and PR. Happy to file it if you haven't.

Findings 8-12 stay skipped, as triaged.

Verification: `make lint`, `make typecheck` (pyright strict), and the full suite green locally; planning is at 100% branch coverage.

dsfaccini:
my two cents, if it can be translated to the delete plan question: it's just a file in the end, right? so an agent would be able to delete it if they want to. analogously the other storage backends.

another question would be whether they should be autodeleted after some time. not a bad idea as well.

the reality is plans are made to be written and read once and discarded forever, thats the optimal plan experience. any recorded decisions made in the plans belongs in the code/docs they outline.

and because they are relatively small text files they are not a concern of data bloat really.

so yeah, not sure what to make of this but the core point is: plans are discardable (that's their recommended nature)

dsfaccini:
This comment was posted by Claude Code using claude-opus-5 on behalf of David.

Filed the `PlanAndHandoff` follow-up as #512 so it survives this PR being merged -- the packet fields, both substrates, and @DEENUU1's pydantic-deepagents prior art (the read-only planner, clarifying answers in the packet, `ask_user` degrading to a recommended default, and dropping the markdown plan file in favour of `PlanStore`).

@DEENUU1 you offered to write this one up; I collected rather than originated it, so please correct anything I flattened -- or take it over outright.

PR #417, opened 2026-07-22
feat(aws-lambda): AWSLambdaDurability for AWS Lambda durable functions

DouweM:
## Summary

`AWSLambdaDurability` checkpoints an agent's I/O into [AWS Lambda durable functions](https://docs.aws.amazon.com/lambda/latest/dg/durable-functions.html) steps, so an invocation that times out, fails, or is retried resumes from the last completed step instead of repeating work already paid for.

AWS has an example of this in their own SDK repo ([aws/aws-durable-execution-sdk-python#552](https://github.com/aws/aws-durable-execution-sdk-python/pull/552)) that demonstrates the sync-to-async bridge. This is the full-engine version of that idea, built to the bar set by `AbsurdDurability` (#399).

### The bridge

Lambda's durable API is synchronous, `context.step(...)` blocks, and every step must be created on the thread Lambda invoked. An agent run is async. `run_durable` hosts the run on a background event loop and services its steps on the handler thread, so all steps are created in one continuous sequence; a step body hands its work back to the agent loop and blocks, keeping the loop free.

Publishing the active bridge in a `ContextVar` is what lets the capability be attached at **agent construction**, the way every in-tree engine is attached, instead of being handed a `DurableContext` per invocation.

**Correction to an earlier version of this description:** I originally claimed a per-run capability cannot checkpoint tool calls. That is wrong -- `Agent.run(capabilities=[...])` calls `for_agent`, so `_register_toolsets` runs and the toolset swap happens at run setup; I verified per-run attachment checkpoints tool calls fine. Construction-time attachment is still the right recommendation (wrapping happens once, the deployed shape is stable, and it matches the other engines), but not for that reason. What the AWS example actually misses is unrelated to attachment timing: it hooks `wrap_tool_execute` rather than wrapping toolsets, so `get_tools` and `get_instructions` are never checkpointed and MCP/dynamic toolsets re-run their I/O on every replay.

### Built on the shared scaffolding

No parallel implementation and no engine-specific toolset subclasses: `BaseDurabilityCapability`, `DurableModel` + segment executors, `DurableFunctionToolset`/`DurableMCPToolset`/`DurableDynamicToolset` via engine factories, `wrap_tool_call_result` + `unwrap_recorded_tool_call_result` for control-flow round-tripping, `resolve_tool_durable_config` for per-tool config, and `EnqueueGuard` for in-step `ctx.enqueue()`.

Behaviour worth calling out:

- **Control-flow signals cross a step as values.** `ModelRetry`, `ApprovalRequired`, and `CallDeferred` escaping raw would be recorded as *failed* steps, retried, and then fail the execution -- which breaks approval and deferred-tool flows. Covered by tests.
- **Tool results are encoded by Pydantic before checkpointing**, so `ToolReturn` and `BinaryContent` round-trip rather than failing the SDK serializer.
- **Model continuation segments checkpoint individually** (Anthropic `pause_turn`, OpenAI background mode), so a long continuation loop is not one step against the 15-minute invocation cap.
- **Transparent outside a durable handler**, including tool-call parallelism, which is only forced sequential inside a durable execution.
- **Nested durable runs are rejected** with an explanatory error instead of deadlocking the handler thread.

## Linked Issue

None yet -- opening this directly at DouweM's request so it can be shared with the AWS team.

## :warning: Dependency change (needs a maintainer wave-through)

This **does** change `pyproject.toml` and `uv.lock`, which the checklist normally forbids:

- adds an `aws-lambda` extra: `aws-durable-execution-sdk-python>=1.7.0`, carrying a `python_version >= "3.11"` marker because the AWS SDK requires 3.11 while the harness floor stays 3.10;
- adds `pydantic-ai-slim>=2.15.0` (needs the shared `DurableDynamicToolset` from pydantic/pydantic-ai#6623);
- adds `mcp` to the dev group so the MCP paths are testable and coverage-measured, matching what #399 did.

Full-repo `pyright` and the full test suite pass with the bump (2914 passed, 38 skipped).

## Testing

36 tests, no Docker and no network: a recording `FakeDurableContext` that round-trips every stored value through the **SDK's real serializer**, so a value that could not be checkpointed in production fails in the test too. Covers checkpointing, resume-from-checkpoint, crash-mid-run retry (completed model step reused while the failed tool re-runs), control-flow signals, tool-result serialization, MCP, dynamic toolsets, per-tool step config, transparency outside a handler, and the nesting guards.

100% branch coverage on the new package.

## Checklist

- [ ] ~Linked issue exists and is referenced above~ -- see above
- [x] `lint && typecheck && test` pass locally
- [ ] ~No changes to `pyproject.toml` or `uv.lock`~ -- see the dependency note above


pydantic_ai_harness/aws_lambda/_bridge.py
```diff
@@ -0,0 +1,332 @@
...
+from pydantic_ai.exceptions import UserError
+
+if TYPE_CHECKING:
+    from aws_durable_execution_sdk_python.config import StepConfig
+
+T = TypeVar('T')
+
+ENGINE_NAME = 'AWS Lambda'
+
+_CANCEL_TIMEOUT_SECONDS = 5.0
```
nvasiu: Is the 5 second timeout is an arbitrary value? If a cancelled run's cleanup runs longer, the handler will return while the cleanup is still happening in the shared loop, so a previous invocation's work could still leak into the next warm-invocation. Could we instead make the timeout configurable, or add a check for the previous invocation's clean up to complete before starting the next one?

pydantic_ai_harness/aws_lambda/_bridge.py
```diff
@@ -0,0 +1,332 @@
...
+    try:
+        loop.call_soon_threadsafe(schedule_run)
+        return bridge.consume()
+    finally:
+        # The loop outlives the invocation, so a run abandoned by a suspension or an error escaping
+        # the handler would otherwise keep running into the next warm invocation. Wait for it to
+        # finish unwinding: an agent coroutine's `finally`/`__aexit__` cleanup runs during
+        # cancellation, and letting that overlap the next invocation would touch shared provider
+        # resources after the execution it belonged to was abandoned.
+        started.wait()
```
nvasiu: This wait is unbounded. It's unlikely but possible for the loop to get stuck between calling `schedule_run` and it setting `started`. This would block the handler until the Lambda itself times out, which would return an opaque error message.

pydantic_ai_harness/aws_lambda/_bridge.py
```diff
@@ -0,0 +1,332 @@
...
+                    # what `create_task(context=...)` does on 3.11+, spelled so it also type-checks
+                    # against the repo's 3.10 target.
+                    task: asyncio.Task[T] = step_context.run(lambda: loop.create_task(operation()))
+                except BaseException as exc:  # pragma: no cover - task creation failing is not reproducible
+                    result.set_exception(exc)
+                    return
+                task.add_done_callback(lambda finished: _forward(finished, result))
+
+            loop.call_soon_threadsafe(schedule)
+            return result.result()
```
nvasiu: Unbounded wait here. If the agent loop dies before the task completes, the handler will get blocked forever. But adding a timeout wouldn't work here because steps can legitimately run long. Maybe could detect if the loop is still alive?

pydantic_ai_harness/aws_lambda/_capability.py
```diff
@@ -0,0 +1,470 @@
...
+        self._default_model_id: str | None = None
+
+    @property
+    def in_durable_context(self) -> bool:
+        return in_durable_context()
+
+    def _bind_to_agent(self, agent: AbstractAgent[AgentDepsT, Any]) -> None:
+        # Lambda durable steps are ad-hoc `context.step(...)` calls, so there is nothing to
+        # register up front beyond the durable toolset wrappers.
+        self._default_model_id = agent.model if isinstance(agent.model, str) else None
```
nvasiu: When the agent is built with a `Model` instance (e.g. `Agent(BedrockConverseModel(...))`) rather than a string id, `_default_model_id` stays None. So in `wrap_model_request` the `model_id == self._default_model_id` suffix suppression will never match, and default model steps won't get shortened names. Could we derive `_default_model_id` from the `Model` instance id in this case?

pydantic_ai_harness/aws_lambda/_capability.py
```diff
@@ -0,0 +1,470 @@
...
+                    request.messages, request.model_settings, request.model_request_parameters
+                )
+                return _response_adapter.dump_python(response, mode='json')
+
+            payload = await _require_bridge().run_step(f'{self.name}__model.request{suffix}', operation, config)
+            return _response_adapter.validate_python(payload)
+
+        async def request_stream_segment(request: ModelRequestContext) -> StreamedActivityResult:
+            async def operation() -> _JsonValue:
+                async with request.model.request_stream(
```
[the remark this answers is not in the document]
nvasiu: Agree with this comment.

pydantic_ai_harness/aws_lambda/_capability.py
```diff
@@ -0,0 +1,470 @@
...
+def _guard_enqueue(ctx: RunContext[AgentDepsT]) -> RunContext[AgentDepsT]:
+    """Make `ctx.enqueue()` raise inside any code that runs within a durable step.
+
+    A replay serves the recorded step output without re-running the code, so messages enqueued
+    from inside the step -- whether a tool call or an `event_stream_handler` delivering events
+    inside the model step or its own handler step -- would be dropped.
+
+    On `pydantic-ai` 2.16.0 this collapses onto the base's `_durable_run_context` /
+    `_durable_run_context_scope` (pydantic/pydantic-ai#6671), which guards both the passed and the
+    ambient run context from one place; this local guard covers the passed context on the 2.15.0 floor.
```
nvasiu: Just to confirm, this means that on 2.15.0 the ambient run context is still unguarded?

pydantic_ai_harness/aws_lambda/_bridge.py
DouweM: Arbitrary, yes — it was a guess at "long enough for ordinary `__aexit__`/`finally` cleanup, short enough to leave the handler room to return." Both of your suggestions are now in, and the second one is what makes the first one safe:

- `run_durable` takes a `cancel_timeout` (still 5s by default), documented as a latency knob.
- Exceeding it no longer leaks. The loop is **retired** rather than reused: `_AgentLoop.retire()` drops the reference, so the next invocation builds a fresh loop with its own loop-bound resources and the leftover cleanup cannot reach into it.

I went with retiring rather than "wait for the previous cleanup before starting the next invocation" because that wait would be unbounded on the *entry* side — a hung cleanup would then wedge the next invocation until Lambda timed it out, which is the failure mode this file exists to avoid. Retiring gives the same isolation with a bounded handler.

The loop is deliberately left *running* when retired: the abandoned run is still unwinding on it, and stopping it out from under a `finally` would strand whatever that cleanup holds.

Covered by `test_an_unwind_that_outlives_the_cancel_timeout_retires_the_loop`.

pydantic_ai_harness/aws_lambda/_bridge.py
DouweM: Fixed, and digging into it turned up a worse bug behind it.

`started` is now set in a `finally`, so the wait cannot outlive a `schedule_run` that raised. But the real hazard was upstream: if `loop.create_task(...)` raised, nothing would ever call `bridge.finish()`, so `bridge.consume()` — which runs *before* this wait — blocked until the function timed out. `schedule_run` now resolves the bridge with the error in that case, so the failure surfaces as itself instead of as a Lambda timeout.

The `finally` that cancels the run also guards on `tasks` being non-empty now, since a failed scheduling leaves nothing to cancel.

pydantic_ai_harness/aws_lambda/_bridge.py
DouweM: Your diagnosis and your suggested fix were both right — a deadline is wrong here for exactly the reason you give, so the wait now polls for the loop's liveness instead:

```python
while True:
    try:
        return result.result(timeout=_LOOP_LIVENESS_POLL_SECONDS)
    except FutureTimeoutError:
        if loop.is_running() or result.done():
            continue
        raise AgentLoopGone(...)
```

`loop.is_running()` is enough on its own — by the time a step body runs, the loop is definitely up, so there is no startup race to worry about. The `result.done()` check closes the window where the result lands between the poll timing out and the liveness check.

One thing worth flagging, because my first attempt at this deadlocked the test suite: the error **must not** be routed into the agent. `consume()` hands ordinary step failures back to the agent run so it can handle them, and that is precisely what cannot work here — the loop that would receive them is the thing that is gone, so `consume()` just went back to blocking on its queue. Hence `AgentLoopGone` is a `BaseException`, like the SDK's own control flow, and leaves the handler.

Covered by `test_a_dead_agent_loop_fails_the_step_instead_of_blocking_forever` (kills the loop mid-step) and `test_a_step_that_outlasts_the_liveness_poll_keeps_waiting` (pins that the poll interval is not a deadline).

pydantic_ai_harness/aws_lambda/_capability.py
DouweM: Good thing to check, but the default model steps do get short names in that case — just by a different route, which is why this reads as a gap.

`Agent(BedrockConverseModel(...))` leaves `ModelRequestContext.model_id` unset (it is the *selection string* a model was resolved from, and there is none for an instance). So `_model_id_for_request` never takes the provenance fast path that `_default_model_id` guards; it falls through to the base's `_find_model_id`, which matches the instance against the capability's model registry and returns `None` for the agent's own model. `model_id` is already `None`, so there is no suffix to suppress.

A string default is the one case that *does* take the provenance path and so arrives carrying its own name — which is the only reason `_default_model_id` exists.

This was in fact already covered, just invisibly: `build_agent` in the tests builds `Agent(FunctionModel(...))`, so every `assert ctx.step_names == ['a__model.request', ...]` in the file is the instance case. I have made it explicit — `test_a_model_instance_default_keeps_one_step_name` now sits next to the string twin — and put the reasoning in a comment at the assignment so the next reader does not have to re-derive it.

pydantic_ai_harness/aws_lambda/_capability.py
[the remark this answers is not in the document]
DouweM: Agreed and fixed — `request_stream` now gets the guarded context, and it is the same guard the event stream gets, so the model body and the handler are covered by one scope.

The test (`test_enqueue_from_a_streaming_model_body_raises`) uses a `FunctionModel` subclass that enqueues from inside its own `request_stream`; I verified it fails on the pre-fix code rather than just passing alongside it.

pydantic_ai_harness/aws_lambda/_capability.py
DouweM: Two answers, because the docstring was wrong on one point and right on the other.

**The 2.15.0 framing was stale.** The harness's package-level floor is `pydantic-ai-slim>=2.18.0`, so the `>=2.15.0` in the `aws-lambda` extra is dead weight — the resolved floor is always 2.18.0 and the base's guard is always available. That comment dates from when this branch was written against a 2.15.0 floor and I did not revisit it. Removed.

**But you were right that the ambient context was unguarded** — not because of the floor, just because this module only ever guarded the context it *passed*. So `get_current_run_context().enqueue(...)` inside a step was a way around the guard, and those messages were dropped on replay just as silently.

Fixed: `_guard_enqueue` is now `_durable_step_scope`, a context manager that installs the guarded context as the ambient one as well, matching the base's `_durable_run_context_scope`. It stays a module-level function rather than collapsing onto the base method only because the durable toolsets here are built by module-level factories that hold no reference to the capability.

Covered by `test_enqueue_via_the_ambient_run_context_inside_a_step_raises`, which I confirmed fails without the scope.

DouweM:
@nvasiu thanks — really glad to hear the real-Lambda testing came back clean, and this was a sharp review. All six comments are addressed in 9a59a16; replies are in each thread. Summary:

**Fixed, behaviour changes:**

1. **Cancel timeout** — now `run_durable(cancel_timeout=...)`, default still 5s. More importantly, exceeding it no longer leaks into the next warm invocation: the shared event loop is *retired* instead of reused, so the next one starts fresh rather than inheriting a loop mid-teardown. I went with that over waiting for the previous cleanup on the entry side, since that wait would be unbounded in the one case it matters.
2. **Unbounded `result.result()`** — polls for loop liveness instead of imposing a deadline, exactly as you suggested, and raises `AgentLoopGone` if the loop is gone. It has to be a `BaseException`: routing it into the agent is what cannot work when the agent's own loop is the thing that died. My first attempt did route it, and deadlocked the suite — good catch on your part.
3. **Unbounded `started.wait()`** — set in a `finally` now. Behind it was a worse bug: a `create_task` failure meant nothing ever called `bridge.finish()`, so `consume()` blocked until the function timed out. That path now resolves the bridge with the error.
4. **`request_stream` got the unguarded context** (the thread you agreed with) — one guard now covers the model body and the event stream.
5. **The ambient run context genuinely was unguarded**, so `get_current_run_context().enqueue(...)` inside a step slipped past. The guard is now a scope that installs the guarded context ambiently too. The 2.15.0 caveat that prompted your question was itself stale — the harness floor is 2.18.0.

**Not a bug:** `_default_model_id` and a `Model` instance. An instance leaves `ModelRequestContext.model_id` unset, so resolution goes through the base's model registry, which already returns `None` for the agent's own model — there is no suffix to suppress. Details in that thread; there is now a named test for the instance case next to the string one.

Each new guard test is verified to fail against the pre-fix code, rather than merely passing next to it. Full suite green at 100% coverage.

No rush on the team discussion — ping me if anything above needs another pass.

pydantic_ai_harness/aws_lambda/_capability.py
```diff
@@ -0,0 +1,489 @@
...
+        durable_config=base_config,
+    )
+
+
+def _build_mcp_toolset(
+    toolset: MCPToolset[AgentDepsT], *, step_name_prefix: str, base_config: Mapping[str, Any] | None
+) -> DurableMCPToolset[AgentDepsT]:
+    name = f'{step_name_prefix}__mcp_server__{toolset.id}'
+
+    async def get_tools_operation(ctx: RunContext[AgentDepsT]) -> dict[str, ToolDefinition]:
```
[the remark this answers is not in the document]
DouweM: Correct, fixed in `469f1b5`. Both MCP discovery operations and the dynamic toolset's `get_tools` now run inside `_durable_step_scope`.

The dynamic case is the one that matters most and the comment understates it: that operation calls `get_dynamic_tools(toolset, ctx)`, where the toolset factory is *user* code, so it is the discovery path most likely to actually enqueue. Tests: `test_enqueue_from_the_toolset_factory_raises`, plus `test_enqueue_while_listing_tools_raises` and `test_enqueue_while_fetching_instructions_raises` for MCP — each verified to fail on the pre-fix code.

Worth noting this also brings the module in line with the declarative base in pydantic/pydantic-ai#6696, which already wraps all three of these in `_durable_run_context_scope`.

pydantic_ai_harness/aws_lambda/_bridge.py
```diff
@@ -0,0 +1,404 @@
...
+    """
+
+    def __init__(self) -> None:
+        self._lock = threading.Lock()
+        self._loop: asyncio.AbstractEventLoop | None = None
+
+    def get(self) -> asyncio.AbstractEventLoop:
+        with self._lock:
+            loop = self._loop
+            if loop is None or loop.is_closed():
```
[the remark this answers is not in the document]
DouweM: Right, and fixed — but the suggested one-line change would have introduced the race flagged in the test-helper comment, so the fix is slightly larger.

`get()` returning before the thread reaches `run_forever()` means a bare `not loop.is_running()` check can fire on a perfectly good loop that simply has not started yet, discarding it and leaking the thread. So `get()` now waits for the loop to actually be running (via a `call_soon` on the loop thread) before publishing it, which makes `is_running()` a sound reuse test:

```python
if loop is not None and not loop.is_closed() and loop.is_running():
    return loop
loop = asyncio.new_event_loop()
running = threading.Event()

def run(loop = loop) -> None:
    asyncio.set_event_loop(loop)
    loop.call_soon(running.set)
    loop.run_forever()

threading.Thread(target=run, daemon=True, name='pydantic-ai-lambda-agent').start()
running.wait()
```

`test_a_stopped_loop_is_not_handed_to_the_next_invocation` covers it, and fails against the old `is_closed()`-only check.

tests/aws_lambda/test_aws_lambda.py
```diff
@@ -0,0 +1,755 @@
...
+        assert 'a__model.cancel_suspended_response' in ctx.step_names
+
+
+def shutdown(loop: asyncio.AbstractEventLoop) -> None:
+    """Stop and close a loop a test built, so it does not outlive the test as a leak.
+
+    The bridge deliberately never closes a loop itself -- a retired one may still be unwinding an
+    abandoned run -- so tests that make their own `_AgentLoop` clean up after it here.
+    """
+    if loop.is_running():
```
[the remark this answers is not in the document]
DouweM: Real race, and it is the same one as the `_AgentLoop.get()` finding — so it is fixed at the source rather than in the helper. `get()` now waits for the loop to reach `run_forever` before returning it, so by the time any caller (test or production) sees a loop from it, `is_running()` is true. `shutdown` needs no special case: the only loop it ever sees in a not-yet-started state is the deliberately-foreign `asyncio.new_event_loop()` in that test, which is genuinely not running and safe to close.

docs/aws-lambda.md
```diff
@@ -0,0 +1,222 @@
...
+
+- **Steps are at least once, and retried by default.** A step is checkpointed after it runs, so an
+  interruption between a tool's side effect and its checkpoint re-runs the tool when the execution
+  resumes. On top of that, the SDK's default retry policy is six attempts with exponential backoff
+  (5s to 60s), applied to every model request and every tool call. Keep tool side effects
+  idempotent. `AT_MOST_ONCE_PER_RETRY` alone is not enough to make a tool run once: it prevents
+  re-execution after an interruption *within* an attempt, but the retry policy still starts further
+  attempts that do execute the body. For a tool that must not repeat, set both:
+
+    ```python {test="skip" lint="skip"}
```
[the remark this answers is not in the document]
DouweM: Disagreeing on the main claim: `RetryPresets.none()` does exist. It is at `aws_durable_execution_sdk_python/retries.py:200-206` in the pinned SDK (1.7.0) and returns `create_retry_strategy(RetryStrategyConfig(max_attempts=1))` — exactly the single-attempt strategy the snippet advertises. I checked the installed package rather than going from memory:

```
$ uv run python -c "from aws_durable_execution_sdk_python.retries import RetryPresets; print(RetryPresets.none())"
```

I suspect the check looked in `.config`, which is where `StepSemantics` lives but not `RetryPresets`.

The secondary point is fair, though, and it is the same trap: the two names come from *different* modules, which makes the snippet easy to get wrong. Added the imports:

```python
from aws_durable_execution_sdk_python.config import StepSemantics
from aws_durable_execution_sdk_python.retries import RetryPresets
```

No change to the guidance itself — a user copying it gets a tool that runs at most once, as documented.

nvasiu (approved):

pydantic_ai_harness/aws_lambda/_bridge.py
```diff
@@ -0,0 +1,425 @@
...
+
+        Called when an abandoned run outlives the cancellation wait. The loop is deliberately left
+        running: the run is still unwinding on it, and stopping it out from under a `finally` or
+        `__aexit__` would strand whatever that cleanup holds. Dropping the reference means the next
+        invocation builds a fresh loop with its own loop-bound resources, so the leftover cleanup
+        cannot reach into the execution that follows it -- which is the whole reason for the wait.
+        """
+        with self._lock:
+            if self._loop is loop:
+                self._loop = None
```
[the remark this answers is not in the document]
DouweM: Fixed in 3dd5b8c2: retired loops stop accepting work immediately, stop after cleanup or a 5s grace period, and close on their owning thread; slow blocking cleanup closes the loop once it releases. Regression tests prove the retired thread and loop terminate.

tests/aws_lambda/test_aws_lambda_mcp.py
```diff
@@ -0,0 +1,284 @@
...
+            name=self._tool_name,
+            description='Add two integers.',
+            parameters_json_schema=_ADD_SCHEMA,
+            metadata=self._tool_metadata,
+        )
+        return {self._tool_name: self.tool_for_tool_def(tool_def, ctx=ctx)}
+
+    async def get_instructions(self, ctx: RunContext[object]) -> InstructionPart | None:
+        if not self.include_instructions or self._instructions_text is None:
+            return None
```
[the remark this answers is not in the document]
DouweM: Fixed in 3dd5b8c2: get_instructions now calls _require_session() like get_tools/call_tool.

tests/aws_lambda/test_aws_lambda_mcp.py
```diff
@@ -0,0 +1,284 @@
...
+        assert server.tool_calls == [('add', {'a': 2, 'b': 3})]
+
+
+class TestMultipleServers:
+    """`CombinedToolset` lists toolsets concurrently, so two servers issue sibling step requests.
+
+    Sibling steps must queue and run one at a time, not be rejected as if they were nested.
+    """
+
+    def test_two_mcp_servers_are_listed_and_called(self) -> None:
```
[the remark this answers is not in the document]
DouweM: Fixed in 3dd5b8c2: the model sequence now calls add then add2, asserting both servers receive calls.

docs/aws-lambda.md
```diff
@@ -0,0 +1,225 @@
...
+instead of repeating the work already paid for.
+
+Lambda keeps a log of durable operations. When an execution resumes, the handler runs again from
+the top and completed steps return their stored results rather than executing. Without
+checkpointing, a resumed run would repeat every model request and tool call.
+
+## Installation
+
+```bash
+pip install "pydantic-ai-harness[aws-lambda]"
```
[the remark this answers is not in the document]
DouweM: Fixed in 3dd5b8c2: both docs surfaces now install [aws-lambda,bedrock]. The aws-lambda extra also gained pydantic-ai-slim[mcp] (same gap as the Restate finding).

pydantic_ai_harness/aws_lambda/_bridge.py
DouweM: Revisited after the declarative rewrite: the 5s default stays heuristic and configurable, but overlong cleanup is now safe; retire() puts the loop through a bounded shutdown (stop after cleanup or grace deadline, close on the owning thread) instead of leaking a reusable thread. Fixed in 3dd5b8c2.

pydantic_ai_harness/aws_lambda/_bridge.py
DouweM: This was still real in the current code; fixed in 3dd5b8c2 with a 5s bound on the startup wait, an explicit AgentLoopGone error, and loop retirement so the handler cannot block forever.

pydantic_ai_harness/aws_lambda/_capability.py
DouweM: Checked against the current code: a Model instance resolves to model_id=None, the journal namer emits no suffix, and the unit is named {name}__model.request, consistent with the string-default short form and pinned by tests. Not a defect, but say if you'd rather see an explicit marker for instance-configured models.

pydantic_ai_harness/aws_lambda/_capability.py
DouweM: Overtaken by the declarative rewrite: the run-context scoping is now owned uniformly by pydantic-ai's BaseDurabilityCapability (the engine no longer has its own ambient-context path), so the 2.15.0-specific gap this asked about no longer exists on this branch.

docs/aws-lambda.md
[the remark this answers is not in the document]
DouweM: Partial correction, e69a4d85: the docs install fix stands, but the aws-lambda extra's mcp dependency is reverted; MCP stays its own opt-in extra by repo convention (matching in-tree temporal/dbos/prefect), the engine never imports pydantic_ai.mcp, and tests get it from the test dependency group.

DouweM:
@nvasiu, a question about the handler-facing API before we settle it, since you know better than we do how you want people writing durable handlers.

### Where we are

The capability exposes two names beyond `AWSLambdaDurability`, and both come from the bridge:

```python
from pydantic_ai_harness.aws_lambda import AWSLambdaDurability, AgentLoopGone, run_durable

@durable_execution
def handler(event: dict[str, Any], context: DurableContext) -> str:
    result = run_durable(lambda: agent.run(event['prompt']), context=context)
    return result.output
```

`run_durable` is public because an agent run is async and the durable API is not. `DurableContext.step()` is synchronous, has to be called from the handler's own thread, and blocks until the step body returns, so the bridge runs the agent on a background event loop and services its steps on the handler thread, keeping every `context.step(...)` in one continuous sequence on the thread Lambda invoked. Something has to drive that from the handler, and `DurableContext` only exists as the handler's argument, so the user has to hand it to us somewhere. `AgentLoopGone` is public for a narrower reason: it is a `BaseException` that deliberately escapes to the handler, because an ordinary step failure routes back into the agent run and that is exactly what cannot work when the loop that would receive it is the thing that is gone.

### What we would like to add

The part that reads awkwardly is the `lambda:` at the call site. A decorator would let the handler simply be async:

```python
from pydantic_ai_harness.aws_lambda import AWSLambdaDurability, durable_agent_handler

@durable_execution
@durable_agent_handler
async def handler(event: dict[str, Any], context: DurableContext) -> str:
    result = await agent.run(event['prompt'])
    return result.output
```

It is a thin wrapper: it returns the synchronous `Callable[[Any, DurableContext], Any]` that `durable_execution` expects, and calls `run_durable(lambda: func(event, context), context=context)` internally. No new machinery, and `run_durable` stays exported for anyone who wants to keep a synchronous handler or call it somewhere other than the top of one.

The reason we think it is worth more than cosmetics: today anything else async in the invocation needs its own `run_durable` call, and a concurrent second one is rejected outright, because the bridge is one-at-a-time per thread. With an async handler body you get a single loop for the whole invocation, so two agent runs, an `asyncio.gather`, or async post-processing between runs all bridge their steps correctly through the one sequence.

### What we would like your read on

1. **Is double decoration acceptable to you?** It is the obvious cost, and we do not love it either. `@durable_execution` has to stay outermost since it is what Lambda invokes. If you would rather not see two decorators in AWS-facing examples, we would keep `run_durable` as the only entrypoint and drop this.
2. **Is there an SDK-side seam that would be cleaner?** If `durable_execution` could accept a coroutine function directly, or expose a documented hook for adapting one, the adapter belongs on your side rather than ours, and both decorators collapse into one.
3. **Is an async-native `DurableContext` on your roadmap?** There is currently no `async def` anywhere in `aws_durable_execution_sdk_python`, which is entirely reasonable for the model you are targeting, and the bridge exists precisely because of it. If an awaitable `step()` is coming, we would rather build toward that than harden a thread bridge we would then retire.

Happy either way. If you prefer the current shape we will leave it exactly as is; this is a question, not a change we are pushing.

pydantic_ai_harness/aws_lambda/_bridge.py
```diff
@@ -0,0 +1,492 @@
...
+            loop = asyncio.new_event_loop()
+            running = threading.Event()
+
+            def run(loop: asyncio.AbstractEventLoop = loop) -> None:
+                asyncio.set_event_loop(loop)
+                loop.call_soon(running.set)
+                try:
+                    loop.run_forever()
+                finally:
+                    loop.close()
```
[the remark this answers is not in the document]
nvasiu: Confirmed issue. A sync tool from an abandoned run can still be running after retirement, overlapping the next invocation. Along with the fix, could we include a test for this scenario: a sync tool (run through the default executor) is still mid-execution when its run is abandoned, asserting that its executor work does not continue running after retirement into the next invocation.

pydantic_ai_harness/aws_lambda/_bridge.py
```diff
@@ -0,0 +1,492 @@
...
+
+        The retired loop gets a bounded grace period for the abandoned task's cleanup. It stops as
+        soon as cleanup finishes, or at the grace deadline if cleanup refuses to finish. The loop's
+        own thread closes it after `run_forever()` exits.
+        """
+        with self._lock:
+            if self._loop is not loop:
+                return
+            thread = self._thread
+            self._loop = None
```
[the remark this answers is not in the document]
nvasiu: Blocking loop acquisition until the retired thread has terminated would reintroduce the unbounded hang issue described in this comment [above](https://github.com/pydantic/pydantic-ai-harness/pull/417#discussion_r3684339100), so we'd probably want to avoid that.

Loop-bound resources should be safe from cleanup, each new invocation gets a fresh loop with its own loop-bound resources. But we're still worried about the cleanup touching any module-global resources.

Maybe instead of blocking we could document that customers shouldn't share mutable module-global resources between the agent run and the handler. The existing bounded cleanup wait means it usually finishes before the next invocation starts.

nvasiu:
@DouweM  Re: your question about the handler-facing API

Let's go ahead with this solution. I also prefer the double decoration over the current awkward run_durable(lambda) shape. One thing to keep in mind, the @durable_execution decorator needs to be the outermost (it is what Lambda invokes). We could document that order, or error if they are in the wrong order.

pydantic_ai_harness/aws_lambda/_capability.py
```diff
@@ -0,0 +1,137 @@
...
+                inside the model-request step; each agent-level event is handled in its own
+                checkpointed step.
+            name: Unique agent name used as the prefix for every step name. Defaults to the
+                agent's `name` when the capability is bound.
+            step_config: Base `StepConfig` fields applied to every step, as a mapping of
+                `retry_strategy`, `step_semantics`, and `serdes`. Per-tool
+                `metadata={'aws_lambda': {...}}` overrides it key by key for that tool.
+        """
+        super().__init__(models=models, event_stream_handler=event_stream_handler, name=name)
+        AWSLambdaOperationConfig(step_config)
```
nvasiu: `AWSLambdaOperationConfig(step_config)` is built here for validation, but then it's discarded. Then the raw `step_config` is stored in the next line, and it gets validated again in `get_durable_operation_backend`.

Since the `AWSLambdaOperationConfig(step_config)` is not stored, this line looks like dead code. But removing it would remove the validation. You could store the validated object and reuse it, or leave a comment explaining what the line is for?

pydantic_ai_harness/aws_lambda/_operation_backend.py
```diff
@@ -0,0 +1,96 @@
...
+    if not config:
+        return None
+    unknown = sorted(set(config) - _STEP_CONFIG_FIELDS)
+    if unknown:
+        raise UserError(
+            f'Unknown {_TOOL_CONFIG_KEY!r} step config {"key" if len(unknown) == 1 else "keys"} '
+            f'{", ".join(repr(key) for key in unknown)}. Supported keys are '
+            f'{", ".join(repr(field) for field in sorted(_STEP_CONFIG_FIELDS))}.'
+        )
+    return StepConfig(**config)
```
nvasiu, line 107: `StepConfig(**config)` is a plain dataclass constructor, so it does not check value types. The above check catches invalid keys, but it will pass if the customer passes in correctly named keys with wrongly typed values. Then it would cause a failure later when the step runs, far from where the config was defined. Worth validating the field types here so bad values fail at build time?

DouweM:
@nvasiu all six items are pushed, one commit each. Two things need your take, and the rest needs nothing from you.

**TL;DR**

1. **A decision:** an abandoned run's cleanup can overlap the next warm invocation. We have documented that rather than changed it. Option to close it further below, if you want the latency instead.
2. **A pushback:** we did not write the executor test the way you asked, because the assertion cannot hold. Details below.
3. Everything else: done, nothing needed.

---

## 1. The overlap: your call

Reviewing your comment we found our own docs claiming more than the code delivers. `run_durable`'s `cancel_timeout` docstring said:

> Exceeding it is safe: the background event loop is retired instead of reused, so the leftover cleanup **cannot overlap the next warm invocation**.

That clause is false. `retire()` ends with `thread.join(timeout=0.1)` and returns regardless, so the retired thread keeps draining while the next invocation may already be running.

A bounded handler and full isolation are mutually exclusive here: something has to give when cleanup refuses to stop. Three ways to land it:

| | behaviour | cost |
| --- | --- | --- |
| **A. shipped** | keep the 100ms join, document the exposure | customers must avoid shared module-global state |
| **B.** | wait on the retired thread for a real budget | latency on an invocation that already blew `cancel_timeout`; narrows the window, does not close it, so A's docs still apply |
| **C.** | block until drained | the unbounded wedge we both rejected [earlier](https://github.com/pydantic/pydantic-ai-harness/pull/417#discussion_r3684339100) |

We shipped A, per your suggestion: docs and docstring now separate what is guaranteed (fresh loop, so cleanup cannot touch the next invocation's loop-bound resources) from what is not (cleanup can still run during that invocation for the grace period, so module-global state and external side effects are exposed).

**You know what Lambda customers actually hit better than we do, so if you would rather spend the latency, say so and we will do B.** It is a small change.

## 2. The executor test: assertion changed

Fixed the bug itself: `clean_up()` now awaits `loop.shutdown_default_executor()` inside the grace budget before `loop.stop()`.

You asked the test to assert the executor work "does not continue running after retirement into the next invocation". That is the isolation guarantee option A gives up, so the test would either fail or pass on timing luck.

It asserts the reachable version instead, and is explicit about the overlap rather than quiet about it: after `retire()` returns it asserts the retired thread is **still alive** and the loop still open, then that the work completes and the loop closes once released. Object if you would rather have the guarantee, but that is decision 1, not a test change.

## 3. Done, no action needed

- **`durable_agent_handler`** (`0863f45a`), as proposed, plus a `cancel_timeout` knob. On decorator order we error rather than only document, having checked the error can fire: your `durable_execution` wrapper is not a coroutine function, so the reversed order is detectable. **One heads-up:** if a future SDK version makes that wrapper a coroutine function, our check goes quiet, so we would want to know.
- **Discarded operation config** (`8abf56e6`): now validated once and stored, after checking the instance is safe to share.
- **Step config value types** (`922f122a`, `220a29d9`): fixed for both the base config and per-tool metadata, which had the asymmetry you spotted. The follow-up commit is a bug we introduced and caught in review: the first version broke the derived-fields forward compatibility, so a valid value for a future `StepConfig` field would have raised `KeyError`.

Full suite 5,647 passed, `aws_lambda` at 100% coverage, lint and pyright clean.

pydantic_ai_harness/aws_lambda/_bridge.py
```diff
@@ -0,0 +1,567 @@
...
+        # finish unwinding: an agent coroutine's `finally`/`__aexit__` cleanup runs during
+        # cancellation, and letting that overlap the next invocation would touch shared provider
+        # resources after the execution it belonged to was abandoned.
+        _active_bridge.reset(token)
+        if tasks:  # pragma: no branch - only empty when scheduling itself failed
+            task = tasks[0]
+            finished = threading.Event()
+
+            def cancel_and_notify() -> None:
+                if task.done():
```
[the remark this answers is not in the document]
nvasiu: > Track invocation-owned tasks

This part is important to avoid cancelling long lived warm resources (the reason we reuse the loop). But tracking is non-trivial, you'd need to snapshot tasks before the run and cancel the new ones.

If a customer starts detached background work like described above, it can only consume resources in their own Lambda environment (which is dedicated and billed to their function). That work is also not checkpointed, so it falls outside what durable functions support.

With that caveat, and since this is low priority, I think we could just add a line to the documentation saying detached background work is not checkpointed.

nvasiu:
@DouweM  Re: https://github.com/pydantic/pydantic-ai-harness/pull/417#issuecomment-5502000700

> ## 1. The overlap: your call

I think option A is sufficient. We already have a bounded wait with `cancel_timeout`, so the abandoned run's cleanup has time to finish before the next invocation. Adding another wait still doesn't guarantee that the cleanup will finish, and just adds latency and another time-out knob. With the current approach, the worst overlap (a sync tool still running on an executor thread) is already handled, so the remaining risk under option A is narrow.

> ## 2. The executor test: assertion changed

That test matches what I had in mind, sounds good to me.

nvasiu (approved):

PR #422, opened 2026-07-22
feat(airflow): native AirflowDurability durable-execution capability

DouweM:
## Summary

Adds `AirflowDurability`, a native durable-execution capability on the capability-based API that is the **capability-shaped reference version** of Apache Airflow's [`apache-airflow-providers-common-ai`](https://github.com/apache/airflow/tree/main/providers/common/ai) durable integration. The goal is to show the Airflow team the capability shape so they can decide whether to adopt it.

Airflow's real integration is a **memoization cache**, not a journal-with-suspend: its durable unit is the whole Airflow task, and recovery is a task *retry* that re-runs the agent from the top and replays each model/tool call from a cache keyed by a positional counter plus a content fingerprint. This capability reproduces that idea as a `BaseDurabilityCapability` subclass.

### How it works

- **Positional step keys.** A per-run counter (isolated per run via `wrap_run`) numbers each operation in the order it is reached (`model_step_0`, `tool_step_1`, `model_step_2`, ...). The counter advances even on a store hit, so a retry that replays earlier steps still reaches later ones at the same positions. Tool calls run sequentially inside a memoized run so their positions are deterministic.
- **Fingerprint-verified replay.** Each entry stores a fingerprint of its request (the *prepared* model request -- `prepare_request` is called first, following the provider's comment on why; message history with per-attempt volatile fields stripped per Prefect's `_cache_policies`; tool name + args + `tool_call_id`). On a positional hit the current fingerprint is compared to the stored one; a mismatch is a miss and the step re-runs live. `tool_call_id` is folded into the tool fingerprint, so a fresh model response cascades to invalidate the tool steps recorded under the old conversation.
- **Pluggable storage.** The store is a small `DurableStorage` Protocol. An `InMemoryDurableStorage` and a `JSONFileDurableStorage` ship for offline use; a backend writing to Airflow's ObjectStorage or task state store would live in the provider and satisfy the same Protocol. Zero required deps beyond pydantic-ai, no `apache-airflow` install needed.

## What it improves over the real Airflow provider

- **Structured tool results round-trip.** `ToolReturn` and `BinaryContent` are serialized through Pydantic (via `wrap_tool_call_result`/`unwrap_recorded_tool_call_result`), so they are memoized. The provider's raw `json.dumps` silently skips them (there is a test proving ours caches them).
- **Control-flow signals round-trip.** `ModelRetry`, `ApprovalRequired`, and `CallDeferred` cross the store as values and re-raise on replay.
- **MCP tool listings/instructions and dynamic-toolset resolution are memoized.** The provider re-runs those live on every attempt (#5884).
- **Graceful degradation preserved.** A failed/absent fingerprint falls back to unverified positional replay; a non-serializable result is skipped with a warning rather than crashing.

## Honest limitations (reference-quality, not a drop-in engine)

- **No suspend/resume.** Recovery is a full re-run from the top, not a resume from a checkpoint. Deferred tools (`ApprovalRequired`/`CallDeferred`) genuinely cannot suspend-and-resume in this model: they end the run with a `DeferredToolRequests` output, memoized like any other result, and resolving one is a fresh run the caller drives (there is a test pinning this current behavior).
- **`in_durable_context` is a construction-time flag** (`durable=`, defaulting `True`), not a runtime check. Unlike an out-of-process engine there is no Airflow-task context to detect, so attaching the capability is what makes the run memoized -- mirroring the provider's `AgentOperator(durable=...)`.
- **Positional keying is order-sensitive**, and a step is stored *after* it runs (keep tool side effects idempotent).
- **`ctx.enqueue()` is guarded** inside memoized tool calls **and both event-stream-handler delivery paths** (model events via `capture_event_stream`, agent events via `_dispatch_event_stream_event`), matching pydantic-ai #6671. Targeting slim 2.15.0, this uses the local `EnqueueGuard` on the passed context; it collapses onto the base `_durable_run_context` / `_durable_run_context_scope` helpers in the 2.16.0 sweep (noted in a comment).

## Tests

`tests/airflow/` (53 tests, no network/Docker/Airflow). In-memory store JSON-round-trips every value so a non-serializable value that would be skipped in prod is observable; a task retry is simulated by re-running the agent against the populated store. Covers: model+tool memoization, retry replay with 0 re-execution, crash-mid-run retry, fingerprint mismatch (miss + warning), `tool_call_id` cascade invalidation, `ToolReturn`/`BinaryContent` round-trip, MCP `get_tools`/`get_instructions`/`call_tool`, dynamic toolsets, control-flow signals, transparency (`durable=False` and capability-absent), runtime-toolset rejection (parametrized), enqueue guard on all three delivery paths, non-serializable graceful skip, event_stream_handler, cancel-suspended-response, both storage backends, and the fingerprint helpers. 100% branch coverage of source **and** tests.

## Dependency change

Bumps `pydantic-ai-slim` from `>=2.14.1` to `>=2.15.0` (required for the shared durable-execution primitives: `DurableDynamicToolset`, `get_dynamic_tools`, `wrap_tool_call_result`, etc.), and the dev extra to `pydantic-ai-slim[mcp,spec]>=2.15.0` for the MCP tests. `uv.lock` updated via `uv add`. Flagging per the checklist -- this mirrors the dependency floor the sibling durability PRs need.

## Linked issue

No linked issue: opened at DouweM's request to share the capability shape with the Airflow team.

## Siblings

Companion to the other durable-execution capability PRs: #417 (AWS Lambda) and #399 (Absurd).

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,6 +187,7 @@ We studied leading coding agents, agent frameworks, and Claw-style assistants to
 | **Reliability** | **Stuck loop detection** | Detect and break out of repetitive agent loops | :construction: [PR&nbsp;#186](https://github.com/pydantic/pydantic-ai-harness/pull/186) | |
 | | **Tool error recovery** | Retry failed tool calls with backoff and budget | :construction: [PR&nbsp;#171](https://github.com/pydantic/pydantic-ai-harness/pull/171) | |
 | | **Tool orphan repair** | Fix orphaned tool calls in conversation history | :construction: [PR&nbsp;#184](https://github.com/pydantic/pydantic-ai-harness/pull/184) | |
+| | **Airflow durability** | Memoize model and tool I/O so an Apache Airflow task retry replays completed work instead of repeating it (reference capability shape) | :white_check_mark: [Docs](pydantic_ai_harness/airflow/) | |
 | **Reasoning** | **Adaptive reasoning** | Adjust thinking effort based on task complexity | :construction: [PR&nbsp;#174](https://github.com/pydantic/pydantic-ai-harness/pull/174) | |
 | | **Current time** | Inject current date/time into system prompt | :construction: [PR&nbsp;#170](https://github.com/pydantic/pydantic-ai-harness/pull/170) | |
 
--- a/docs/airflow.md
+++ b/docs/airflow.md
@@ -1,0 +1,102 @@
+---
+title: Airflow Durability
+description: Memoize a Pydantic AI agent's model and tool I/O into a pluggable store so an Apache Airflow task retry replays completed work instead of repeating it.
+---
+
+# Airflow Durability
+
+`AirflowDurability` makes an agent recover across an Apache Airflow task retry by memoizing its I/O.
+Every model request, function tool call, MCP call, and dynamic-toolset resolution is stored under a
+positional step key plus a fingerprint of the request that produced it. If the agent run fails, the
+Airflow task retry re-runs it from the top: each completed operation is served from the store without
+re-executing, so finished work is not repeated and tokens are not re-spent, while the operation that
+failed and everything after it runs live.
+
+It is the capability-shaped reference version of Airflow's `common.ai` durable provider, which
+memoizes the same operations but is attached through an operator rather than a user-attached
+capability.
+
+[Source](https://github.com/pydantic/pydantic-ai-harness/tree/main/pydantic_ai_harness/airflow/)
+
+> The API may change between releases. Where practical, breaking changes ship with a deprecation warning.
+
+## How it works
+
+The durable unit is the whole agent run, which maps to one Airflow task. There is no
+suspend-and-resume: recovery is a full re-run against a store that outlived the failed attempt.
+
+1. **Positional step keys.** A per-run counter numbers each operation in the order it is reached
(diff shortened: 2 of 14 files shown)


DouweM:
Closing this one. On review it isn't actually an Airflow integration: it reimplements the *caching algorithm* that `apache-airflow-providers-common-ai` uses (positional counter + content fingerprint, replay-and-verify) as a `BaseDurabilityCapability` subclass over a generic storage protocol, but it touches nothing Airflow — no `AgentOperator`, no task retries, no object-storage cache, no HITL.

That's structural, not an oversight. Airflow's durable unit is an **Airflow task** and recovery is a **task retry**; the integration lives in the operator/scheduler layer, which a user-attached capability can't reach, and Airflow already ships it in their own provider package (deliberately outside `pydantic_ai.durable_exec`). So a harness capability would be the wrong layer.

The valuable output from the exercise was the fit analysis, which we've kept as design input for the runtime-capability abstraction (pydantic/pydantic-ai#5477): a capability isn't necessarily the whole integration, and "is there a runtime durable-context signal?" is a real axis (Airflow has none observable from the agent process).

We'd be glad to collaborate with the Airflow `common.ai` folks on the right way to give Airflow a mention in the harness repo — likely a thin operator/host that runs the agent, rather than a capability.

Issue #441, opened 2026-07-23
`StepStore`: bound snapshot growth with opt-in `max_snapshots_per_run`


dsfaccini:
This issue was posted by Claude Code using claude-opus-4-8 on behalf of David.

**Problem.** `StepPersistence` writes a new full-history snapshot at every settled tool-call boundary, keyed by an incrementing `seq`, and never prunes. Within one `Agent.run` the snapshot count equals the number of tool-call steps, and because snapshot *k* holds the full history up to step *k*, total stored bytes are O(steps^2). Long single runs (e.g. a 200-step coding run) pay a large storage cost. `README.md` documents this as a deliberate boundary today ("It does not clean up old snapshots/events. Retention is the caller's responsibility"), so the current behavior is intended -- this issue proposes an opt-in bound, not a silent change.

Only `latest_snapshot` reads snapshots (`continue_run`/`fork_run` both go through it). Serving both read modes needs at most two snapshots per run: the newest overall (for `include_interrupted=True`) and the newest `complete`. Every other snapshot is dead weight for current and planned consumers (the append-only `events` log is the audit substrate; arbitrary-point forking is tracked in #321).

**Proposed change.** Add an opt-in `max_snapshots_per_run: int | None = None` constructor argument to `InMemoryStepStore`, `FileStepStore`, `SqliteStepStore` (and `MongoStepStore` when #382 lands). Pruning runs inside each store's `save_snapshot`.

- **Default `None` = today's unbounded behavior, byte-for-byte.** No `StepStore` Protocol change (adding a required method would break third-party implementers the README invites), no default-behavior change, existing DBs/dirs untouched.
- **Retain-invariant** (correctness-safe even at `N=1`): always keep `{top-N by seq} + {max seq} + {max seq where state='complete'}`. This survives the case where the last N snapshots are all `interrupted` and the last resumable `complete` sits below the window.
- A shared `_retained_seqs(...)` helper holds the invariant in one place; each backend prunes in its own idiom (list rebuild / `unlink` / indexed `DELETE`).

**Non-goals (deliberate, flagged so they aren't read as oversights).**
- No auto-delete-on-save default -- silent deletion of a documented-durable artifact is off by default; one arg enables it.
- No base-`StepStore` Protocol method -- keeps third-party stores working.
- No TTL / age-based expiry -- wrong axis for per-snapshot resume correctness; belongs at whole-run granularity and is Mongo/Redis-native. Separate follow-up.
[...]

Issue #254, opened 2026-05-25
Capability: `MediaExternalizer` — externalize `BinaryContent` to URL / `UploadedFile` before the model sees the message


dsfaccini:
## Goal

Reduce wire payload and token cost on multimodal runs by rewriting `BinaryContent` parts in `ModelRequest` messages **before** the model sees them. The capability sits in front of `StepPersistence` (or stands alone) and chooses, per part, between three modalities exposed by pydantic_ai:

1. **Inline bytes** (`BinaryContent`) — universal, expensive on the wire
2. **URL** (`ImageUrl` / `AudioUrl` / `VideoUrl` / `DocumentUrl`) — cheap when supported; if the model doesn't accept the URL natively, pyai providers silently download the bytes server-side, so emitting a URL is always *safe* (you lose wire savings, not correctness)
3. **Provider upload reference** (`UploadedFile`) — cheap when supported; not portable across providers (file IDs are per-provider); requires a one-time upload step. v2-main `messages.py:726` confirms it's still present and supported by Anthropic, OpenAIChat, OpenAIResponses, Bedrock (S3 URIs), Google (Gemini Files / GCS), Xai.

This is the spiritual successor to [pyai #3807](https://github.com/pydantic/pydantic-ai/pull/3807) (closed: "this is now achievable with capabilities"), built against the modern capability surface.

## Bottom layer is already shipped (PR #251)

`MediaStore` protocol (in `pydantic_ai_harness.media`):

- `put(data, *, context)` → `media+sha256://<hex>` URI
- `get(uri, *, context)` → bytes
- `exists(uri, *, context)` → bool
- `public_url(uri, *, context)` → `str | None` (user-supplied callable; sync or async, takes `(uri, MediaContext)`)

Every method takes a `MediaContext` — frozen dataclass with `media_type`, `filename`, `metadata`. **New context fields are added non-breakingly**, so the capability spec below stays stable across iterations.

Concrete stores: `DiskMediaStore`, `SqliteMediaStore`, `S3MediaStore`. Each supports a `key_strategy=` callable for backend layout control. URI scheme is content-addressed so dedup is automatic across stores. See PR #251 for the protocol + `StepPersistence` integration.

## Capability shape

```python
from pydantic_ai import Agent
from pydantic_ai_harness import MediaExternalizer, StepPersistence
from pydantic_ai_harness.media import S3MediaStore, make_static_public_url

store = S3MediaStore(
[...]

PR #446, opened 2026-07-24
Add first-party MongoDB backend + externalize large text parts

### LINKED ISSUE #382: Step Persistence and Memory add mongodb
### Capability Name

MongoMemoryStore, MongoStepStore

### Description

We hope to increase MongoDB storage. Our company team uses MongoDB for all projects and it is not possible to migrate data to PGDB

### Use Case

[step-persistence](https://pydantic.dev/docs/ai/harness/step-persistence/)

[memory](https://pydantic.dev/docs/ai/harness/memory/)

_No response_

### Prior Art / References
MongoStepStore and MongoMediaStore
```python
"""MongoDB-backed StepStore implementation.

Collections: ``runs``, ``events``, ``snapshots``, ``tool_effects``, ``counters``.
Concurrency: atomic ``$inc`` for sequence counters + unique-index upsert for
tool effects match the access pattern of one capability per ``Agent.run``.
"""

from __future__ import annotations

import json
import re
from collections.abc import Mapping
from datetime import datetime
from typing import TYPE_CHECKING

from bson.binary import Binary
from pymongo import AsyncMongoClient
from pymongo import ReturnDocument

if TYPE_CHECKING:
    from pymongo.asynchronous.database import AsyncDatabase
from pydantic_ai.messages import ModelMessage, ModelMessagesTypeAdapter

from pydantic_ai_harness.media import (
    MediaContext,
    MediaStore,
    externalize_media,
    restore_media,
)
from pydantic_ai_harness.media._store import (
    PublicUrlResolver,
    _EMPTY_CONTEXT,
    _coerce_metadata_mapping,
    _resolve_public_url,
    media_uri_for,
    parse_media_uri,
)
from pydantic_ai_harness.step_persistence._store import (
    _AutoMedia,
    _DEFAULT_MEDIA_THRESHOLD_BYTES,
    _EVENT_KIND_TABLE,
    _TOOL_STATUS_TABLE,
    _opt_str,
    _snapshot_state,
)
from pydantic_ai_harness.step_persistence._types import (
    ContinuableSnapshot,
    RunRecord,
    StepEvent,
    ToolEffectRecord,
)


def _str_str_dict(value: object) -> dict[str, str]:
    if value is None:
        return {}
    if isinstance(value, dict):
        return {str(k): str(v) for k, v in value.items()}
    if isinstance(value, str):
        return json.loads(value)
    raise ValueError(f'expected dict|str|None, got {type(value).__name__}')


def _event_to_doc(event: StepEvent) -> dict:
    return {
        'run_id': event.run_id,
        'kind': event.kind,
        'step_index': event.step_index,
        'timestamp': event.timestamp.isoformat(),
        'conversation_id': event.conversation_id,
        'parent_run_id': event.parent_run_id,
        'agent_name': event.agent_name,
        'tool_call_id': event.tool_call_id,
        'tool_name': event.tool_name,
        'error': event.error,
        'metadata': dict(event.metadata),
    }


def _event_from_doc(doc: dict) -> StepEvent:
    kind_raw = doc['kind']
    kind = _EVENT_KIND_TABLE.get(kind_raw)
    if kind is None:
        raise ValueError(f'unknown event kind: {kind_raw!r}')
    return StepEvent(
        run_id=doc['run_id'],
        kind=kind,
        step_index=doc['step_index'],
        timestamp=datetime.fromisoformat(doc['timestamp']),
        conversation_id=_opt_str(doc.get('conversation_id')),
        parent_run_id=_opt_str(doc.get('parent_run_id')),
        agent_name=_opt_str(doc.get('agent_name')),
        tool_call_id=_opt_str(doc.get('tool_call_id')),
        tool_name=_opt_str(doc.get('tool_name')),
        error=_opt_str(doc.get('error')),
        metadata=_str_str_dict(doc.get('metadata')),
    )


def _run_to_doc(record: RunRecord) -> dict:
    return {
        '_id': record.run_id,
        'conversation_id': record.conversation_id,
        'parent_run_id': record.parent_run_id,
        'agent_name': record.agent_name,
        'metadata': dict(record.metadata),
        'started_at': record.started_at.isoformat(),
    }


def _run_from_doc(doc: dict) -> RunRecord:
    return RunRecord(
        run_id=doc['_id'],
        conversation_id=_opt_str(doc.get('conversation_id')),
        parent_run_id=_opt_str(doc.get('parent_run_id')),
        agent_name=_opt_str(doc.get('agent_name')),
        metadata=_str_str_dict(doc.get('metadata')),
        started_at=datetime.fromisoformat(doc['started_at']),
    )


def _tool_effect_to_doc(record: ToolEffectRecord) -> dict:
    return {
        'tool_call_id': record.tool_call_id,
        'tool_name': record.tool_name,
        'run_id': record.run_id,
        'status': record.status,
        'started_at': record.started_at.isoformat(),
        'ended_at': record.ended_at.isoformat() if record.ended_at is not None else None,
        'idempotency_key': record.idempotency_key,
        'effect_summary': record.effect_summary,
    }


def _tool_effect_from_doc(doc: dict) -> ToolEffectRecord:
    status_raw = doc['status']
    status = _TOOL_STATUS_TABLE.get(status_raw)
    if status is None:
        raise ValueError(f'unknown tool effect status: {status_raw!r}')
    ended_at_raw = doc.get('ended_at')
    return ToolEffectRecord(
        tool_call_id=doc['tool_call_id'],
        tool_name=doc['tool_name'],
        run_id=doc['run_id'],
        status=status,
        started_at=datetime.fromisoformat(doc['started_at']),
        ended_at=datetime.fromisoformat(ended_at_raw) if isinstance(ended_at_raw, str) else None,
        idempotency_key=_opt_str(doc.get('idempotency_key')),
        effect_summary=_opt_str(doc.get('effect_summary')),
    )


_COLLECTION_NAME_RE = re.compile(r'^[A-Za-z_][A-Za-z0-9_]*$')


class MongoMediaStore:
    """MongoDB-backed `MediaStore`. One document per blob in a collection keyed by sha256 hex.

    Content-addressed; the same bytes written twice produce the same URI and
    the second write is a no-op (idempotent). The collection uses ``_id`` as
    the hex digest -- any user-chosen key would break dedup.

    Pass ``client=`` (shared ``AsyncMongoClient``) or ``db_url=`` +
    ``database=`` (creates and owns its own client).

    Document layout:

    .. code-block:: json

        {
          "_id": "<sha256-hex>",
          "media_type": "image/png",
          "bytes": "<Binary>",
          "size_bytes": 12345,
          "metadata": "{}"
        }

    ``$setOnInsert`` makes writes idempotent. ``metadata`` is stored as
    canonical JSON of ``context.metadata`` and read back via
    ``get_metadata(uri)``.
    """

    def __init__(
        self,
        *,
        client: AsyncMongoClient | None = None,
        db_url: str | None = None,
        database: str | None = None,
        collection: str = 'media',
        public_url: PublicUrlResolver | None = None,
    ) -> None:
        if client is not None:
            self._client = client
            self._own_client = False
        elif db_url is not None:
            self._client = AsyncMongoClient(db_url)
            self._own_client = True
        else:
            raise ValueError('provide either `client=` or `db_url=`')

        if database is None:
            raise ValueError('`database=` is required')

        if not _COLLECTION_NAME_RE.fullmatch(collection):
            raise ValueError(f'invalid collection name: {collection!r}')

        self._db: AsyncDatabase = self._client[database]
        self._collection_name = collection
        self._public_url_resolver = public_url
        self._indexes_ready = False

    async def _ensure_indexes(self) -> None:
        if self._indexes_ready:
            return
        await self._db[self._collection_name].create_index(
            'media_type', sparse=True
        )
        self._indexes_ready = True

    # ── MediaStore protocol ────────────────────────────────────────────

    async def put(self, data: bytes, *, context: MediaContext = _EMPTY_CONTEXT) -> str:
        await self._ensure_indexes()
        uri = media_uri_for(data)
        digest = parse_media_uri(uri)
        metadata_json = json.dumps(dict(context.metadata))
        await self._db[self._collection_name].update_one(
            {'_id': digest},
            {
                '$setOnInsert': {
                    'media_type': context.media_type,
                    'bytes': Binary(data),
                    'size_bytes': len(data),
                    'metadata': metadata_json,
                }
            },
            upsert=True,
        )
        return uri

    async def get(self, uri: str, *, context: MediaContext = _EMPTY_CONTEXT) -> bytes:
        digest = parse_media_uri(uri)
        doc = await self._db[self._collection_name].find_one(
            {'_id': digest}, {'bytes': 1}
        )
        if doc is None:
            raise FileNotFoundError(f'media not found: {uri}')
        return bytes(doc['bytes'])

    async def exists(self, uri: str, *, context: MediaContext = _EMPTY_CONTEXT) -> bool:
        digest = parse_media_uri(uri)
        doc = await self._db[self._collection_name].find_one(
            {'_id': digest}, {'_id': 1}
        )
        return doc is not None

    async def public_url(
        self, uri: str, *, context: MediaContext = _EMPTY_CONTEXT
    ) -> str | None:
        return await _resolve_public_url(self._public_url_resolver, uri, context)

    async def get_metadata(
        self, uri: str, *, context: MediaContext = _EMPTY_CONTEXT
    ) -> Mapping[str, str]:
        digest = parse_media_uri(uri)
        doc = await self._db[self._collection_name].find_one(
            {'_id': digest}, {'metadata': 1}
        )
        if doc is None:
            raise FileNotFoundError(f'media not found: {uri}')
        return _coerce_metadata_mapping(json.loads(doc.get('metadata', '{}')))


class MongoStepStore:
    """MongoDB-backed step-persistence store.

    Pass ``client=`` (a shared ``AsyncMongoClient``) or ``db_url=`` +
    ``database=`` (the store creates and owns its own client). When
    ``client=`` is used the caller is responsible for its lifecycle.

    ``media_store`` defaults to ``None`` (binary payloads stay inline
    inside snapshot documents). Pass a ``MediaStore`` instance (e.g.
    ``DiskMediaStore``, ``S3MediaStore``) to externalize large
    ``BinaryContent`` parts.

    Collections created on first use:

    - ``runs``        — ``_id = run_id``
    - ``events``      — ``(run_id, seq)`` compound index
    - ``snapshots``   — ``(run_id, seq)`` + ``(run_id, state, seq)`` indexes
    - ``tool_effects`` — ``(run_id, tool_call_id)`` unique upsert key
    - ``counters``    — ``_id`` per entity for atomic ``seq`` allocation
    """

    def __init__(
        self,
        *,
        client: AsyncMongoClient | None = None,
        db_url: str | None = None,
        database: str | None = None,
        media_store: MediaStore | None | _AutoMedia = 'auto',
        media_threshold_bytes: int = _DEFAULT_MEDIA_THRESHOLD_BYTES,
    ) -> None:
        if client is not None:
            self._client = client
            self._own_client = False
        elif db_url is not None:
            self._client = AsyncMongoClient(db_url)
            self._own_client = True
        else:
            raise ValueError('provide either `client=` or `db_url=`')

        if database is None:
            raise ValueError('`database=` is required')

        self._db: AsyncDatabase = self._client[database]

        resolved: MediaStore | None
        if media_store == 'auto':
            resolved = MongoMediaStore(client=self._client, database=database)
        else:
            resolved = media_store
        self._media_store: MediaStore | None = resolved
        self._media_threshold_bytes = media_threshold_bytes
        self._indexes_ready = False

    async def _ensure_indexes(self) -> None:
        if self._indexes_ready:
            return

        # runs — filtered list queries
        await self._db.runs.create_index('conversation_id', sparse=True)
        await self._db.runs.create_index('parent_run_id', sparse=True)
        await self._db.runs.create_index('started_at')

        # events — ordered by seq per run
        await self._db.events.create_index([('run_id', 1), ('seq', 1)])

        # snapshots — latest-per-run, optionally filtered by state
        await self._db.snapshots.create_index([('run_id', 1), ('seq', -1)])
        await self._db.snapshots.create_index([('run_id', 1), ('state', 1), ('seq', -1)])

        # tool_effects — unique per (run_id, tool_call_id) for upsert;
        # also an index for list_unresolved queries
        await self._db.tool_effects.create_index(
            [('run_id', 1), ('tool_call_id', 1)], unique=True
        )
        await self._db.tool_effects.create_index([('run_id', 1), ('status', 1)])

        self._indexes_ready = True

    async def _next_seq(self, name: str) -> int:
        doc = await self._db.counters.find_one_and_update(
            {'_id': name},
            {'$inc': {'seq': 1}},
            upsert=True,
            return_document=ReturnDocument.AFTER,
        )
        assert doc is not None
        return doc['seq']

    # ── Run ────────────────────────────────────────────────────────────

    async def register_run(self, record: RunRecord) -> None:
        await self._ensure_indexes()
        await self._db.runs.insert_one(_run_to_doc(record))

    async def get_run(self, *, run_id: str) -> RunRecord | None:
        doc = await self._db.runs.find_one({'_id': run_id})
        if doc is None:
            return None
        return _run_from_doc(doc)

    async def list_runs(
        self,
        *,
        parent_run_id: str | None = None,
        conversation_id: str | None = None,
    ) -> list[RunRecord]:
        query: dict = {}
        if parent_run_id is not None:
            query['parent_run_id'] = parent_run_id
        if conversation_id is not None:
            query['conversation_id'] = conversation_id

        cursor = self._db.runs.find(query).sort('started_at', 1)
        return [_run_from_doc(doc) async for doc in cursor]

    # ── Events ─────────────────────────────────────────────────────────

    async def append_event(self, event: StepEvent) -> None:
        await self._ensure_indexes()
        doc = _event_to_doc(event)
        doc['seq'] = await self._next_seq('events')
        await self._db.events.insert_one(doc)

    async def list_events(self, *, run_id: str) -> list[StepEvent]:
        cursor = self._db.events.find({'run_id': run_id}).sort('seq', 1)
        return [_event_from_doc(doc) async for doc in cursor]

    # ── Snapshots ──────────────────────────────────────────────────────

    async def save_snapshot(self, snapshot: ContinuableSnapshot) -> None:
        await self._ensure_indexes()
        messages_json: object = json.loads(
            ModelMessagesTypeAdapter.dump_json(snapshot.messages).decode('utf-8')
        )
        if self._media_store is not None:
            messages_json = await externalize_media(
                messages_json,
                media_store=self._media_store,
                threshold_bytes=self._media_threshold_bytes,
            )

        seq = await self._next_seq('snapshots')
        doc = {
            'seq': seq,
            'run_id': snapshot.run_id,
            'step_index': snapshot.step_index,
            'conversation_id': snapshot.conversation_id,
            'parent_run_id': snapshot.parent_run_id,
            'agent_name': snapshot.agent_name,
            'timestamp': snapshot.timestamp.isoformat(),
            'state': snapshot.state,
            'messages': json.dumps(messages_json),
        }
        await self._db.snapshots.insert_one(doc)

    async def latest_snapshot(
        self, *, run_id: str, include_interrupted: bool = False
    ) -> ContinuableSnapshot | None:
        query: dict = {'run_id': run_id}
        if not include_interrupted:
            query['state'] = 'complete'

        doc = await self._db.snapshots.find_one(query, sort=[('seq', -1)])
        if doc is None:
            return None

        messages_json: object = json.loads(doc['messages'])
        if self._media_store is not None:
            messages_json = await restore_media(
                messages_json, media_store=self._media_store
            )

        messages: list[ModelMessage] = ModelMessagesTypeAdapter.validate_python(
            messages_json
        )
        return ContinuableSnapshot(
            run_id=run_id,
            step_index=doc['step_index'],
            messages=messages,
            conversation_id=_opt_str(doc.get('conversation_id')),
            parent_run_id=_opt_str(doc.get('parent_run_id')),
            agent_name=_opt_str(doc.get('agent_name')),
            timestamp=datetime.fromisoformat(doc['timestamp']),
            state=_snapshot_state(doc.get('state')),
        )

    # ── Tool Effects ───────────────────────────────────────────────────

    async def record_tool_effect(self, record: ToolEffectRecord) -> None:
        await self._ensure_indexes()
        doc = _tool_effect_to_doc(record)
        await self._db.tool_effects.update_one(
            {'run_id': record.run_id, 'tool_call_id': record.tool_call_id},
            {'$set': doc},
            upsert=True,
        )

    async def get_tool_effect(
        self, *, run_id: str, tool_call_id: str
    ) -> ToolEffectRecord | None:
        doc = await self._db.tool_effects.find_one(
            {'run_id': run_id, 'tool_call_id': tool_call_id}
        )
        if doc is None:
            return None
        return _tool_effect_from_doc(doc)

    async def list_unresolved_tool_effects(
        self, *, run_id: str
    ) -> list[ToolEffectRecord]:
        cursor = self._db.tool_effects.find(
            {'run_id': run_id, 'status': 'started'}
        )
        return [_tool_effect_from_doc(doc) async for doc in cursor]

```

### LINKED ISSUE #440: Externalize large text parts (not just `BinaryContent`) + GridFS for oversized `MediaStore` blobs
This issue was posted by Claude Code using claude-opus-4-8 on behalf of David.

**Problem.** `StepPersistence` snapshots store the full message history in one record. Externalization to the `MediaStore` fires only on `BinaryContent` >= `media_threshold_bytes` (64 KiB). Large *text* parts -- notably tool-return payloads -- stay inline. On a `MongoStepStore` (#382) this can push a snapshot document past MongoDB's hard 16MB BSON limit, failing the write. SQLite/File backends don't fail but carry the same bloat.

**Proposed change.**

1. Extend the externalization walk (`pydantic_ai_harness/media/_walker.py`) to also externalize text parts above a threshold, replacing them with the same content-addressed markers used for binary. Applies to all backends.
2. Add a GridFS path to `MongoMediaStore` for blobs that exceed the single-document limit, so the media layer -- not the snapshot -- owns >16MB storage.

**Scope / non-goals.** This is not snapshot retention/GC (unbounded snapshot *count* is a separate concern). It does not change the `StepStore` protocol.

**Origin.** Surfaced by @InfernalAzazel's design pushback on #382.

---

**Update (implementation in #446).** The MongoDB media layer uses content-addressed manual chunking (a `media` doc + `media_chunks` docs each well under 16MB), not GridFS. Reasons: GridFS keys files by ObjectId and does no sha256 dedup (we'd have to layer it on anyway), and native async GridFS has no in-memory test fake, so its blob path could not meet the repo's 100% branch-coverage bar without shipping the deprecated `motor` driver. Manual chunking meets the same goal (media layer owns >16MB, the snapshot document never hits the BSON cap) while staying dedup-correct and fully testable.

dsfaccini:
This pull request was posted by Claude Code using claude-opus-4-8 on behalf of David.
David: no line review (orchestrated).

Closes #382
Closes #440

## Summary

Three coupled changes, all needed for MongoDB shops to use step-persistence without a Postgres migration and without blowing MongoDB's 16MB document cap.

### 1. Externalize large text parts, not just `BinaryContent` (closes #440)

`media/_walker.py` previously externalized only `BinaryContent` >= threshold. A part whose string `content` reaches the threshold (e.g. a big tool-return string) is now offloaded to the content-addressed `MediaStore` too and re-inlined on read, symmetric with the binary path. This benefits **all** backends (File/Sqlite/Mongo), not just Mongo. A `__harness_external_text__` marker discriminates the two on restore (text decodes UTF-8 into `content`; binary decodes base64 into `data`). The existing `media_threshold_bytes` governs both -- no second knob (deliberate: the 16MB driver is aggregate snapshot size, and a text threshold measured in UTF-8 bytes is the natural companion to the binary byte threshold). Below-threshold parts pass through byte-identical, so no behavior change.

### 2. First-party MongoDB backend (closes #382)

`MongoStepStore` and `MongoMediaStore` mirror the Sqlite/S3 backend lifecycle: constructor takes a shared `AsyncMongoClient` (`client=`) **or** a connection string it owns (`db_url=`, released via `await store.aclose()`); `database=` always required. `MongoStepStore` uses collections `runs` / `events` / `snapshots` / `tool_effects` / `counters` (atomic `$inc` for monotonic `seq`), with `runs._id = run_id` enforcing the single-shot `run_id` contract. Both are lazily exported from their package `__init__` so `pymongo` stays optional.

`pymongo` is a new optional dependency behind the `mongodb` extra (`pip install pydantic-ai-harness[mongodb]`), added via `uv add pymongo --optional mongodb`, following the single-package `exa`/`modal` extra shape.

### 3. Manual chunking instead of GridFS (16MB safety)

`MongoMediaStore` stores each blob as sha256-addressed chunks across a `media` document (`_id = <digest>`) and a sibling `media_chunks` collection (`chunk_size_bytes` default 8MB), so no single document nears the 16MB BSON cap at any blob size. Chunking is unconditional (a small blob is one chunk -- branchless write path). A lazy compound index on `media_chunks (files_id, n)` keeps reassembly reads off a collection scan, and `get` verifies the reassembled length against the stored `size_bytes`, converting a lost/truncated chunk from a silent short read into a loud error.

**Why manual chunking, not the requested GridFS driver:** the orchestrator approved this deviation from #440's literal "GridFS" wording after I surfaced it. Three reasons: (a) content-addressed dedup -- the digest is the files `_id`, so identical bytes are a no-op second write, whereas GridFS keys files by `ObjectId` and does no dedup, forcing a sha256 layer on top anyway; (b) the native async GridFS driver (`gridfs.asynchronous.AsyncGridFSBucket`) rejects every in-memory fake, so it could not meet the repo's 100%-coverage bar without a live mongod, and its only fakeable path is the deprecated `motor` driver; (c) plain async collection ops are fully fakeable. #440 will be synced to reflect this.

## Tests + coverage

100% branch coverage on all new code, full suite green, no live services. Tests run against `mongomock-motor` (dev group) -- the async collection surface the store uses is fakeable in-memory.

Two type boundaries are marked, both narrow and deliberate:

- **Test shim** (`_mock_client`, one site per test module): the fake is not a `pymongo.AsyncMongoClient`, so the constructed client carries one `# pyright: ignore`.
- **Source privates** (~14 `# pyright: ignore[reportPrivateUsage]` across the two `_mongo.py` files): the Mongo stores reuse the serialization/record converters and sentinels already living as `_`-prefixed helpers in `step_persistence/_store.py` and `media/_store.py`, rather than duplicating that tested logic. Follow-up: promoting those shared serialization helpers to a non-private internal surface is deferred (net-neutral -- it is a naming/visibility refactor of code these backends already share, with no behavior change).

## Deliberate scope boundaries

- **No `MongoMemoryStore`.** #382 also mentions a Mongo-backed Memory store; deferred to a follow-up to keep this PR reviewable. Nothing here blocks it -- it would reuse the same client/extra conventions.
- **Serialization-helper promotion deferred** (see Tests + coverage above), net-neutral.
- **#447 (media restore sentinel-key collision)** is a pre-existing sharp edge in the walker's marker representation. This PR widens the same marker scheme to text, so it touches the surrounding code, but it does **not introduce** the collision -- the sentinel-key design predates this PR. Tracked separately in #447; left out of scope here so the fix can be reasoned about on its own.

## Related

- #254 (`MediaExternalizer`) -- cross-referenced where text externalization is documented; it will reuse these stores.
- #441 (bound snapshot growth) -- the snapshot-*count* concern #440 explicitly scopes out.
- #447 (restore sentinel-key collision) -- pre-existing, not introduced here (see above).
- #448 (orphaned media blobs on snapshot insert failure) -- deferred; if a snapshot `insert_one` fails after its media blobs were written, those blobs are not rolled back. Out of scope here (media GC is its own concern); tracked in #448.

## Docs

`media/README.md` + `docs/media.md` and `step_persistence/README.md` + `docs/step-persistence.md` updated in step for the new stores, text externalization, and the `mongodb` extra; top-level README extras list updated.


dsfaccini:
This comment was posted by Claude Code using claude-opus-4-8 on behalf of David.

Triaged all Macroscope findings (David: orchestrated, no line review). Commits `9680433`, `c409c52`, `77cf661`:

- `media/_walker.py` `_is_text_part` too loose -> **fixed** `9680433`: now requires a string `part_kind` AND string `content`, so nested tool-return data with a `content`/`uri` key is no longer mistaken for a part.
- `media/_walker.py` no recursion into sibling fields -> **fixed** `9680433`: markers are built from recursively-processed preserved fields (and restored recursively), so nested large payloads externalize and round-trip.
- `step_persistence` README/docs/docstring false 16MB "by default" guarantee -> **fixed** `c409c52`: reworded to per-value offload at `media_threshold_bytes`, not an aggregate cap.
- `media/README.md` 16MB claim vs inline manifest metadata -> **fixed** `77cf661`: scoped to the blob payload, notes metadata is held inline.
- `MongoStepStore.list_runs` lexicographic `started_at` sort -> **fixed** `c409c52`: sorts by parsed datetime instant, matching the base stores.
- `chunk_size_bytes` unbounded -> **fixed** `77cf661`: rejects values above a 16MiB-minus-headroom ceiling.
- chunk-layout "silent corruption" (media/_mongo.py) -> **dismissed with evidence**: the shipped `size_bytes` reassembly guard makes silently-wrong bytes impossible for content-addressed chunks (all chunks under a digest are slices of the same content; a mismatched-layout retry over/under-counts length and raises). Proven by `test_stale_chunks_from_crashed_write_never_silently_corrupt`. Chunk identity intentionally unchanged.
- orphaned media blobs on snapshot insert failure -> **deferred to #448** (same class as the documented no-media-GC non-goal).

100% branch coverage retained, no new `# pragma`. Resolving these threads.

docs/media.md
```diff
@@ -24,7 +24,7 @@ A conversation that carries images, audio, or other `BinaryContent` inlines thos
 
 ## Building blocks, not a capability
 
-These are building blocks. There is no class you add to `Agent(capabilities=[...])` yet. [`StepPersistence`](step-persistence.md) already uses them to keep snapshots small when messages carry `BinaryContent`, and a forthcoming `MediaExternalizer` capability will reuse the same stores to rewrite `BinaryContent` into URL parts before the model sees them.
+These are building blocks. There is no class you add to `Agent(capabilities=[...])` yet. [`StepPersistence`](step-persistence.md) already uses them to keep snapshots small when messages carry `BinaryContent` or large text (e.g. a big tool-return string), and a forthcoming `MediaExternalizer` capability ([#254](https://github.com/pydantic/pydantic-ai-harness/issues/254)) will reuse the same stores to rewrite `BinaryContent` into URL parts before the model sees them.
```
dsfaccini, line 27: worth leaving a comment on that issue so future work takes this into consideration without the risk of it not seeing it. worth adding a CE rule to the repo to always leave a comment on an open issue if the issue is referenced in the docs/code, such that it's easily trackable as a TODO

docs/step-persistence.md
```diff
@@ -297,20 +297,22 @@ The helper reads the active `run_id` from the `StepPersistence` `ContextVar` and
...
+`BinaryContent` payloads (images, audio, documents, video) inlined as base64 inside a snapshot would balloon every file or row containing the message; a large text part (e.g. a big tool-return string) does the same and can push a `MongoStepStore` snapshot past MongoDB's 16MB document cap ([#440](https://github.com/pydantic/pydantic-ai-harness/issues/440)). The file, sqlite, and mongo backends externalize any `BinaryContent.data` -- and any part whose string `content` -- at or above **64 KiB** through a configured `MediaStore`, leaving a URI reference in the snapshot. The same `media_threshold_bytes` governs binary and text alike; there is no separate text knob. Round-trip is transparent: `latest_snapshot(...).messages[*]` returns the original `BinaryContent` bytes and text.
```
dsfaccini: same here

dsfaccini:

Claude here: coordinated review across #339, #413, #442, and #446 (David requested the holistic pass; no separate human line review).

Two standalone fixes and two cross-PR parity actions:

1. Large-text externalization can corrupt legitimate structured tool results. The walker recursively visits arbitrary mappings and treats any mapping with string `part_kind` + string `content` as a message part; it then overwrites `uri`, and restore removes that key. Pydantic AI permits arbitrary mappings as `ToolReturnPart.content`, so a real round-trip of `{'part_kind': ..., 'content': <large string>, 'uri': 'source://original'}` loses the original `uri`. Please scope matching to actual message-part positions/type-aware structures, or preserve colliding fields, and add this nested structured-result regression test. Walker: https://github.com/pydantic/pydantic-ai-harness/blob/77cf6619df6e8db85a5ebd5be37a3d9033af42ba/pydantic_ai_harness/media/_walker.py#L57-L100 and https://github.com/pydantic/pydantic-ai-harness/blob/77cf6619df6e8db85a5ebd5be37a3d9033af42ba/pydantic_ai_harness/media/_walker.py#L128-L205; upstream content contract: https://github.com/pydantic/pydantic-ai/blob/v2.14.1/pydantic_ai_slim/pydantic_ai/messages.py#L1174-L1189

2. The slim CI jobs do not actually test an installation without PyMongo. They sync the global `dev` group; this PR adds `mongomock-motor`, which transitively installs Motor/PyMongo, so lazy-import regressions can pass every slim job. Please add an isolated base-install import smoke test or keep the fake dependency out of the slim environment. Sources: https://github.com/pydantic/pydantic-ai-harness/blob/77cf6619df6e8db85a5ebd5be37a3d9033af42ba/.github/workflows/main.yml#L76-L103 and https://github.com/pydantic/pydantic-ai-harness/blob/77cf6619df6e8db85a5ebd5be37a3d9033af42ba/pyproject.toml#L80-L93

3. Once rebased, `MongoStepStore` needs parity with both concurrent store changes: #442's `max_snapshots_per_run` option/pruning invariant, and #413's ordered, complete-by-default, media-restoring `list_snapshots`. Without the latter, `SnapshotHistorySource(MongoStepStore(...))` fails at search time. Please own those Mongo implementations/tests here after #442/#413 settle their shared retention/recovery contract. Sources: https://github.com/pydantic/pydantic-ai-harness/blob/db7d5a248a58eb3d1609ad0b604500abe80101dc/pydantic_ai_harness/step_persistence/_store.py#L68-L97 and https://github.com/pydantic/pydantic-ai-harness/blob/806fbfc8405af104e8159daee8b2093adbe26d40/pydantic_ai_harness/conversation_search/_source.py#L67-L85

4. Docs parity: the unified media and step-persistence docs still say metadata/public-URL behavior applies to “all three stores,” although Mongo supports both. Please update the count/details while resolving the rebase. https://github.com/pydantic/pydantic-ai-harness/blob/77cf6619df6e8db85a5ebd5be37a3d9033af42ba/docs/media.md#L110-L126

Suggested coordination: #442 first; #413 next with explicit bounded-retention semantics; then this already-conflicting PR rebases and implements both new first-party-store seams.

dsfaccini:
This reply was posted by Claude Code using claude-opus-4-8 on behalf of David.

David directed this; Claude assessed each item against the code and implemented the fixes. No line-by-line human review.

Status on the four items (head d9c28adb05ec851e05091d8a9da1cd2f4be3e591):

**1. Done.** The walker stored the externalized-blob reference under `uri`, which overwrote a genuine `uri` on any `ToolReturnPart.content` mapping that collided on `part_kind` + string `content`. Confirmed against the upstream contract: `ToolReturnContent` permits `Mapping[str, ToolReturnContent]`, so this is reachable with real data. Moved the reference to a namespaced `__harness_external_uri__` key, and restore now drops only the marker keys, so a colliding `uri` (or any other field) round-trips intact. Regression test added: `tests/media/test_media.py::TestMediaWalker::test_colliding_uri_field_round_trips` builds `{'part_kind': 'tool-return', 'content': <70k>, 'uri': 'source://original', ...}` and asserts the original `uri` survives externalize then restore.

**2. Done.** Confirmed the slim job pulls PyMongo transitively via `mongomock-motor` in the `dev` group, so a top-level pymongo import would pass every existing job. Added a `base-import` CI job that runs `uv sync --locked --no-default-groups` (no groups, no extras, so pymongo is genuinely absent -- verified via dry-run) and imports the package plus the `media` and `step_persistence` subpackages, guarded by a `find_spec('pymongo') is None` assertion so the smoke test fails loudly if a base dependency ever pulls pymongo back in. Wired into the `check` gate. Chose a dedicated job over restructuring the `dev` group, since `mongomock-motor` feeds four jobs (test, test-floor, test-latest, localstack-integration) and moving it out has a large blast radius.

**3. Owned as follow-up on this PR, not implemented yet**, for the reason you gave: the shared retention/recovery contract is still moving. Confirmed on `main`: `conversation_search`, `SnapshotHistorySource`, `list_snapshots`, and `max_snapshots_per_run` do not exist yet (they land with #413/#442). So no user can compose `SnapshotHistorySource(MongoStepStore(...))` today, and there is no confusing failure to guard against right now, so I added no early error or doc note. Once #442 and #413 settle their contract, this PR rebases and adds both Mongo seams (`max_snapshots_per_run` pruning and the ordered, complete-by-default, media-restoring `list_snapshots`) with tests. No skipped/xfail placeholders added.

**4. Done.** Updated the "all three stores" metadata/public-URL claims to "all four" now that `MongoMediaStore` supports both, across https://github.com/pydantic/pydantic-ai-harness/blob/first-party-mongodb-backend/docs/media.md, https://github.com/pydantic/pydantic-ai-harness/blob/first-party-mongodb-backend/docs/step-persistence.md, and the step_persistence README, and added the MongoMediaStore metadata-persistence bullet to the per-store list.

Conflict with `main` is resolved (merge, not rebase); the PR is MERGEABLE. CI is running.

pydantic_ai_harness/media/_walker.py
```diff
@@ -117,12 +195,22 @@ async def restore_media(node: object, *, media_store: MediaStore) -> object:
...
+    # Inverse of `_maybe_externalize_*`: drop the marker keys, restore the
+    # externalized field, keep every other field the original part carried
+    # (including a genuine `uri`, which the namespaced `_URI_KEY` never shadows).
+    # Preserved fields are recursively restored so a nested marker (from the
+    # externalize-side recursion) is re-inlined too.
+    restored: dict[str, object] = {}
+    for key, value in node.items():
+        if key in (_EXTERNAL_MARKER, _TEXT_MARKER, _URI_KEY):
+            continue
+        restored[key] = await restore_media(value, media_store=media_store)
```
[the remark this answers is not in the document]
dsfaccini: This reply was posted by Claude Code using claude-opus-4-8 on behalf of David.

David directed; Claude assessed and implemented. No line-by-line human review.

Fixed in fee4703fd62bdccd716239e417d9c0ec2dbb98ca. `_restore_external` now falls back to the plain `uri` key as the blob reference, but only when `_URI_KEY` is absent, so a pre-rename marker restores while a new-format marker's genuine `uri` stays data and is never consumed as a reference (the loss the `_URI_KEY` change fixed). The consumed reference key is dropped from the restored dict, and a genuine `uri` on a normal non-marker part never reaches this path. Regression test `tests/media/test_media.py::TestMediaWalker::test_legacy_uri_marker_restores` restores a legacy marker (`_EXTERNAL_MARKER` + `uri`, no `_URI_KEY`) and asserts a real user `uri` on a normal part is untouched.

pydantic_ai_harness/media/_walker.py
```diff
@@ -87,19 +126,58 @@ async def _maybe_externalize_binary(
...
+
+async def _maybe_externalize_text(
+    node: dict[str, object],
+    media_store: MediaStore,
+    threshold_bytes: int,
+) -> dict[str, object] | None:
+    # `_is_text_part` already verified `content` is a string.
+    content_value = node['content']
+    assert isinstance(content_value, str)
+    raw = content_value.encode('utf-8')
```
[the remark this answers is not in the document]
dsfaccini: This reply was posted by Claude Code using claude-opus-4-8 on behalf of David.

David directed; Claude assessed and implemented. No line-by-line human review.

Fixed in fee4703fd62bdccd716239e417d9c0ec2dbb98ca. The text path now encodes with `errors='surrogatepass'`, and `restore_media` decodes with the same handler, so a below-threshold string carrying an unpaired surrogate passes through untouched instead of raising, and an above-threshold one round-trips exactly. Tests added: `test_sub_threshold_surrogate_passes_through` (sub-threshold lone surrogate returns unchanged, no blob written) and `test_above_threshold_surrogate_round_trips`.

docs/media.md
dsfaccini: This reply was posted by Claude Code using claude-opus-5 on behalf of David.

Done, both halves.

**Comment on the referenced issue:** https://github.com/pydantic/pydantic-ai-harness/issues/254#issuecomment-5110739623 -- records that `docs/media.md` now promises `MediaExternalizer` will reuse these stores, and spells out the two constraints that follow (reuse the `MediaStore` protocol so URIs stay interchangeable; the walker now externalizes text too, so decide explicitly whether text is in scope for a pre-model rewrite). Cross-links #447 as a live sharp edge for that work.

**CE rule:** filed as #493 rather than added here -- context-file changes get their own PR, not a ride-along on a feature PR. The issue proposes the rule text, names `CLAUDE.md` / `agent_docs/review-checklist.md` as the home, and flags the one real design question: a mechanical check can enforce the backlink but not the constraint prose, which is the part that carries the value.

docs/step-persistence.md
dsfaccini: This reply was posted by Claude Code using claude-opus-5 on behalf of David.

Done: https://github.com/pydantic/pydantic-ai-harness/issues/440#issuecomment-5110741981

Carries forward the qualifier that only lives on this docs page -- externalization is a per-value offload, not an aggregate cap, and one `media_threshold_bytes` governs binary and text alike -- plus the gaps that survive the close (#447, #448, #441).

CE rule tracked in #493 (separate PR, see the `docs/media.md` thread).

PR #454, opened 2026-07-24
fix(macroscope): scope CLI processes to their runs

strawgate:
Claude here: this follows up on the two post-merge Macroscope findings from #350:

- [Cancelled reviews can leave subprocess descendants running](https://github.com/pydantic/pydantic-ai-harness/pull/350#discussion_r3632772526)
- [Relative command paths are resolved outside the configured working directory](https://github.com/pydantic/pydantic-ai-harness/pull/350#discussion_r3632772534)

## What changed

- Route cancellation and other exceptional exits through the existing process-group termination and reap path, then re-raise the original exception.
- Remove the separate `shutil.which()` preflight so the actual spawn resolves relative executables using the configured `cwd`.
- Preserve the install hint for a missing executable while reporting a missing working directory as a launch error.
- Add regressions for cancellation propagation, process-group cleanup, relative command lookup, and missing working directories.

## Root cause

The timeout path explicitly terminated the process group, but external cancellation fell through to `Process.aclose()`. The command preflight also ran before spawning and therefore resolved path-valued commands against the harness process directory instead of the configured review directory.

## Validation

- `uv run pre-commit run --all-files --verbose`
- `make lint`
- `make typecheck`
- `make test` -- 3,086 passed, 35 skipped
- `make testcov` -- 100% branch coverage

### DIFF
--- a/pydantic_ai_harness/macroscope/_toolset.py
+++ b/pydantic_ai_harness/macroscope/_toolset.py
@@ -3,7 +3,6 @@
 from __future__ import annotations
 
 import os
-import shutil
 import signal
 import subprocess
 from collections.abc import Iterable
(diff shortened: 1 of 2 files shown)


strawgate:
Closing this out; not carrying it forward.

PR #460, opened 2026-07-25
feat(permission_policy): context-aware policy channel for identity/RBAC decisions

### LINKED ISSUE #459: PermissionPolicy: allow decisions to depend on the run context (caller identity / tenant / delegation), not just tool name and args
### Capability Name

PermissionPolicy — context-aware policy channel (`context_policy`)

### Description

The in-flight `PermissionPolicy` capability (#340, `claude/permission-policy`) decides tool access from `(tool_name, command, args)`. It has **no way to decide from the run's caller identity** — role, tenant, or delegation chain in `RunContext.deps`. `Rule.args` is `Callable[[dict], bool]` (no `ctx`); `resolve()` takes no `ctx`; the only context-aware hook, `on_ask`, fires **only for the `ask` verdict** (human-approval resolution), not for automated allow/deny.

Consequence: you cannot express "let role=admin run `deploy`, deny it for role=viewer" without routing every identity decision through the deferred-approval machinery (`ask` → `ApprovalRequired` → `DeferredToolRequests`), which is heavyweight and semantically wrong — an RBAC denial is not "a human declined to approve."

### Use Case

Multi-tenant and delegated agents: when building an agent whose tool access depends on **who** is asking (role/tenant/delegation chain carried in `ctx.deps`), not just which tool or command. E.g. an orchestrator that delegates to sub-agents with narrower identities, where a sub-agent must be denied `deploy` even though the tool and args are identical to an allowed caller's.

### Proposal

Add an optional `context_policy` channel — a callable `(RunContext, tool_name, args) -> Verdict | None` (sync or async) — merged into the existing decision with the **same most-restrictive-wins** rule (`deny > ask > allow`) already used for the command-safety channel. So an identity `deny` can never be green-lit by a broad `allow` rule (it can only *tighten*, never loosen), preserving the capability's safety invariant. It is the missing third input alongside "what tool" (rules) and "what command" (command-safety): **"who is asking"**. ~20 lines + tests, no new dependency, reuses the capability's own `_RANK`/`Verdict`.

- `wrap_tool_execute` (already has `ctx`): evaluate `context_policy`, merge most-restrictive with the existing `_decide` result. `prepare_tools` / `bare_name_verdict` untouched.

### Prior Art / References

- Builds directly on #340 (`claude/permission-policy`). Implemented + verified on a branch off that head; happy to open the PR as a co-PR against that branch (draft) and retarget to `main` when it merges.

Knucklessg1:
## Summary

Standalone `PermissionPolicy` capability against `main`: an allow/ask/deny rule engine over
tool calls, with the shell-command safety mechanics (compound-command splitting, a
conservative-parse gate, wrapper stripping, a read-only safelist, and a flag denylist) plus
an optional **context policy** channel that decides from the `RunContext` -- caller identity,
tenant, or role in `ctx.deps` -- merged into the rule/command-safety decision with the same
most-restrictive-wins rule (`deny > ask > allow`) used elsewhere in the capability.

This was previously a draft co-PR against `claude/permission-policy` (#340, the in-flight
`PermissionPolicy` capability). #340 has been quiet for two weeks, is `DIRTY` against `main`,
and carries unresolved CodeRabbit CRITICAL findings in its shell-safety layer. Rather than
continue depending on it, this PR reimplements the primitives it needed, targets `main`
directly, and no longer requires #340 to land first.

## Linked Issue

Fixes #459 · Refs #340

## What changed vs. depending on #340

- **Shell-command validators now consume the whole argv, not a few fixed positions.**
  `_is_safe_sed` previously validated only `argv[1]`/`argv[2]` and never inspected anything
  appended after them, so `sed -n 2p -e "3w /tmp/out"` (a destructive second script tacked on
  after the recognized-safe prefix) passed unnoticed and was auto-approved -- a CRITICAL
  finding. The same fixed-position/raw-token shape recurred across the module: flag matching
  didn't canonicalize `--flag=value` forms (`rg --pre=/bin/sh`, `find -exec=/bin/sh`,
  `base64 --output=/tmp/x` all bypassed their denylists), `rm`'s destructive-flag check
  enumerated literal combo strings instead of decomposing bundled short options (`-vrf`
  evaded `{-rf, -fr, -rF}`), only `git branch` checked its own flags so `git diff --output=`
  could write an arbitrary file, `sudo`/`doas` were peeled to their inner command even though
  they change *who* runs it, `xargs` was peeled even though its real argv comes from stdin at
  run time and isn't visible in the command string, and a command substitution hiding inside
  a double-quoted string (`echo "$(touch /tmp/pwn)"`) wasn't gated. All of these are fixed by
  replacing position/raw-token checks with whole-argv consumption and a shared flag
  canonicalizer -- see `_command.py`'s module docstring for the design rationale, and
  `_safelist.py` for the one residual gap called out explicitly (ripgrep/`find` short-option
  bundling isn't decomposed, unlike `rm`'s, because their short-option alphabets include
  value-taking flags where blind splitting would misclassify).
- **`wrap_tool_execute`'s approval check is scoped to this policy, not any approval.**
  `ctx.tool_call_approved` reflects whether *a* call with this `tool_call_id` was approved,
  not which mechanism approved it. `_pending_asks` now tracks which tool_call_ids this policy
  instance itself put on hold, so an unrelated approval (a core `requires_approval=True`
  tool, a different capability) can no longer silently satisfy this policy's own outstanding
  `ask`.
- Also fixed: a stray leading space in the escalation note appended to a guarded tool with no
  description.

## Design (context policy, unchanged from the original proposal)

- New optional field `context_policy: ContextPolicy[AgentDepsT] | None` and type alias
  `ContextPolicy = Callable[[RunContext, str, dict], Verdict | None | Awaitable[...]]`.
- Wired in `wrap_tool_execute`: evaluate `context_policy`, and if it returns a verdict, merge
  it most-restrictive with the rule/command-safety decision via `more_restrictive` (`_rules.py`).
- `prepare_tools` / `bare_name_verdict` untouched (context decisions are per-call, not
  per-listing) -> `deny_removes_tool` behavior unchanged.
- No new dependency (stdlib + `pydantic_ai`).

## Test evidence

- **139 tests** in `tests/permission_policy/`, including adversarial red-team tests for each
  bypass above -- each was verified to fail against the old fixed-position/raw-token logic
  before the fix, then pass after it.
- **100% branch coverage** on `pydantic_ai_harness/permission_policy/*`.
- Full repo gate green: `ruff format --check`, `ruff check`, `pyright` (strict), and the full
  suite -- 3971 passed, 47 skipped, 0 failed, 100% branch coverage repo-wide (`make testcov`).
- No changes to `pyproject.toml` or `uv.lock`.

## Residual risk (stated plainly)

Ripgrep and `find` have larger short-option surfaces than `rm`, including flags that take an
attached value, so this analyzer does not decompose their bundled short options (e.g. `rg -oz`
bundling `-o` and `-z` evades the `-z` check). This is documented in `_safelist.py` and the
README rather than silently left closed. If this matters for your threat model, prefer a
stricter `default_verdict` for those tools or route them through `rules`.

## Checklist

- [x] No changes to `pyproject.toml` or `uv.lock`


Knucklessg1:
This is now a self-contained implementation against `main` that doesn't depend on #340. #340 has been quiet for a couple of weeks, so I retargeted this PR and reimplemented the `PermissionPolicy` primitives it needs directly, rather than continue to build on top of it.

While reimplementing the shell-command safety layer, I addressed the outstanding CodeRabbit findings on it -- most notably the CRITICAL one, where `_is_safe_sed` only validated `argv[1]`/`argv[2]` and never inspected anything appended afterward, so a destructive second `-e` script tacked onto an otherwise-safe `sed -n {N}p` passed unnoticed. That turned out to be one instance of a broader pattern in the module (flag matching not canonicalizing `--flag=value` forms, `rm`'s bundled short options, `sudo`/`doas`/`xargs` wrapper stripping, a command substitution hiding inside a double-quoted string). The PR description has the full list, the fix, and adversarial tests for each. One gap is called out as a known residual risk rather than silently closed (ripgrep/`find` short-option bundling) -- happy to discuss whether that needs closing before this merges.

I also noticed `main` picked up `ToolGuardrail` (#470) while this was in flight, which covers similar ground with a different design (generic guard callbacks vs. this PR's rule engine + shell-command analysis). I left both in the README rather than picking a winner -- that's a product call for you.

Happy to fold this into #340 instead if you'd rather go that direction once it's unblocked -- this is just meant to be reviewable and unblocked on its own in the meantime.

Issue #392, opened 2026-07-21
RFC: capability class naming convention + rename pass


strawgate:
Claude here: filing this as an RFC after a full sweep of the merged capability surface, at a maintainer's request.

## Summary

Codify a naming convention for capability classes and rename the rough outliers. Framed as an RFC -- the clear renames are high-confidence, but a set of borderline names is opened for maintainer input.

Swept all **22 user-facing `AbstractCapability` subclasses** on `main`. The good news: **no capability is named `*Capability`** -- that convention is already followed (the only `*Capability` identifiers are `AuthoredCapability`, an internal `BaseModel`, and `CapabilityValidationError`, both in `runtime_authoring` and legitimately *about* capabilities). What remains is uneven: imperative verb-phrase names, one abbreviation, `-ing` adjectival forms, and inconsistency within the `compaction` family.

## Motivation

A capability is a **noun the user adds**: `Agent(capabilities=[Shell(), FileSystem(), Memory()])`. Names like `ClearToolResults()` read as a *command*, not a thing; `PyaiDocs` uses a nonstandard abbreviation; `OverflowingToolOutput` uses an `-ing` adjective where the noun is the overflow. Within one module, `compaction` mixes nouns (`SlidingWindow`, `TieredCompaction`) with verbs (`ClampOversizedMessages`, `ClearToolResults`, `DeduplicateFileReads`).

## Proposed naming principles (to add to `agent_docs/capability-authoring.md`)

1. **Noun, not imperative verb** -- the class names the behavior/thing provided (`MessageTruncation`, not `ClampOversizedMessages`).
2. **No abbreviations** in the class name (`PydanticAIDocs`, not `PyaiDocs`).
3. **Prefer a plain noun over an `-ing` adjectival form** when one exists.
4. **Keep a family consistent** -- all members of a module read the same way.
5. Reads naturally at the call site in `capabilities=[...]`.

`capability-authoring.md` currently guides *module* names ("prefer a longer descriptive name, e.g. `overflowing_tool_output` not `overflow`") but says nothing about *class* names. This RFC fills that gap.

## Full inventory & recommendation

| Capability | Module | One-line | Recommendation |
|---|---|---|---|
| `CodeMode` | code_mode | sandboxed `run_code` | **Keep** (shipped) |
| `FileSystem` | filesystem | file tools | **Keep** (shipped) |
| `Shell` | shell | command execution | **Keep** (shipped) |
[...]

Issue #101, opened 2026-03-26
Dynamic Context Window Discovery & Persistence Guard


adtyavrdhn:
> Parent: #1 — [Tracking] Capabilities abstraction for pydantic-ai

## Goal

Dynamically discover actual model context window limits from provider error messages, and only persist **provider-confirmed** limits to disk — never cache guessed/fallback values.

## Context

- **Category:** Primitive (foundational infrastructure, extends #35)
- **Serializability:** N/A — internal infrastructure
- **Capability interface methods:** Affects `ModelProfile.context_window` resolution
- **Implementation status:** Not started
- **Blocked by:** #35 — Expose context window size on ModelProfile

### The problem

Static model metadata (from genai-prices or model registries) can be **wrong or outdated**:
- OpenRouter routing changes can alter effective context limits
- New model versions may increase/decrease limits without metadata updates
- Custom endpoints (Ollama, vLLM) have no registry entry
- Provider-imposed limits may differ from the model's theoretical maximum

When a wrong limit is cached permanently, agents either:
1. **Underutilize** context (conservative limit cached) — wastes capacity
2. **Overflow repeatedly** (optimistic limit cached) — breaks sessions

### How Hermes solves this

Hermes implements **adaptive tier stepping with a persistence guard** (commit `43af094a`):

1. **Start** with configured or cached limit for the model
2. **On context overflow error** (400/401 mentioning "context length"):
   - Parse the error message for the **numeric limit** (e.g., "exceeds limit of 200000 tokens")
   - If found: cache to disk (this is **provider-confirmed**)
   - Step down to next tier: 200K → 128K → 96K → 64K
3. **On fallback guess** (no numeric limit in error):
   - Use the stepped-down tier **in-memory only**
   - Never persist to disk — prevents wrong guesses from poisoning future sessions

The key insight: **only limits parsed from actual provider error messages are reliable enough to cache.** Heuristic fallbacks must stay ephemeral.

### Why this matters for pydantic-harness

Issue #35 plans to expose `context_window` on `ModelProfile`, auto-populated from genai-prices. But this is a static lookup — it doesn't handle:
- Models not in the registry (custom endpoints, new releases)
- Providers that impose lower limits than the model's max
- Registry data that's simply wrong

A persistence guard ensures that once the harness learns the *real* limit (from a provider error), it remembers it — but never remembers a guess.

### Proposed design

```python
class ModelProfile:
    context_window: int | None              # Static from registry
[...]

Issue #100, opened 2026-03-26
Tool Schema Token Accounting in Context Budget


adtyavrdhn:
> Parent: #1 — [Tracking] Capabilities abstraction for pydantic-ai

## Goal

Include tool definition tokens in context window budget calculations, preventing unexpected overflow when agents have large toolsets (50+ tools from MCP servers, multiple capabilities).

## Context

- **Category:** Primitive (foundational infrastructure)
- **Serializability:** N/A — internal accounting
- **Capability interface methods:** Affects #21 (Compaction) threshold detection and #46 (Cost/Token Budget)
- **Implementation status:** Not started
- **Blocked by:** #35 — Expose context window size on ModelProfile

### The problem

Context budget calculations typically sum `system_prompt_tokens + message_tokens` to decide when to compact or whether the context window is full. However, **tool schemas are also sent with every request** and can consume 20-30K tokens when an agent has 50+ tools (common with MCP servers, filesystem + shell + git + custom capabilities).

Ignoring tool schema tokens means:
- Compaction (#21) triggers too late — the window is already 20-30K tokens past the real limit
- Cost/Token Budget (#46) underestimates actual spend
- Context overflow errors occur unexpectedly despite compaction being "enabled"

### How Hermes solves this

Hermes recently fixed this (commit `43af094a`):

```python
def estimate_request_tokens_rough(messages, system_prompt="", tools=None):
    """Sums character counts of sys, msgs, AND tools, divides by 4."""
    total_chars = len(system_prompt) + sum(len(str(m)) for m in messages)
    if tools:
        total_chars += len(str(tools))  # <-- This was missing before
    return total_chars // 4
```

The fix was specifically motivated by production incidents where agents with 50+ tools (via MCP) would hit context overflow despite having compaction enabled, because the compaction threshold check didn't account for the ~25K tokens consumed by tool schemas.

### Why this matters for pydantic-harness

The harness is designed for composable capabilities, each of which can add tools via `get_toolset()`. A realistic production agent might have:
- FileSystem (#25): 5 tools
- Shell (#26): 1-2 tools
- Git (#62): 5+ tools
- MCP (#27): 10-50 external tools
- Code Execution (#33): 1-2 tools
- Memory (#30): 2-3 tools
- SubAgent (#32): 2-3 tools

That's 30-70 tools = 15-35K tokens of schema overhead **on every request**.

### Proposed design

Extend `RunUsage` or `ModelProfile` to expose tool schema token count:

[...]

PR #465, opened 2026-07-25
feat(compaction): resolve thresholds against the model's context window

### LINKED ISSUE #463: Capability: `ContextBudget` -- compaction thresholds relative to the model's real context window
### Capability Name

`ContextBudget`

### Description

Every compaction strategy in `pydantic_ai_harness.compaction` triggers on an **absolute** budget: `max_messages`, `max_tokens`. Nothing in the harness knows how big the model's context window actually is, so the number has to be hardcoded by the application:

```python
SummarizingCompaction(max_tokens=180_000)  # correct for which model, exactly?
```

That number is wrong the moment the agent runs on a different model. Set it for a 200K model and a 1M-context model compacts five times earlier than it needs to, paying for summaries it did not need and losing detail it could have kept. Set it for the 1M model and a 200K model blows the request before compaction ever fires.

`ContextBudget` closes that gap. It is a capability that:

1. **Resolves the run's context window** and turns a *fraction* into the absolute budget the existing strategies already understand. `genai-prices` — already in the tree via `pydantic-ai-slim` — carries this as `ModelInfo.context_window`, so no new dependency is required:

   ```python
   from genai_prices.data_snapshot import get_snapshot

   _, model_info = get_snapshot().find_provider_model(
       model_ref='claude-sonnet-4-6', provider=None,
       provider_id='anthropic', provider_api_url=None,
   )
   model_info.context_window  # 1_000_000
   ```

   Resolution happens in `for_run`, so it follows the model the run actually uses rather than whatever was known at construction. Unknown models (`LookupError`) and known models without a recorded window (`gemini-2.5-pro` returns `None` today) fall back to a documented conservative default.

2. **Reports context usage.** A callback — or an OTel event, whichever fits the harness's conventions better — fires with `(fraction, used_tokens, window_tokens)` before and after every model request. This is what drives a live "context: 73%" gauge, which every terminal assistant has and which no harness capability currently makes possible. Today an application has to re-count the history itself, duplicating `estimate_token_count` and guessing the denominator.

3. **Makes compaction callable outside a run.** `CompactionStrategy.compact(messages, ctx)` requires a `RunContext`, which a host application does not have between runs. It uses `ctx` only for `ctx.model` and `ctx.usage`, so an overload taking an explicit model would let an application compact on demand — the `/compact` command every coding assistant ships. Optionally with a `focus` hint folded into `summary_prompt`, so the user can say *what the summary must not lose*.

Points 1–3 are one capability in our implementation, but they are separable; happy to split into three PRs if that reviews better. Point 1 alone is the one that unblocks other issues.

### Why this belongs in the harness, not in applications

Point 1 is the missing half of `#35`. Both `#100` (tool-schema token accounting) and `#101` (dynamic window discovery) are marked *blocked by #35* — they are all reaching for "how big is the window, really". A resolver inside the harness unblocks all three without waiting on the core change: when `ModelProfile.context_window` lands (`pydantic-ai#4538`), `ContextBudget` switches to it in one place and the fallback becomes dead code.

### Deliberate non-goals

- **Not a new compaction strategy.** `ContextBudget` drives the existing ones (`SummarizingCompaction`, `SlidingWindow`, `TieredCompaction` …) via the `CompactionStrategy` protocol. It decides *when*, they decide *how*.
- **Not overflow handling.** Oversized single tool results stay `OverflowingToolOutput`'s job (`#185`, `#332`).
- **No caching of guessed limits.** Only registry- or provider-confirmed numbers are treated as reliable, per the persistence-guard argument in `#101`.

### Use Case

A terminal coding assistant, model chosen at startup, sometimes swapped mid-session.

**Today**, the application must pick a constant:

```python
Agent(model, capabilities=[SummarizingCompaction(max_tokens=180_000)])
```

On `claude-sonnet-4-6` (1M window) it compacts at 18% of capacity — every long session pays for summaries it did not need, and the model loses detail it had room to keep. On a 128K model the same agent never compacts before the provider rejects the request. The application cannot fix this generically, because it has no supported way to ask "how big is this model's window?" — `ModelProfile` has no `context_window` field as of 2.18.

**With `ContextBudget`:**

```python
from pydantic_ai import Agent
from pydantic_ai_harness.compaction import ContextBudget, SummarizingCompaction

agent = Agent(
    capabilities=[
        ContextBudget(
            strategy=SummarizingCompaction(keep_messages=20),
            threshold=0.9,
            on_usage=lambda fraction, used, window: status_bar.set(fraction),
        ),
    ],
)
```

The same code compacts at 900K on Sonnet and at 115K on a 128K model, and the status bar tracks it without the application re-counting anything.

The second half of the use case is the interactive one. When the user types `/compact` — or the model itself decides the history has gone stale — compaction has to run **outside** `agent.run()`, against the history the application holds:

```python
history = await budget.compact(history, focus='the auth refactor, not the earlier CSS work')
```

This is the part no current harness API supports, and it is why applications end up reimplementing summarization instead of using the capability they already have.

| Hook | Use |
|---|---|
| `for_run` | Resolve the context window against `ctx.model` and reset per-run state, so the budget follows the model the run actually uses (including a run that differs from construction time). |
| `before_model_request` | Count the pending history, report usage, and delegate to the wrapped `CompactionStrategy` when over threshold or when a compaction was requested. Reports usage a second time after a successful compaction, so a gauge shows the drop. |
| `get_toolset` | Optional single tool (`compact_conversation(focus)`) letting the model free context on its own terms. Off by default. |

Also relevant:

- **`CompactionStrategy`** — `ContextBudget` composes rather than replaces; anything satisfying the protocol can be wrapped, including `TieredCompaction`.
- **`estimate_token_count`** — reused as-is for the accounting, including the `tokenizer` override.
- **Span parity** — a driver-run compaction should emit the same `compact_messages` span `compact_with_span` emits, so instrumented runs look identical however compaction was triggered. `compact_with_span` is currently private to `compaction._shared`; exporting it (or letting the driver pass through it) would avoid a duplicated implementation.

**Sequencing note:** the natural implementation touches `_summarizing_compaction.py` and `_shared.py`, which `#339` is actively rewriting. Happy to land after it, or to build against its post-merge shape — whichever the maintainers prefer.

### Prior Art / References

**Related harness issues**

- [#35 Expose context window size on ModelProfile and usage on RunContext](https://github.com/pydantic/pydantic-ai-harness/issues/35) — the core-side fix. `ContextBudget` is the harness-side path that works before it lands, and the single place that switches over once it does.
- [#100 Tool Schema Token Accounting in Context Budget](https://github.com/pydantic/pydantic-ai-harness/issues/100) — blocked by #35; becomes implementable once the window is resolvable. Tool schemas are part of the same budget.
- [#101 Dynamic Context Window Discovery & Persistence Guard](https://github.com/pydantic/pydantic-ai-harness/issues/101) — blocked by #35; the provider-error-driven refinement layered on top of registry lookup. Its persistence guard ("only cache provider-confirmed limits") is the right rule and is why this proposal caches nothing.

**Related harness PRs**

- [#339 compaction: receipts, pin contract, keep-user-messages, anchored incremental summarization](https://github.com/pydantic/pydantic-ai-harness/pull/339) — overlapping files (`_shared.py`, `_summarizing_compaction.py`, `_sliding_window.py`). Sequencing coordination needed; no conceptual conflict.
- [#366 compaction: bound output retry context](https://github.com/pydantic/pydantic-ai-harness/pull/366) — another strategy that would sit under the same fraction-based trigger.
- [#332 overflow: grep/slice query surface + uniform explicit elision markers](https://github.com/pydantic/pydantic-ai-harness/pull/332) / [#185 ToolOutputManagement](https://github.com/pydantic/pydantic-ai-harness/pull/185) — the adjacent, deliberately separate concern (one oversized result vs. whole-history budget).

**Upstream Pydantic AI**

- [pydantic-ai#4538](https://github.com/pydantic/pydantic-ai/issues/4538) — context window on `ModelProfile`, usage on `RunContext`.
- [pydantic-ai#4137](https://github.com/pydantic/pydantic-ai/issues/4137) — first-class context compaction API.
- [pydantic-ai#4060](https://github.com/pydantic/pydantic-ai/issues/4060) — `ContextWindowExceeded` exception; the failure mode a fraction-based trigger exists to avoid.

**Existing implementations**

- [`summarization-pydantic-ai`](https://github.com/vstorm-co/summarization-pydantic-ai) (vstorm-co) — `ContextManagerCapability` has shipped this shape for ~a year: fraction trigger, `on_usage_update`, before/after-compaction hooks, an agent-callable compact tool. Its window resolution is the part that never worked properly, which is precisely the argument for solving it once, in the harness, where `genai-prices` already lives.
- [`pydantic-deep`](https://github.com/vstorm-co/pydantic-deepagents) (vstorm-co) — consumes the above to drive the CLI's context gauge and `/compact`. Being ported onto harness compaction now; this proposal is the piece that has no harness equivalent.
- **Hermes** — the source cited in #100/#101 for adaptive tier stepping and tool-schema accounting.
- **Claude Code / Codex CLI** — the established UX this enables: a persistent context percentage in the status line, plus a user-invoked `/compact [focus]`.

**Data availability**

`genai-prices` (already a transitive dependency of `pydantic-ai-slim`) exposes `ModelInfo.context_window`. Spot-checked on 0.0.72:

| Model id | Resolved window |
|---|---|
| `anthropic:claude-sonnet-4-6` | 1,000,000 |
| `anthropic:claude-opus-4-7` | 1,000,000 |
| `openai:gpt-4.1` | 1,000,000 |
| `openai:gpt-5.4` | 1,050,000 |
| `google-gla:gemini-2.5-pro` | `None` (known model, no recorded window) |
| `bedrock:us.anthropic.claude-sonnet-4-5` | `LookupError` (prefixed ref not matched) |

The last two rows are why the proposal specifies an explicit fallback rather than assuming a lookup always succeeds.

DEENUU1:
## Summary

Every size-based strategy in `compaction` triggered on an absolute budget (`max_tokens`,
`target_tokens`, `max_context_tokens`) that only an application that knows its model can set, and
which is wrong as soon as the model changes. Configured for a 200K model, a 1M-context model
compacts at a fifth of its capacity and pays for summaries it did not need; configured for the 1M
model, a 128K one never compacts before the provider rejects the request. Nothing in the harness
could resolve the window, so the number had to be a constant.

This adds `max_fraction` (`target_fraction` on `TieredCompaction`, `max_context_fraction` on
`WarnNearLimits`), resolved per run from the model's real context window:

```python
SummarizingCompaction(max_fraction=0.9, keep_messages=20)
```

The same configuration compacts at 900K on a 1M model and at 115K on a 128K one. `max_tokens` and
`max_fraction` are mutually exclusive: a strategy taking both would have to pick one and discard the
other, leaving the caller unable to tell which budget was in force.

Two additions the resolved window makes possible:

- **`ReportContextUsage`** reports `used_tokens` / `window_tokens` / `resolved` before each request,
  so an application can drive a context gauge without re-counting the history and guessing the
  denominator. It only observes; it never edits history. `used_tokens` counts message parts the same
  way the triggers do, plus the most recent `ModelRequest.instructions` once. Tool schemas are
  outside that count, which the field documents (tool schemas are #100's job).
- **`compact_now`** runs a strategy outside an agent run, for a host-driven `/compact`. A strategy's
  `compact` takes a `RunContext` an application holding a conversation *between* runs does not have,
  though it reads only `model` and `usage` from it. `focus` steers strategies that write prose, via a
  new `SummarizingCompaction.with_focus`, and is ignored by the ones that drop or blank content by
  rule, since they have nothing to steer. A compaction that changes the history emits the same
  `compact_messages` span the in-run path emits (pass `tracer=`), so an instrumented application sees
  one shape however compaction was triggered.

The window is resolved from `ModelRequestContext.model` -- the model the request will be sent to,
not the one the run started with. Core documents replacing it from a capability, so a model-routing
capability ordered ahead of compaction would otherwise route to one window and compact against
another.

Not every model resolves: `google-gla:gemini-2.5-pro` (known, no recorded window), Bedrock-prefixed
references, local endpoints, and any `FallbackModel` (its `model_id` is a composite `fallback:...`)
all come back empty. Every capability that takes a fraction therefore also takes
`fallback_context_window`, defaulting to a conservative 200K, so a deployment whose size you know is
not stuck compacting at a fifth of its capacity:

```python
SummarizingCompaction(max_fraction=0.9, fallback_context_window=1_000_000)
```

`resolve_context_window` is exported so applications can read the number directly. It is also the
single place to switch over when core grows `ModelProfile.context_window`
([pydantic-ai#4538](https://github.com/pydantic/pydantic-ai/issues/4538)); the fallback then becomes
dead code.

The registry is also wrong for some shipping models, in both directions. Against `genai-prices`
0.0.71, `anthropic:claude-sonnet-4-5` records 1M against a real 200K window, and
`anthropic:claude-opus-4-6` records 200K against a real 1M. Over-recording is the direction that
breaks a run, since the trigger lands above the real ceiling and compaction never fires. Both doc
surfaces name the affected models and say to pass `context_window=` explicitly on Sonnet-class
Anthropic models. Harness cannot correct the registry, so this is documented rather than
test-enforced: no test asserts a specific window number, so an upstream correction will not turn the
suite red.

**Scope note.** The change touches `_shared.py`, `_sliding_window_compaction.py`, and
`_summarizing_compaction.py`, which #339 is also rewriting. The overlap is mechanical -- one added
field and one changed call site per strategy -- with no conceptual conflict. Happy to land after #339
or rebase onto it, whichever you prefer.

### Boundary notes

**Output-limit accounting.** This changes what the limits resolve to, so the final accounting:

| Configuration | Trigger resolves to |
|---|---|
| `max_tokens=N` | `N` (unchanged behaviour) |
| `max_fraction=f`, window resolved | `max(1, int(window * f))` |
| `max_fraction=f`, window unresolved | `max(1, int(fallback_context_window * f))`, default 200K |
| neither set | `None` -- no token trigger, as before |

`max(1, ...)` keeps a tiny window from flooring the trigger to `0`, which the strategies reject as a
limit. `max_messages` is untouched. Existing configurations resolve exactly as they did: with
`max_fraction` unset, `resolve_token_trigger` returns `max_tokens` without consulting anything.

**No network in the request path.** The window comes from `genai_prices.data_snapshot.get_snapshot()`,
which reads the snapshot bundled with the package. `genai-prices` also ships `update_prices` /
`wait_prices_updated_*`, which do fetch over the network. None of them are called here. A model the
snapshot does not know raises `LookupError`, which is caught and reported as "unknown"; a known model
with no recorded window (`gemini-2.5-pro` today) returns `None` through the same path. Neither result
is cached, so a stale guess cannot outlive the process -- the persistence-guard rule from #101.

**Usage accounting for `compact_now`.** The throwaway `RunContext` gets a fresh `RunUsage` by
default, so a summarization call made outside a run does not silently vanish from anyone's counter:
pass `usage=` to accumulate it into your own. `deps=` is exposed for a strategy that reads them and
defaults to `None`.

`compact_now` allocates no resources beyond the context object, starts no processes, and reads no
files, so there is no cleanup path to fail.

## Linked Issue

Fixes #463

## Checklist

This PR moves our library built at Vstorm into pydantic-ai-harness https://github.com/vstorm-co/summarization-pydantic-ai

---

*Body updated by Claude Code on behalf of David, alongside the maintainer push described in
[this comment](https://github.com/pydantic/pydantic-ai-harness/pull/465#issuecomment-5112284151):
main merged in, `ContextUsageMonitor` renamed to `ReportContextUsage` per the #392 naming pass, and
the review findings applied. Only the statements this PR made that the code no longer matches were
changed.*


DouweM:
Naming note from the capability naming pass (#392, implemented in #480): the convention we adopted uses an imperative verb phrase when a capability acts on the run and one verb phrase states its entire contract, and avoids agent-nouns like `*Monitor`/`*Tracker` for single-behavior capabilities (we just renamed `CacheStabilityMonitor` to `WarnOnCacheBusts` for exactly this reason). `ContextUsageMonitor` reports context usage to the app before each model request, so `ReportContextUsage` would state the contract directly. The `ContextUsage` payload type keeps its noun name. Posted by Claude Code, acting on @DouweM's behalf.

dsfaccini:
This comment was posted by Claude Code using claude-opus-5 on behalf of David.

David has not reviewed these findings in detail -- they come from an automated review pass, so please push back on anything that looks wrong.

Reviewed at `1379ed6`. CI green including `coverage`, `lint` and `lowest-versions`; the 100% branch bar is met; doc snippets pass and every snippet's kwargs match the source and respect `kw_only`; the `pydantic/pydantic-ai#4538` and `pydantic/pydantic-ai-harness#100` links both resolve and say what the docs claim. Macroscope's "High" positional-signature finding is genuinely fixed by `8077608` and is not re-raised.

The config surface is fully backward compatible: the new fields are `kw_only`, positional constructor order is byte-identical on all five capabilities (`SlidingWindow`, `ClearToolResults`, `SummarizingCompaction`, `TieredCompaction`, `LimitWarner`), and `_shared.py::resolve_token_trigger` returns `max_tokens` verbatim without consulting anything when `max_fraction` is unset. `TieredCompaction.target_tokens` widened from required `int` to `int | None = None`.

Sourcing the window from `genai-prices` rather than a hardcoded table is the right call, and so is not waiting on `ModelProfile`: `ModelProfile` in the installed `pydantic_ai` 2.18.0 has no `context_window` field, `pydantic/pydantic-ai#4538` asks for exactly that, and the PR names it and isolates the switchover to one function. No network in the request path, and lookups are cached (~0.4us resolved, ~5us unresolved).

## Blocking

**1. The registry is wrong in both directions for shipping models, and the over-estimate makes compaction never fire.**

`_context_window.py::resolve_context_window`. Measured against the currently bundled `genai-prices` 0.0.71 and cross-checked with `platform.claude.com/docs/en/build-with-claude/context-windows` ("Other Claude models, including Claude Sonnet 4.5, have a 200k-token context window"; "For every model with a 1M-token context window, 1M is the default: you don't need a beta header"):

| model id | registry | real | direction |
| --- | --- | --- | --- |
| `anthropic:claude-sonnet-4-5` | 1,000,000 | 200,000 | 5x over (unsafe) |
| `anthropic:claude-opus-4-6` | 200,000 | 1,000,000 | 5x under (over-compacts) |
| `anthropic:claude-sonnet-4-6` | 1,000,000 | 1,000,000 | ok |
| `anthropic:claude-opus-4-5` / `4-7` / `4-8` / `haiku-4-5` | correct | | ok |
| `google:gemini-2.5-pro` (and all `google-gla:` / `google-vertex:`) | none -> 200K fallback | 1,000,000 | 5x under |

Failing scenario: `Agent('anthropic:claude-sonnet-4-5', capabilities=[SummarizingCompaction(max_fraction=0.9)])`. The trigger resolves to 900,000 tokens against a 200,000-token window. Compaction never fires and the run dies on a 400 `invalid_request_error` ("prompt is too long"). That is the exact failure the feature exists to prevent, produced by the recommended default value used in every doc snippet.

The docs describe this abstractly ("the registry records the maximum a model can be made to accept") and attribute it to beta or tier gating, but never name a concrete affected model. Here it is not gating: the registry entry is simply wrong. There is also no runtime signal, since `ContextUsageMonitor` exposes `resolved` but the strategies themselves surface nothing.

Harness cannot validate the registry itself. Minimum fix is a named-model warning in both docs plus guidance to set `context_window=` explicitly on Anthropic Sonnet-class models. The better fix is upstreaming corrections to `genai-prices`.

net-positive -> do. Not deferrable: this ships a default that silently disables compaction on a widely used model.

**2. The token estimator changed for every existing absolute-threshold user, with no migration note.**

`_shared.py::_collect_text` now also counts `ThinkingPart` / `CompactionPart` content, `RetryPromptPart` content, `NativeToolCallPart` / `NativeToolReturnPart`, and the most recent `ModelRequest.instructions`. The same history measures higher, so an unchanged absolute `max_tokens` config compacts earlier. Measured 2.39x on a synthetic thinking-heavy tool-calling history (10 turns, 6K-char thinking blocks, 8K-char instructions).

Concretely: `ClearToolResults(max_tokens=40_000)` now fires at what previously measured around 17K. What gets cleared is unchanged (`keep_pairs`, `min_clear_tokens`, `rebuild_with_cleared` untouched); only when.

Both `compaction/README.md` and `docs/compaction.md` frame the wider counting purely as a `max_fraction` accuracy improvement and say nothing about existing `max_tokens` deployments. Anyone on `SummarizingCompaction(max_tokens=N)` starts paying for summary LLM calls noticeably sooner after upgrading, with nothing in the release to explain it.

net-positive -> do. A few doc lines, and not doing it ships a behavior change users have to reverse-engineer.

**3. Three doc surfaces state the opposite of what the code does about instructions.**

`_context_usage.py::ContextUsage.used_tokens`'s docstring says instructions "are not part of the count", and the `ContextUsageMonitor` sections of both `compaction/README.md` and `docs/compaction.md` repeat it -- while both files say the opposite ("and the instructions, once") a few paragraphs earlier. Verified at runtime: a history with 4,000 chars of `instructions` estimates 1,100 tokens versus 100 without. A gauge built on the documented claim under-reports by exactly the instruction size.

net-positive -> do. Doc and comment only, and it is a factual error in a public docstring.

## Non-blocking

**4. Tests import private helpers, which CLAUDE.md forbids.** `tests/compaction/test_context_budget.py` imports `DEFAULT_CONTEXT_WINDOW` and `split_model_id` from `...compaction._context_window`, and `estimate_token_count`, `resolve_token_trigger`, `validate_token_trigger` from `...compaction._shared`. Two of those (`DEFAULT_CONTEXT_WINDOW`, `estimate_token_count`) are already re-exported from the public `pydantic_ai_harness.compaction` by this same PR, so those are one-line swaps. Three test classes (`TestSplitModelId`, `TestValidateTokenTrigger`, `TestResolveTokenTrigger`, 17 tests) exist only to drive private helpers directly, and around 40 `monkeypatch.setattr` calls target private module path strings.

net-positive -> do.

**5. No `Agent`-level test, and `resolve_context_window` is monkeypatched out of every window test.** The new 1141-line file contains zero `Agent(...)` constructions; hook tests use a local `_FakeCtx` dataclass and a one-field `_FakeModel` with `# type: ignore[arg-type]`. The consequence is that the entire class of bug in finding 1 is untestable by construction, because no test ever resolves a real model id through the real registry.

Two specific gaps: `TestModel` (`test:test`, the shape every downstream suite hits) is never used as the request model in a fraction test; and `TestResolveContextWindow.test_a_fallback_model_does_not_resolve` hand-writes `'fallback:openai,anthropic:fallback:gpt-4o,claude-sonnet-4-6'`, which is not the format a real `FallbackModel` emits (`fallback:test:test,test:test`). It would keep passing if core changed the real format to something that does resolve.

net-positive -> do. At minimum one test each for a real `TestModel` and a real `FallbackModel` instance.

**6. `max_fraction` is effectively inert under `TestModel`.** `test:test` does not resolve, so `max_fraction=0.9` becomes a 180,000-token trigger. A user testing their compaction config with `TestModel` (which `conftest.py` mandates) sees it never fire unless they build a 720K-character history or pass `context_window=` / `fallback_context_window=`. Nothing in the docs says so.

net-positive -> do. One doc line.

**7. `ContextUsageMonitor` and `ContextUsage` are missing from the `## API reference` autodoc block in `docs/compaction.md`.** The prose documents them fully; the `:::` list was never updated.

net-positive -> do.

**8. Two parity nits.** `docs/compaction.md`'s `LimitWarner` section names only `max_iterations` / `max_context_tokens` / `max_total_tokens`, while the README's parallel section also names `max_context_fraction`. And the newly exported `SupportsFocus` protocol is never named in either doc, unlike `CompactionStrategy` which is.

net-positive -> do. One line each.

**9. `genai-prices` is used but is not a declared dependency of harness.** `pyproject.toml` lists only `httpx` and `pydantic-ai-slim>=2.18.0`; `genai-prices>=0.0.70` arrives transitively through slim. Harness now calls `data_snapshot.get_snapshot` / `find_provider_model` and reads `ModelInfo.context_window`, none of which are in `genai_prices.__all__` (`('Usage', 'calc_price', 'UpdatePrices', 'wait_prices_updated_sync', 'wait_prices_updated_async', '__version__')`). CI's `lowest-direct` job does not lower transitive deps, so this floor is untested. A rename upstream, or slim dropping the dep, breaks harness at runtime with nothing in the manifest to catch it.

net-positive but out-of-scope -> defer, and this one needs a tracking issue to be a valid defer: add `genai-prices>=X` as a direct dependency. AICAs cannot touch `pyproject.toml` in this repo, so it has to be an issue.

**10. The estimator now has an irreducible floor.** Instructions are counted but no strategy can remove them (`find_token_cutoff` drops whole messages; `_instructions_text` re-reads the surviving tail). A trigger below the instruction size means every request compacts forever, which for `SummarizingCompaction` is an LLM call per turn. Only reachable with a small `context_window` / `fallback_context_window` or a small fraction.

net-neutral -> skip. Real but requires a pathological config; naming it is the useful part.

**11. A nested tier's own `max_fraction` / `context_window` / `fallback_context_window` are silently ignored.** `TieredCompaction._escalate` calls `tier.compact(...)` unconditionally, and a tier's trigger lives in its `before_model_request`. Pre-existing with `max_tokens`, unchanged here, and documented in the `CompactionStrategy` protocol docstring -- but this PR adds three more silently inert knobs.

net-neutral -> skip.

## Unknown-model paths, all traced

`TestModel` -> `model_id == 'test:test'` -> unresolved -> 200K fallback. `FallbackModel` -> composite `fallback:test:test,test:test` -> unresolved. Bedrock-prefixed ids and OpenAI-compatible or local endpoints -> unresolved. A registry entry with a `None` or non-positive window (`google:gemini-2.5-pro` today) -> treated as unresolved. All fall to `fallback_context_window`. That handling looks right.

## Decisions worth a maintainer sanity check

- **`compact_now` plus `TieredCompaction` on a short history is a no-op.** Macroscope flagged that a user-issued `/compact` does nothing when already under target, and `47c146f` resolved it by documenting the behavior rather than bypassing the gate. Defensible, but it means the `/compact` UX the feature is pitched for silently does nothing on short conversations unless the user passes the tier directly.
- **`ClampOversizedMessages` deliberately has no fraction family**, because it triggers per-part rather than on total history. Correct scoping, not an omission.
- **`FilePart` is deliberately excluded from counting** (binary payload). Reasonable, but it means an image-heavy history under-reports.

dsfaccini:
This comment was posted by Claude Code using claude-opus-5 on behalf of David.

Pushed the merge, the rename, and findings 1-8 straight to this branch (`1379ed6..8148e79`, three commits on top of a merge). @DEENUU1 -- sorry for landing on your branch rather than relaying; David okayed the direct push, same as on #404. Shout if you'd rather have it as a separate branch to cherry-pick.

I re-validated every finding against the branch HEAD before touching anything. None had gone stale.

**Merge, not rebase** (`36caea5`). All six conflicts were #480 fallout: `_limit_warner` -> `_warn_near_limits`, `_sliding_window` -> `_sliding_window_compaction`, `LimitWarner` -> `WarnNearLimits`, `SlidingWindow` -> `SlidingWindowCompaction`. Resolved by taking main's names and re-applying this PR's content on top, then sweeping the PR-owned files that merged clean but still carried the old spellings (`test_context_budget.py`, both compaction docs).

**Rename** (`228bde2`). `ContextUsageMonitor` -> `ReportContextUsage`, per @DouweM's note above; `ContextUsage` keeps its noun name. Module renamed to `_report_context_usage.py`, mirroring how #480 moved `_limit_warner.py` alongside its class. No deprecation alias: the `_RENAMED` shims on main cover names that have shipped, and this one never has.

**Docs** (`f2903c5`) -- findings 1, 2, 3, 6, 7, 8.

Finding 1 is the one worth a look. I re-measured against the installed `genai-prices` 0.0.71 and the table in the review still holds:

| model id | recorded | real |
| --- | --- | --- |
| `anthropic:claude-sonnet-4-5` | 1,000,000 | 200,000 |
| `anthropic:claude-opus-4-6` | 200,000 | 1,000,000 |
| `google:gemini-2.5-pro` | unrecorded | 1,000,000 |

I spot-checked `opus-4-5`/`4-7`/`4-8`, `haiku-4-5`, `sonnet-4-6`, `gpt-4o` and `gpt-5.1` -- all correct, so both docs name only the wrong ones, say that over-recording is the direction that kills a run, and tell you to pass `context_window=200_000` on Sonnet-class Anthropic models. Harness can't fix the registry, so this is documented rather than test-enforced: no test asserts a specific window number, so a legitimate upstream correction won't turn the suite red.

Finding 2 got a migration note on both surfaces. The wider counting means an unchanged absolute `max_tokens` compacts earlier; I measured 4.2x on a thinking-heavy synthetic history against the review's 2.4x on its own, and since the multiplier is entirely history-shaped the docs say "several times" rather than pinning a number.

**Tests** (`8148e79`) -- findings 4 and 5.

The ~40 `monkeypatch.setattr` calls on private module paths collapse to one helper that patches `genai_prices.data_snapshot.get_snapshot` instead, so the real `resolve_context_window` path stays under test rather than being stubbed out. `TestSplitModelId` / `TestValidateTokenTrigger` / `TestResolveTokenTrigger` are gone, their coverage re-expressed through public constructors and hooks. No `# pragma: no cover` was needed -- the one genuinely unreachable branch (`resolve_token_trigger`'s `model is None`) disappeared once I narrowed the parameter, since every call site passes a non-optional `model`.

New `TestThroughAnAgentRun` covers fraction compacting and not-compacting through a real `Agent(TestModel(), ...)`, and `test_a_fallback_model_does_not_resolve` now builds a real `FallbackModel` and asserts on its actual `model_id` rather than a hand-written composite string.

### Left alone

Findings 10 and 11 stay skipped as triaged. Finding 9 (`genai-prices` should be a direct dependency rather than transitive through slim) is real -- none of the three symbols we use are in `genai_prices.__all__` -- but AICAs can't touch `pyproject.toml` here, so it wants its own issue.

The PR body still describes `ContextUsageMonitor` and will need a pass before merge.

Verification: `make lint`, `make typecheck` (pyright strict), `make test` and `make testcov` green locally, 100% branch coverage bar met; CI green on `8148e79`.

PR #486, opened 2026-07-28
feat: Browser Use

laithrw:
## Summary

Adds `BrowserUse`, a capability that gives an agent a real web browser through the [Browser Use](https://browser-use.com) CLI.

- One `browser_exec` tool: the model writes Python, the CLI runs it inside a real web browser -- your local Google Chrome browser, a headless Chromium browser, or a browser running in the cloud.
- The CLI runs as a subprocess, so it adds no Python dependencies.

## Checklist


DouweM:
After discussing with Laith on Slack, we're going the SDK route as in #419.

PR #500, opened 2026-07-29
fix(planning): avoid leading CachePoint without user content

### LINKED ISSUE #438: Planning tail CachePoint can be the first user content block (UserError on Anthropic/Bedrock)
This issue was posted by Claude Code using claude-opus-4-8 on behalf of David. David directed the session and confirmed the finding, but has not re-checked it line by line.

## Summary

`Planning.wrap_model_request` in `pydantic_ai_harness/planning/_capability.py` appends an ephemeral tail `UserPromptPart` whose content leads with a `CachePoint`:

    last = messages[-1]
    if isinstance(last, ModelRequest):
        reminder = UserPromptPart(content=[CachePoint(ttl=self.cache_ttl), _reminder_text(render_plan(items))])
        messages[-1] = replace(last, parts=[*last.parts, reminder])

The only guard is `isinstance(last, ModelRequest)`. It does not check that `last.parts` already contributes a user-mappable content block. Anthropic and Bedrock reject a `CachePoint` that is the first content of a user message (`AnthropicModel._add_cache_control_to_last_param` raises `UserError('CachePoint cannot be the first content in a user message ...')`), so a tail `ModelRequest` whose only parts map to no user content (a system-prompt-only tail, an empty user prompt, or an instructions-only request) presents the leading `CachePoint` as the first block and aborts the run.

A `if not items: return` guard earlier in the method delays injection until a plan exists (`write_plan` has run), so this is not reachable on the first request the way it is for a fire-on-every-request reminder. But once a plan exists, a later request whose tail `ModelRequest` carries no user content still hits it.

## Scope

Pre-existing on `main`, independent of and out of scope for https://github.com/pydantic/pydantic-ai-harness/pull/414 . The same pattern was present in that PR's `SystemReminders` and was guarded there during review; this issue records the sibling occurrence in `Planning` for a maintainer to decide on separately. It does not prescribe a fix.

## Reference

- https://github.com/pydantic/pydantic-ai-harness/blob/main/pydantic_ai_harness/planning/_capability.py -- `Planning.wrap_model_request`

dk3yyyy:
## Summary

- only prepend the planning-tail `CachePoint` when the request already contains provider-mappable user content
- preserve cache boundaries for user prompts, tool returns, retry prompts, typed text, and media
- omit the invalid leading marker for system-only, empty, or cache-marker-only content
- update the capability documentation to describe the conditional breakpoint

## Linked Issue

Fixes #438

## Checklist

## Tests

- `make lint`
- `make typecheck`
- `make test` (3,365 passed, 45 skipped)
- `make testcov` (3,365 passed, 45 skipped; 100% coverage)

### DIFF
--- a/pydantic_ai_harness/planning/_capability.py
+++ b/pydantic_ai_harness/planning/_capability.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 
 from collections.abc import Callable, Sequence
 from dataclasses import dataclass, field, replace
-from typing import TYPE_CHECKING, Literal
+from typing import TYPE_CHECKING, Literal, cast
 
 from pydantic_ai.capabilities import AbstractCapability
 from pydantic_ai.messages import (
@@ -13,6 +13,7 @@ from pydantic_ai.messages import (
     ModelRequestPart,
     ModelResponse,
     TextContent,
+    UserContent,
     UserPromptPart,
 )
 from pydantic_ai.tools import AgentDepsT, RunContext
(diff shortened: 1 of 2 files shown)


dk3yyyy:
Closing this as superseded by #570 ("Fix `Planning` cache-point order"), which fixes the same underlying issue #438 with a simpler approach — placing a stable `<plan-reminder>` opening tag before the `CachePoint` so the breakpoint is never leading.

For the record, the Macroscope review comment on `2a2c121` (early return on any `CachePoint`) was addressed in `f695b70` — `_with_cache_point` now verifies a breakpoint is preceded by non-empty text content and displaces/removes invalid ones, and CI was green. Keeping the `fix/planning-cachepoint-first-content` branch in case the cache-point normalization logic is ever wanted.

PR #513, opened 2026-07-30
feat: Browser Use

### LINKED ISSUE #418: Capability: BrowserUse -- delegate open-ended web tasks to a browser-use agent
### Capability Name

BrowserUse

### Description

Adds one tool, `browse_web(task)`, that hands a self-contained natural-language
goal to an autonomous [browser-use](https://github.com/browser-use/browser-use)
agent. That agent drives a real Chromium over CDP with its own
perception-action loop (indexed DOM, screenshots, planning, loop detection,
self-healing) and returns a text result, or JSON when an output schema is
configured.

**The problem.** Scripted browser tools (goto, click a selector, extract text)
are cheap and deterministic when the flow is known. On an unknown page layout
or a fuzzy goal ("find the price of the Pro plan", "fill in this form") the
host model ends up micro-managing a DOM it cannot perceive well, paying a
model round-trip per click and getting stuck on dynamic pages.

**The approach.** browser-use is an agent, not a tool library, so the
capability delegates to it rather than unbundling it into low-level tools --
the same shape `ExaAgent` and `Subagents` already use in this repo. Unbundling
would duplicate the scripted-browser story while pulling in a much heavier
dependency tree.

Notable design points:

- **Model wiring.** `llm` accepts a Pydantic AI model or model name, wrapped
  in a `PydanticAIChatModel` that implements browser-use's chat-model
  protocol. One provider configuration for host and sub-agent, Logfire spans
  for the sub-agent's calls, and structured output through Pydantic AI's tool
  calling -- which also sidesteps browser-use's forced `response_format`
  schema that some backends reject (Anthropic models behind OpenRouter reject
  `minimum` on an integer field). browser-use's own wrappers (`ChatAnthropic`
  and friends) are accepted as-is. Without an explicit model browser-use falls
  back to its hosted `ChatBrowserUse`, which is a separate account and bill,
  so the docs call that out.
- **Complete configuration.** The capability promotes the options it owns
  (allowed domains, headless, vision, secrets, output schema, step budget,
  sub-agent system-message extension), and `BrowserAgentSettings` mirrors
  every remaining `browser_use.Agent` option at its upstream default, so
  judge, planning, timeouts, and custom action registries need no custom
  factory.
- **Session lifetime.** `session_scope='call'` gives every call a fresh
  session killed in a `finally` (concurrent calls then run in parallel);
  `session_scope='agent'` keeps one session alive across calls behind a lock,
  so tabs and logins carry over until `aclose()` or an `async with` block
  exits.
- **Safety.** `allowed_domains` is enforced by browser-use's `BrowserProfile`
  rather than by prompt, and `sensitive_data` lets the sub-agent type
  credentials the model never sees.

The extra is marked `python_version >= '3.11'` (browser-use's floor; the
harness supports 3.10), and its tests skip at collection when the package is
absent, matching the `exa` pattern.

Implementation is complete on a branch (100% branch coverage, pyright strict
clean, verified end to end against a real Chromium); a PR follows if the
direction is agreed.

### Use Case

A support-ops agent has to check a carrier's tracking page for a customer
order. The page is behind a cookie wall, renders results into a container
whose markup changes between carriers, and sometimes asks to confirm a
region. Writing selectors for each carrier is a maintenance treadmill.

````python
from pydantic_ai import Agent

from pydantic_ai_harness.browser_use import BrowserUse

agent = Agent(
    capabilities=[
        BrowserUse(
            llm='anthropic:claude-sonnet-4-6',
            allowed_domains=['*.carrier.example'],
        )
    ],
)

result = agent.run_sync('Where is parcel 1Z999AA10123456784 right now?')
````

The host agent stays high-level: it formulates one goal, and the sub-agent
handles the cookie wall, the search box, and the result layout, returning the
status text. The host then answers the customer in their language.

Two more shapes the capability is built for:

- **Structured extraction.** `output_schema=Product` makes the sub-agent
  return validated JSON, so the host gets typed data instead of prose, and a
  parse failure surfaces as a retry rather than malformed output.
- **Authenticated flows.** `sensitive_data` plus `allowed_domains` lets an
  agent log in to an internal tool without the sub-agent's model ever seeing
  the credentials, and `session_scope='agent'` keeps that login alive across
  several follow-up tasks in one conversation.

No agent-graph lifecycle hooks. The capability contributes:

- **`get_toolset()`** -- a `FunctionToolset` providing the single `browse_web`
  tool. It owns the browser session lifecycle, so it is built once and cached
  on the capability (an `'agent'`-scoped shared session needs exactly one
  owner).
- **`get_instructions()`** -- short static delegation guidance for the host
  model (when to reach for `browse_web`, how to phrase a goal). Replaceable
  via `guidance`, disabled with `''`, following `ExaSearch`.
- **`from_spec()`** -- agent-spec construction for the serializable subset of
  options. The model, browser profile, output schema, agent settings, and
  factory are not spec-serializable.
- **Application lifecycle**: `aclose()` and async-context-manager support, for
  the shared-session scope. Nothing else in the agent loop is intercepted.

Extension seam: `browser_agent` takes a `BrowserAgentFactory` receiving a
`BrowserTask` dataclass, for browser-use options beyond the settings mirror
(callbacks, skills, injected state) and for substituting a fake in tests so no
browser launches.

### Prior Art / References

- [browser-use](https://github.com/browser-use/browser-use) (MIT, v0.13.6) --
  the delegated agent. Relevant surfaces: `Agent` (`agent/service.py`),
  `BrowserSession` (`browser/session.py`), `BaseChatModel` (`llm/base.py`),
  `AgentHistoryList` (`agent/views.py`).
- [browser-use documentation](https://docs.browser-use.com)
- `ExaAgent` in this repo (`pydantic_ai_harness/exa/_agent.py`) -- the
  existing precedent for delegating to an external agent behind one tool.
- `Subagents` (`pydantic_ai_harness/subagents/`) -- the in-process delegation
  counterpart.
- [Pydantic AI capabilities](https://pydantic.dev/docs/ai/capabilities/overview/)
  and [toolsets](https://pydantic.dev/docs/ai/tools-toolsets/toolsets/).
- Complementary low-level approach for comparison: the Playwright toolset in
  [pydantic-deep](https://github.com/vstorm-co/pydantic-deepagents), which is
  proposed separately. The two are not competitors: scripted tools for known
  flows, `BrowserUse` for goals that need judgement about unfamiliar pages.

laithrw:
## Summary

Builds on #419: Browser Use Cloud support, live progress narration, and a docs rework.

- **Cloud browsers**: set `use_cloud=True` to run each session in a [Browser Use Cloud](https://cloud.browser-use.com) browser instead of a local Chromium -- for sites that block automated traffic, or to take the run fully off your machine and fan out many browsers in parallel. Needs `BROWSER_USE_API_KEY`; the browser is provisioned when a session starts, released when it stops, and follows `session_scope` like a local one. Spec-serializable.
- **Live progress**: set `progress` (e.g. `progress=print`) to watch the task and each step's stated goal live while a call runs -- a `browse_web` call can run for minutes and was silent until it returned. Wired through a new `BrowserTask.on_step` field that the default factory forwards as browser-use's `register_new_step_callback`.
- **Docs**: README and docs page restructured to match the harness capability-doc shape (problem/solution, what agents do with it, choosing a browser, cloud setup incl. Chrome-profile sync, configuration block), with deep links into Browser Use's docs (prompting guide, supported models, `BrowserProfile` parameters, authentication, output format, sensitive data) instead of re-documenting its surface. Docstrings trimmed throughout.

Verified end to end with real LLM runs: a host agent delegating to `browse_web` in a real headless Chromium (default path and `output_schema` path), on this branch's locked browser-use version.

### Boundary notes

`use_cloud` folds into `BrowserProfile` (the session's typed overloads split cloud and local kwargs) and rides the existing session lifecycle: `BrowserSession.kill()` dispatches `BrowserStopEvent`, whose handler calls the cloud client's `stop_browser`, so the existing shielded, timeboxed teardown (with pending-cleanup retries) stops a billed cloud browser even on a failed or cancelled run. The local-launch `headless=True` default is skipped for cloud sessions; a cloud browser has no local window.

## Linked Issue

Fixes #418

## Checklist

- [ ] Tests added/updated for new behavior (coming in a follow-up commit)


dsfaccini:
hey @laithrw I'm closing this in favour of https://github.com/pydantic/pydantic-ai-harness/pull/560

PR #514, opened 2026-07-30
Recognize externalized `media` markers by shape, not by sentinel alone

### LINKED ISSUE #447: media restore: sentinel-key collision can break or corrupt snapshot restore
This issue was posted by Claude Code using claude-opus-4-8 on behalf of David.

**Problem.** `restore_media` in `pydantic_ai_harness/media/_walker.py` treats *any* dict anywhere in the message tree carrying the sentinel key (`__harness_external_media__` today, and `__harness_external_text__` as of #446) as an externalized marker and resolves its `uri` via `media_store.get`. A message part whose own payload legitimately contains that key collides with the marker protocol.

Reachable through the public API (verified): a `ToolReturnPart` whose `content` is
`{'__harness_external_media__': True, 'uri': 'media+sha256://000...0'}`
survives `ModelMessagesTypeAdapter.dump_json`, and on `latest_snapshot` either raises `FileNotFoundError` (snapshot won't load -> run can't resume) or, if the embedded `uri` collides with a real digest, silently inlines the wrong bytes into the tool return.

**Pre-existing.** The binary marker and the "any dict with the sentinel = marker" restore logic predate #446 (`git show main:.../media/_walker.py`). #446 widened the externalize side to large text parts (adding `__harness_external_text__`); it did not introduce the restore collision. Filing separately rather than blocking #446.

**Mitigation options (non-trivial; an in-band JSON sentinel is inherently collidable).**
- Restrict the walker to known structural positions (the `parts` arrays / known `part_kind` dicts) rather than any dict in the tree.
- Self-verify on restore: re-hash the resolved bytes and require `media_uri_for(bytes) == uri`, turning a silent wrong-bytes restore into a loud error (does not fix the false-positive `FileNotFoundError`, but closes the corruption path).

**Origin.** Adversarial review of #446.

dsfaccini:
This pull request was posted by Claude Code using claude-opus-5 on behalf of David.
David: no line review (he picked the issue, I implemented).

Closes #447

## Summary

`restore_media` decided "this dict is an externalized marker" from the sentinel key alone. `ToolReturnContent` permits any `Mapping[str, ...]`, so a tool could return `{'__harness_external_media__': True, 'uri': ...}`, that payload survives `ModelMessagesTypeAdapter.dump_json`, and on the way back in the walker resolved the payload's own `uri` through the `MediaStore` -- raising `FileNotFoundError` and failing the snapshot load, or, if the embedded URI matched a real digest, inlining unrelated bytes into the tool return.

`_is_external_marker` now requires the node to be shaped like something `externalize_media` actually wrote: the same discriminator the externalize side gates on, plus the absence of the field that was moved out.

- binary marker: `kind == 'binary'` and no `data`
- text marker: `__harness_external_text__` plus a string `part_kind` (or `kind == 'text-content'`), and no `content`

The externalize side is unchanged. The two sides are now symmetric -- restore accepts exactly the shapes externalize produces.

**Backward compatible.** Markers written by every prior release are binary and preserve `kind == 'binary'` (`_preserve_fields` drops only the externalized field), so persisted snapshots keep restoring. Covered by the existing legacy-marker tests, which pass untouched.

## Why not the re-hash mitigation

#447 offered a second option: re-hash the resolved bytes on restore and require `media_uri_for(bytes) == uri`. **That does not close the corruption path it was proposed for**, and I skipped it rather than ship a guard that reads as protection it does not provide.

The stores are content-addressed: `get(uri)` returns bytes whose sha256 *is* the digest in `uri`, by construction. In the collision scenario the walker resolves the payload's embedded URI, the store returns the blob that legitimately lives at that digest, and re-hashing it reproduces that same URI. The check passes and the wrong bytes are inlined anyway.

Re-hashing would catch a *different* fault -- a store handing back corrupted or truncated bytes. That is a real integrity property, but it belongs to the store, not the walker, and `MongoMediaStore` already length-verifies its reassembly. Not in scope here; happy to file it if you want it as its own issue.

## Residual, stated plainly

An in-band JSON sentinel is collidable by construction and this narrows the collision rather than closing it. A payload that sets the sentinel **and** mimics a serialized part shape **and** omits the field the walker moves out is still read as a marker. Closing that entirely means out-of-band framing (a marker table beside the snapshot, or a reserved envelope the message tree cannot express), which is a `StepStore` record-format change -- a much larger swing than this bug warrants. The docstring on `_is_external_marker` and both docs pages say so rather than implying the hole is sealed.

## Tests

Three regression tests, all through the public API surface:

- `test_tool_return_payload_carrying_the_sentinel_is_not_a_marker` -- the exact #447 repro, built from a real `ToolReturnPart` through `ModelMessagesTypeAdapter`. The embedded URI points at a **real** blob in the store, so the test asserts the payload is left alone because it is not marker-shaped, not because the reference happened to be dead.
- `test_sentinel_payload_keeping_its_content_is_not_a_text_marker`
- `test_sentinel_payload_keeping_its_data_is_not_a_binary_marker`

`test_restore_raises_when_marker_missing_uri` changed shape: its fixture was a bare sentinel dict, which is now correctly not a marker. It gains `kind: 'binary'` so it still exercises the loud failure on a genuinely malformed marker.

Full suite green, 100% branch coverage, no live services.

## Docs

`media/README.md` and `docs/media.md` updated in step (one sentence each, same text), stating both the fix and the residual.


.agents/skills/branch-context/append-handoff.sh
dsfaccini: weird stuff snuick in here, pelase remove

tests/media/test_media.py
```diff
@@ -666,6 +667,59 @@ async def test_unusable_uri_key_does_not_fall_back_to_a_genuine_uri(self, tmp_pa
...
+        """A tool return whose own content mimics a marker survives the round trip.
+
+        `ToolReturnContent` permits any mapping, so a tool can return the
+        sentinel key and a `uri` of its own. Resolving that through the store
+        would either fail the snapshot load or inline unrelated bytes into the
+        tool return -- the payload has to stay data.
+        """
+        import json as _json
+
+        from pydantic_ai.messages import ModelMessage, ModelMessagesTypeAdapter, ModelRequest, ToolReturnPart
```
dsfaccini: whyt he ibnline deps? this is an anti pattern in this repo

dsfaccini: This reply was posted by Claude Code using claude-opus-5 on behalf of David.

Fixed in `a49d928`, and it turned into more than the two lines.

Straight answer on the why: I copied the surrounding file. `tests/media/test_media.py` already had 20 inline imports before this PR, so the new test matched its neighbours rather than the repo standard. That is an explanation, not a defence -- matching a local habit is how an anti-pattern spreads.

Hoisting only my own imports did not work: the module-level `from pydantic_ai.messages import ...` then collided with the same names re-imported inline two tests below, and pyright failed on `ModelMessage` being unused. Half-hoisted was worse than either end state, so the whole file moved -- `json`, `base64`, and all `pydantic_ai.messages` names now import at module scope, and the `_json` alias is gone.

Net -18 lines. `make format`, `make lint`, `make typecheck` clean, full suite green at 100% branch coverage.

Two things left alone deliberately:

- `tests/media/test_media.py:799` keeps its inline `from pydantic_ai_harness.media._s3 import _canonical_uri` -- it carries a `# pyright: ignore[reportPrivateUsage]` and is scoped to the one test that reaches into a private helper. Hoisting it would widen a private import to the whole module.
- The rest of the repo still has this pattern in 14 other test files (`tests/code_mode/test_code_mode.py` has 57, `tests/compaction/test_compaction.py` 24). Not swept here -- that is a mechanical repo-wide cleanup, not part of a marker-collision fix. Say the word and it goes in its own PR.

pydantic_ai_harness/media/_walker.py
```diff
@@ -97,7 +97,33 @@ def _is_text_part(node: dict[str, object]) -> bool:
 
 
 def _is_external_marker(node: dict[str, object]) -> bool:
```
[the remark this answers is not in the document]
dsfaccini: This reply was posted by Claude Code using claude-opus-5 on behalf of David.

Confirmed and fixed in `cddbd90`. Real bug, and mine -- introduced by this PR.

`_preserve_fields` copies every source field onto the marker, so a binary payload that already carried a `__harness_external_text__` key handed its own marker that key. Restore then took the text branch, failed the text-shape check, returned "not a marker", and left the bytes stranded behind an unresolved URI.

Took the second suggested route rather than stripping the key. `_is_external_marker` is now `_marker_kind`, returning `text | binary | None`: the text branch requires a text *shape* (`part_kind`, or `kind == "text-content"`), so a binary marker falls through to the binary branch regardless of the flag. `_restore_external` takes that kind as an argument instead of re-reading `_TEXT_MARKER`, so the detection and the restore cannot disagree about which marker they are looking at.

Stripping the key on the externalize side would also have worked for restore, but it silently deletes a field the caller sent, which is the same class of data loss `__harness_external_uri__` was introduced to avoid.

One consequence worth naming, not in the original finding: `_TEXT_MARKER` was being dropped unconditionally on restore. On a binary marker that key is the caller's data, not ours, so the round trip was losing it. It is now dropped only on text markers.

Regression test `test_binary_part_carrying_the_text_marker_still_round_trips` covers exactly this shape. Full suite green at 100% branch coverage, and green again under a slim install.

.agents/skills/branch-context/append-handoff.sh
dsfaccini: This reply was posted by Claude Code using claude-opus-5 on behalf of David.

Already gone -- removed in `adbe64a`, a few minutes after you flagged it. The PR diff has zero `.agents/` files now.

Cause, so it does not recur: `ph-worktree` runs `copy-harness-config.sh`, which replaces these git-tracked files with symlinks into the tooling repo and deletes `adopt-pr/SKILL.md`. A fresh worktree is therefore dirty before any work starts, and my `git add -A` staged all of it as deletions. Restored from `main` rather than force-pushed.

Logged as a papercut -- the fix belongs in `copy-harness-config.sh` (add those paths to `.git/info/exclude`, or use `skip-worktree`), so a tracked-file substitution never shows up as a deletion in the first place.

tests/media/test_media.py
dsfaccini: This reply was posted by Claude Code using claude-opus-5 on behalf of David.

Both halves are done.

**This file:** fixed in `a49d928`. Hoisting only my own imports was not enough -- the module-level `from pydantic_ai.messages import ...` then collided with the same names re-imported inline two tests below, so the whole file moved. `json`, `base64` and all `pydantic_ai.messages` names now import at module scope and the `_json` alias is gone.

**The repo:** the pattern was in 14 other test files, so a lint rule rather than one more hand-corrected file. `PLC0415` is now enabled and the 146 existing sites are cleared -- #518, which has merged into this branch (that is why this PR now shows 26 files).

It already caught something: `tests/guardrails/test_tool_guardrail.py` landed on `main` while #518 was in flight, carrying two inline imports. CI lint failed here on the merge, and `cddbd90` hoists them.

Three exemptions, each for a stated reason rather than convenience: `pydantic_ai_harness/**` (deferred imports there are architectural -- `__getattr__` lazy exports that keep `pymongo`/`modal` out of the base install), `tests/stackone/**` (`mcp`/`fastmcp` ship only with the `stackone` extra, so module-scope imports break slim CI), and 10 `# noqa: PLC0415` sites where importing is itself the assertion.

pydantic_ai_harness/media/_walker.py
```diff
@@ -147,6 +198,7 @@ async def _maybe_externalize_binary(
     # are recursively walked so any nested externalizable payload still goes out.
     marker = await _preserve_fields(node, 'data', media_store, threshold_bytes)
     marker[_EXTERNAL_MARKER] = True
```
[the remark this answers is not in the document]
dsfaccini: This reply was posted by Claude Code using claude-opus-5 on behalf of David.

Confirmed, and it is one key wider than reported. Reproduced against `c3c11c0`:

```
__harness_external_field__: in='user-value' out=<LOST>
__harness_external_uri__:   in='user-uri'   out=<LOST>
__harness_external_text__:  in='user-text'  out='user-text'
```

`_write_reference` overwrites `_URI_KEY` unconditionally too, so it has the same hole. `_TEXT_MARKER` is already correct -- this PR changed `_restore_external` to drop it only on a text marker, for exactly this reason.

Tracked separately in #541 rather than fixed here. The fix you suggest -- preserving the source value under an escaped key, as `_write_reference` does for the plain `uri` -- means choosing an escape scheme, which changes the on-disk marker format and interacts with the fallback branch `_marker_kind` already carries for markers written before `_FIELD_KEY` existed. That is a format generation worth designing rather than appending to this PR, which is a detection fix.

Reachability is narrow (a tool returning a `kind: 'binary'` mapping with >= 64 KB base64 `data` that also uses the harness's private `__harness_external_*` keys) and it sits inside the residual collision the `_marker_kind` docstring and both docs pages already declare open. It is silent, though, which #541 records.

PR #520, opened 2026-07-31
fix(media): decode URL-safe binary payloads

### LINKED ISSUE #511: `externalize_media` decodes `BinaryContent` with standard base64, but pydantic-ai serializes URL-safe — crashes on most real binary, silently corrupts some
pydantic-ai serializes `BinaryContent.data` as **URL-safe** base64 (`ser_json_bytes='base64'` on the message config; pydantic-core's `'base64'` mode emits the `-`/`_` alphabet, with padding). `externalize_media` decodes it with **standard** `base64.b64decode` (`media/_walker.py`, in `_maybe_externalize_binary`).

The two alphabets only differ in characters 62/63, so low-entropy payloads encode identically and everything looks fine. Real binary (an image, audio) almost always produces `-` or `_`, and then:

- **Usually it crashes**: `b64decode` discards the URL-safe characters, the remaining length isn't a multiple of 4 → `binascii.Error: Incorrect padding`, failing the snapshot save.
- **Sometimes it silently corrupts**: if the number of discarded characters happens to be a multiple of 4, `b64decode` succeeds and returns *different bytes*, which get stored as the blob. `restore_media` then re-inlines the wrong bytes.

This affects every store the same way (File/Sqlite/Mongo, and any custom store) since they all share the walker.

## Reproduction

```python
import asyncio, json, tempfile

from pydantic_ai import ModelMessagesTypeAdapter
from pydantic_ai.messages import BinaryContent, ModelRequest, UserPromptPart
from pydantic_ai_harness.media import DiskMediaStore, externalize_media


async def main():
    messages = [
        ModelRequest(parts=[UserPromptPart(content=[
            'look at this',
            BinaryContent(data=bytes(range(256)) * 500, media_type='application/octet-stream'),
        ])])
    ]
    node = json.loads(ModelMessagesTypeAdapter.dump_json(messages))
    store = DiskMediaStore(directory=tempfile.mkdtemp())
    await externalize_media(node, media_store=store, threshold_bytes=64 * 1024)


asyncio.run(main())
# binascii.Error: Incorrect padding
```

The silent-corruption case, without the walker:

```python
import base64

payload = b'\x94\x8f\x8a\x85\x80{vqlgb]XSNID?:50+&!\x1c\x17\x12\r\x08\x03\xfe\xf9\xf4\xef\xea'
encoded = base64.urlsafe_b64encode(payload).decode()  # what pydantic-ai emits
decoded = base64.b64decode(encoded)                   # what the walker does
assert decoded != payload                             # 32 wrong bytes, no error
```

## Failing test

```python
async def test_binary_externalization_round_trips_high_entropy_bytes(tmp_path):
    """Real binary produces URL-safe base64 chars; the walker must accept them."""
    payload = bytes(range(256)) * 500  # covers all byte values -> hits chars 62/63

    messages = [
        ModelRequest(parts=[UserPromptPart(content=[
            'look at this',
            BinaryContent(data=payload, media_type='application/octet-stream'),
        ])])
    ]
    node = json.loads(ModelMessagesTypeAdapter.dump_json(messages))
    store = DiskMediaStore(directory=tmp_path)

    lean = await externalize_media(node, media_store=store, threshold_bytes=64 * 1024)
    full = await restore_media(lean, media_store=store)

    restored = ModelMessagesTypeAdapter.validate_python(full)
    binary = restored[0].parts[0].content[1]
    assert binary.data == payload
```

## Why the existing tests pass

Every binary fixture in the suite happens to be alphabet-neutral — its base64 never contains a character 62/63, so both alphabets agree: `b'\x89PNG\r\n' + b'\x00' * (70 * 1024)`, `b'\xab' * 100_000`, `b'\x89PNG fake'`, `b'\x00'`. A `bytes(range(256))`-style fixture would have caught this.

## Suggested fix

In `_maybe_externalize_binary`, decode with `base64.urlsafe_b64decode(data_value)` — it translates `-`/`_` and leaves `+`/`/` untouched, so it accepts both alphabets. The restore side needs no change: `restore_media` re-encodes with standard `b64encode`, and pydantic-ai's `val_json_bytes='base64'` validates that fine (verified round-trip).

## Versions

pydantic-ai-harness 0.14.0, pydantic-ai 2.21.0, Python 3.14.

dk3yyyy:
## Summary

Pydantic serializes `BinaryContent.data` with the URL-safe base64 alphabet. The media walker decoded that value with the standard base64 decoder, which could raise `Incorrect padding` or return corrupted bytes when the payload contained `-` or `_`.

This change uses `base64.urlsafe_b64decode`, which accepts Pydantic's URL-safe encoding while remaining compatible with standard base64. The regression test serializes high-entropy `BinaryContent` through Pydantic, externalizes and restores it through the public media API, validates the restored messages, and compares the recovered bytes.

## Linked Issue

Fixes #511

## Verification

- Focused regression before the fix: failed with `binascii.Error: Incorrect padding`
- Focused regression after the fix: `1 passed`
- Media module: `82 passed, 1 skipped`
- `make lint`: passed
- `make typecheck`: `0 errors, 0 warnings, 0 informations`
- `make test`: `3893 passed, 47 skipped`
- `make testcov`: `3893 passed, 47 skipped`; 100% statement and branch coverage

## Checklist

- [x] Tests added for the corrected behavior

### DIFF
--- a/pydantic_ai_harness/media/_walker.py
+++ b/pydantic_ai_harness/media/_walker.py
@@ -136,7 +136,7 @@ async def _maybe_externalize_binary(
     # `_is_binary_part` already verified `data` is a string; the cast is safe.
     data_value = node['data']
     assert isinstance(data_value, str)
-    raw = base64.b64decode(data_value)
+    raw = base64.urlsafe_b64decode(data_value)
     if len(raw) < threshold_bytes:
         return None
     media_type_value = node.get('media_type')
--- a/tests/media/test_media.py
+++ b/tests/media/test_media.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import json
 import os
 import sqlite3
 from datetime import datetime, timezone
@@ -10,6 +11,8 @@ from typing import TypeGuard
 
 import pytest
 from httpx import AsyncClient, MockTransport, Request, Response
+from pydantic_ai import ModelMessagesTypeAdapter
+from pydantic_ai.messages import BinaryContent, ModelMessage, ModelRequest, UserPromptPart
 
 from pydantic_ai_harness.media import (
     DiskMediaStore,
@@ -261,6 +264,28 @@ class TestExternalizeRestoreWalker:
         restored_text = _json.dumps(restored)
         assert '"kind": "binary"' in restored_text
         assert b64_payload in restored_text  # bytes restored exactly
+
+    async def test_binary_round_trip_accepts_pydantic_urlsafe_base64(self, tmp_path: Path) -> None:
+        payload = bytes(range(256)) * 500
+        messages: list[ModelMessage] = [
+            ModelRequest(
+                parts=[UserPromptPart(content=[BinaryContent(data=payload, media_type='application/octet-stream')])]
+            )
+        ]
+        node: object = json.loads(ModelMessagesTypeAdapter.dump_json(messages))
+        store = DiskMediaStore(tmp_path)
+
+        externalized = await externalize_media(node, media_store=store, threshold_bytes=64 * 1024)
+        restored = await restore_media(externalized, media_store=store)
+        [restored_message] = ModelMessagesTypeAdapter.validate_python(restored)
+        assert isinstance(restored_message, ModelRequest)
+        [restored_part] = restored_message.parts
+        assert isinstance(restored_part, UserPromptPart)
+        assert not isinstance(restored_part.content, str)
+        [restored_binary] = restored_part.content
+        assert isinstance(restored_binary, BinaryContent)
+
+        assert restored_binary.data == payload
 
     async def test_binary_restore_reuses_write_context(self, tmp_path: Path) -> None:
         """A context-dependent storage key is found with the binary media type."""
@@ -561,7 +586,6 @@ class TestExternalizeRestoreWalker:
         import json as _json
 
         from pydantic_ai.messages import (
-            ModelMessage,
             ModelMessagesTypeAdapter,
             ModelRequest,
             TextContent,


dsfaccini (commented):
_This review was posted by Claude Code using claude-opus-4-8 on behalf of David._ Validated by David's coding agent (reproduced the bug + verified the fix and round-trip symmetry against current `main`); David directed this but has not personally read the diff line-by-line.

Independently validated #511 on current `main` and confirm this fix is correct and complete:

- **Both failure modes reproduced.** The crash — `binascii.Error: Incorrect padding` on `bytes(range(256)) * 500` (128 KB, over the 64 KB threshold) — and the silent corruption — `base64.b64decode` returns 248 bytes vs. the correct 256 from `urlsafe_b64decode` on a sub-threshold payload, no error raised.
- **Fix is correct.** `pydantic_ai` serializes `BinaryContent.data` URL-safe (`ser_json_bytes='base64'` → pydantic-core's `-`/`_` alphabet); the walker decoded it standard at `_walker.py:139`. `urlsafe_b64decode` is a strict superset (translates `-`/`_`→`+`/`/`, leaves `+`/`/` untouched), so it accepts both alphabets with no downside.
- **Restore side correctly needs no change.** `_restore_external` re-encodes with standard base64, and pydantic-core's base64 *validation* accepts both alphabets, so the round-trip validates.
- **Scope is right.** Line 139 is the only buggy base64 site in the media module; the restore re-encode is correct. The regression test pins exactly the alphabet-hitting payload and fails before / passes after.

CI green, mergeable. Thanks @dk3yyyy for the fix and @voberoi for the report + MRE.

voberoi:
Thanks! This looks right.

PR #535, opened 2026-08-04
fix: propagate usage limits to nested capability runs

### LINKED ISSUE #528: Nested agent runs in `compaction` and `tool_output_limits` ignore the parent's `usage_limits`

Two merged capabilities start a nested `Agent` run from inside a hook and thread the parent's usage onto it, but not the parent's usage limits:

- `pydantic_ai_harness/compaction/_summarizing_compaction.py` -- `SummarizingCompaction._summarize`
- `pydantic_ai_harness/tool_output_limits/_capability.py` -- the `summarize` band action

Both call `agent.run(prompt, usage=ctx.usage)`. Two consequences follow.

**The nested run silently applies pydantic-ai's default `request_limit=50`.** `Agent.run` resolves `usage_limits or UsageLimits()`, and that default is then checked against the *shared* parent usage object. A run configured with `UsageLimits(request_limit=200)` therefore stops summarizing from its 50th request onward.

**The nested run can overshoot the parent's `request_limit` by one.** Both hooks run after `ModelRequestNode._prepare_request` called `check_before_request`, so the parent request is already approved when the nested run spends a slot.

Fix, as applied to `LLMReminder` in #414:

```python
def _reserved_limits(limits: UsageLimits | None) -> UsageLimits | None:
    if limits is None or limits.request_limit is None:
        return limits
    return replace(limits, request_limit=max(0, limits.request_limit - 1))
```

Reserving the slot is what enforces the ceiling: the nested run raises `UsageLimitExceeded` before the parent's approved request goes out, and each capability degrades in its own documented way instead of the run exceeding its budget.

Found while triaging [this review comment](https://github.com/pydantic/pydantic-ai-harness/pull/414#discussion_r3693781152) on #414.

Solaris-star:
## Summary

Nested model runs started by `SummarizingCompaction` and the `ToolOutputLimits` summarize action already share the parent `RunUsage`, but they did not receive the parent's `UsageLimits`. Pydantic AI therefore applied its default `request_limit=50` to the nested run.

This could stop a configured run early or allow the already-approved parent request to overshoot its request budget by one.

## Changes

- Share the one-request reservation logic used by `LLMReminder` in a small internal usage helper.
- Pass the reserved parent limits to compaction and tool-output summarization runs.
- Add regression coverage for a parent `request_limit=5` being forwarded as `4`.

## Verification

- `make lint`
- `make typecheck`
- `make test` -- 4282 passed, 47 skipped
- `make testcov` -- 100% branch coverage

Fixes #528

### DIFF
--- a/pydantic_ai_harness/_usage.py
+++ b/pydantic_ai_harness/_usage.py
@@ -1,0 +1,19 @@
+"""Shared helpers for nested agent usage accounting."""
+
+from __future__ import annotations
+
+from dataclasses import replace
+
+from pydantic_ai.usage import UsageLimits
+
+
+def reserved_usage_limits(limits: UsageLimits | None) -> UsageLimits | None:
+    """Hold one request back for a nested model call made from a hook.
+
+    A hook can run after the parent request has passed its pre-request limit
+    check. Reserving one slot prevents the nested call from allowing that
+    already-approved parent request to exceed ``request_limit``.
+    """
+    if limits is None or limits.request_limit is None:
+        return limits
+    return replace(limits, request_limit=max(0, limits.request_limit - 1))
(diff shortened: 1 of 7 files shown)


mpfaffenberger:
@Solaris-star Would you by any chance want to reopen this?

PR #543, opened 2026-08-05
fix(step-persistence): expose explicit AgentSpec fields

### LINKED ISSUE #537: `StepPersistence.from_spec` publishes no fields to the AgentSpec schema

`StepPersistence.from_spec` (`pydantic_ai_harness/step_persistence/_capability.py`) is declared as:

```python
@classmethod
def from_spec(cls, *args: Any, **kwargs: Any) -> StepPersistence[Any]: ...
```

Core reads `from_spec`'s signature to build the capability's JSON schema
(`pydantic_ai/agent/spec.py` `_build_capability_schema_types._get_schema_target`, then
`pydantic_ai/_spec.py` `build_schema_types`), and `build_schema_types` explicitly drops
`VAR_POSITIONAL` and `VAR_KEYWORD` parameters. With both dropped there are no type hints left, so
only the bare-name branch fires and the generated schema is:

```json
[{"const": "StepPersistence", "type": "string"}]
```

No `$defs` entry, no fields. A spec that configures `StepPersistence` still *loads* correctly at
runtime, but it fails validation against the schema the `# yaml-language-server: $schema=` line
points at -- so every configured block is a red squiggle in the editor and fails any CI that
validates specs.

Found while fixing the identical defect in `SpendLimits` during the #474 post-merge review. The fix
shape there:

- name every spec-expressible parameter explicitly, keyword-only;
- re-declare the `AbstractCapability` fields (`id`, `description`, `defer_loading`), which currently
  reach `cls()` through `**kwargs` and would otherwise be a silent regression;
- keep a `**unsupported: Any` catch-all so runtime-only fields keep their specific `UserError`
  (core drops it from the schema, so it costs nothing there);
- use a `TypedDict` for any nested structure whose dataclass has a callable field, since
  `filter_serializable_type` only strips callables at the top level of a union;
- add a schema test, following `tests/stackone/test_capability.py`
  `test_agent_spec_schema_excludes_runtime_client_and_accepts_scalar_actions`.

`StepPersistence` has its own field surface to get right, so it is not a mechanical copy -- hence a
separate issue rather than widening the #474 review PR.

The other five capabilities with `from_spec` (`Planning`, `ExaSearch`, `ExaAgent`, `Memory`,
`StackOne`) already use explicit signatures and are unaffected.

Solaris-star:
## Summary

`StepPersistence.from_spec` previously accepted only `*args, **kwargs`. Pydantic AI drops variadic parameters when building an `AgentSpec` schema, so configured `StepPersistence` capabilities appeared as a bare string with no configurable fields in YAML/JSON schema validation.

## Changes

- Expose the spec-supported fields with an explicit keyword-only signature.
- Preserve the `AbstractCapability` fields (`id`, `description`, `defer_loading`).
- Keep a runtime-only `**unsupported` catch-all with a clear `UserError`.
- Preserve the memory, file, and sqlite backend behavior.
- Add schema-generation, AgentSpec loading, unknown-field, and backend regression coverage.

## Verification

- `make lint`
- `make typecheck`
- `make test` — 4480 passed, 47 skipped
- `make testcov` — 100% coverage

Fixes #537

### DIFF
--- a/pydantic_ai_harness/step_persistence/_capability.py
+++ b/pydantic_ai_harness/step_persistence/_capability.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 from dataclasses import dataclass, field, replace
 from datetime import datetime, timezone
+from pathlib import Path
 from typing import Any, Literal
 from uuid import uuid4
 
@@ -133,8 +134,8 @@ class StepPersistence(AbstractCapability[AgentDepsT]):
         cls,
         *,
         backend: Literal['memory', 'file', 'sqlite'] = 'memory',
-        directory: str = '.step-persistence',
-        database: str = '.step-persistence.db',
+        directory: str | Path | None = None,
+        database: str | Path | None = None,
         max_snapshots_per_run: int | None = None,
         agent_name: str | None = None,
         run_id: str | None = None,
@@ -145,37 +146,53 @@ class StepPersistence(AbstractCapability[AgentDepsT]):
         defer_loading: bool = False,
         **unsupported: Any,
     ) -> StepPersistence[Any]:
-        """Construct from a serialised spec.
+        """Construct a spec-serializable capability.
 
-        Supports `backend='memory'` (default), `backend='file'` (with
-        `directory`), or `backend='sqlite'` (with `database`). Raises
-        `ValueError` for any other `backend` value -- silently falling
-        back to in-memory storage would turn a typo into accidental
-        non-durability.
+        The explicit keyword-only signature is intentional: Pydantic AI uses
+        it to publish the capability fields in `AgentSpec` JSON schema. The
+        three backend-construction knobs are `backend`, `directory`,
+        `database`, and `max_snapshots_per_run`; the remaining arguments are
+        `StepPersistence`'s serializable base fields.
 
-        `max_snapshots_per_run` (default `None`, unbounded) is forwarded to
-        the constructed store to bound per-run snapshot growth.
+        `store` and other runtime-only values are not accepted here because a
+        live store cannot be represented in YAML or JSON. Construct the
+        capability directly in Python when a custom store is needed.
         """
+        if 'store' in unsupported:
+            raise UserError(
+                '`store` is runtime-only and cannot be loaded from an AgentSpec. '
+                'Construct StepPersistence(store=...) in Python instead.'
+            )
         if unsupported:
             raise UserError(f'StepPersistence has no spec field(s) {sorted(unsupported)}.')
+        if backend != 'file' and directory is not None:
+            raise ValueError('directory is only valid with backend="file"')
+        if backend != 'sqlite' and database is not None:
+            raise ValueError('database is only valid with backend="sqlite"')
         if backend == 'memory':
             store: StepStore = InMemoryStepStore(max_snapshots_per_run=max_snapshots_per_run)
         elif backend == 'file':
             from pydantic_ai_harness.step_persistence._store import FileStepStore
 
-            store = FileStepStore(directory, max_snapshots_per_run=max_snapshots_per_run)
+            store = FileStepStore(
+                directory or '.step-persistence',
+                max_snapshots_per_run=max_snapshots_per_run,
+            )
         elif backend == 'sqlite':
             from pydantic_ai_harness.step_persistence._store import SqliteStepStore
 
-            store = SqliteStepStore(database=database, max_snapshots_per_run=max_snapshots_per_run)
-        else:  # pragma: no cover - Literal validation rejects this through the schema
+            store = SqliteStepStore(
+                database=database or '.step-persistence.db',
+                max_snapshots_per_run=max_snapshots_per_run,
+            )
+        else:
             raise ValueError(f'unknown backend {backend!r}; expected `memory`, `file`, or `sqlite`')
         return cls(
             store=store,
             agent_name=agent_name,
             run_id=run_id,
             parent_run_id=parent_run_id,
-            metadata=metadata or {},
+            metadata=metadata if metadata is not None else {},
             id=id,
             description=description,
             defer_loading=defer_loading,
--- a/pydantic_ai_harness/step_persistence/_capability.py
+++ b/pydantic_ai_harness/step_persistence/_capability.py
@@ -4,12 +4,13 @@ from __future__ import annotations
 
 from dataclasses import dataclass, field, replace
 from datetime import datetime, timezone
-from typing import Any
+from typing import Any, Literal
 from uuid import uuid4
 
 from pydantic_ai import CallToolsNode, ModelRequestNode
 from pydantic_ai.capabilities import AbstractCapability
 from pydantic_ai.capabilities.abstract import AgentNode, NodeResult, WrapRunHandler
+from pydantic_ai.exceptions import UserError
 from pydantic_ai.messages import ModelMessage, ModelResponse, ToolCallPart
 from pydantic_ai.models import ModelRequestContext
 from pydantic_ai.run import AgentRunResult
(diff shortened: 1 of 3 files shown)


dsfaccini (changes requested):
This review was posted by Claude Code using claude-opus-5 on behalf of David. David directed the review and approved posting it; he has not gone through the diff line by line.

Thanks for picking this up so quickly, and for going past the diff -- the schema test asserting `additionalProperties`, the field set and the `backend` enum is stronger than the StackOne prior art the issue pointed at, and `test_agent_spec_loads_step_persistence` exercising a real `Agent.from_spec` is exactly right.

Verified empirically rather than from the diff: on `main` the `$defs` entry for `StepPersistence` is absent; on `fdfe40a` `spec_params_StepPersistence` carries all 11 fields with `additionalProperties: false`, and a realistic configured block validates while `backend: disk` and an unknown key are both rejected. `make lint`, `make typecheck` and `make testcov` all pass on the branch with the 100%-coverage bar met. So the defect in #537 is genuinely fixed.

Four things before merge.

## 1. A mismatched backend option is now silently dropped (`step_persistence/_capability.py`, `StepPersistence.from_spec`)

This is a regression the PR introduces, not a pre-existing wart. On `main`, `from_spec(backend='memory', directory='/durable')` raised `TypeError` because the leftover kwarg reached `cls()`. On this branch it returns an `InMemoryStepStore` and drops `/durable`. Same for `backend='file', database='runs.db'`. Confirmed end to end: `Agent.from_spec({'capabilities': [{'StepPersistence': {'backend': 'memory', 'directory': '/durable'}}]}, ...)` now succeeds with a non-durable store.

The schema makes it worse rather than better -- `spec_params_StepPersistence` has no conditional, so an editor now green-lights a spec that silently does not persist. That undercuts the reason #537 exists.

Both sibling capabilities with this exact `backend`/`directory`/`database` shape already guard, and neither is cited: `memory/_capability.py` `Memory.from_spec` (`directory is only valid with backend="file"`, plus the `database` twin) and `planning/_capability.py` `Planning.from_spec` (the `database` guard). Macroscope flagged the same thing independently.

One subtlety: with `directory: str = '.step-persistence'` you cannot tell "user passed the default" from "not passed". `Memory` accepts that. Cleaner here would be `directory: str | None = None` / `database: str | None = None`, resolving the default inside the matching branch -- the guard becomes exact and it reads the way `max_snapshots_per_run` already does.

## 2. The `# pragma: no cover` comment states something untrue (same method)

It reads `# pragma: no cover - Literal validation rejects this through the schema`. Nothing validates a spec against the JSON schema before `from_spec` runs -- `Agent.from_spec({'capabilities': [{'StepPersistence': {'backend': 'disk'}}]}, ...)` reaches the branch and surfaces the `ValueError`. `Planning.from_spec` carries the opposite comment in-tree and is correct: "`backend` is a `Literal`, but this is the deserialization entry point: the value arrives from a YAML or JSON spec, where nothing has checked it."

The branch is already covered by your own `test_unknown_backend_raises`, so the pragma is not needed at all. Please drop both. (`Memory`'s equivalent `else` has no pragma.)

## 3. Remove the `cast` in the test (`tests/step_persistence/test_step_persistence.py`, `TestFromSpecBackendValidation::test_unknown_backend_raises`)

`CLAUDE.md` "Coding standards" and `agent_docs/capability-authoring.md` "API Design" both rule out casts. Two in-repo idioms exist for this exact test and neither uses one: `tests/planning/test_planning.py` uses `# type: ignore[arg-type]`, and `tests/memory/test_memory.py` drives it through `Agent.from_spec` with no annotation escape at all. The `Agent.from_spec` form is better here -- it exercises the real deserialization path, and it independently disproves the premise behind item 2.

## 4. Record *why* the parameters are named (`from_spec` docstring)

The docstring is unchanged from the variadic version, so it documents neither the new base fields, nor `**unsupported`, nor the reason the signature is explicit. `spend/_capability.py` `SpendLimits.from_spec` -- the reference fix for the same defect -- spends a paragraph on it. Without that line the next tidy-up reverts this fix, which matters more than usual given the defect has now recurred across two capabilities and different authors.

## Smaller things

- **`store` deserves a specific message.** `**unsupported` correctly raises a `UserError`, but `store` -- the field a user is most likely to try -- gets `has no spec field(s) ['store']`, which reads as "no such field" when the field exists and is merely runtime-only. `SpendLimits.from_spec` intercepts its runtime-only set first and says so ("these take callables or a live store. Construct the capability in code to use them."). Worth mirroring.
- **Undisclosed narrowing.** `directory` went from `str | Path` to `str`, and two existing tests were rewritten to `str(tmp_path)` to accommodate it. Defensible (`Memory.from_spec` uses `str`), but it is a real signature change for Python callers and should be a stated decision rather than a test edit. `str | Path` also works -- pydantic renders `Path` as `{"type": "string", "format": "path"}`.
- **Test placement.** The new tests landed mid-run inside `TestStepPersistenceCapability`, splitting `test_from_spec_file_backend` from `test_from_spec_file_backend_default_directory`. `TestFromSpecBackendValidation` is the obvious home.
- **Import.** `from pydantic_ai.capabilities.combined import CombinedCapability` reaches into a submodule; it is exported from `pydantic_ai.capabilities`.
- **Sentinel intent.** `metadata or {}` works, but `metadata if metadata is not None else {}` states it.

## Field surface: complete, and no docs change needed

Checked every constructor field against `dataclasses.fields()`. All 8 are accounted for plus the 3 store-construction knobs, `store` is correctly runtime-only, and nothing is silently dropped. There is no nested dataclass with a callable field here, so the `filter_serializable_type` trap #537 warns about is not in play.

Two store knobs stay unexposed and were not exposed before either, so neither is a regression: `media_store` (a Protocol, correctly runtime-only) and `media_threshold_bytes` (an `int`, spec-expressible, and documented as user-facing in `docs/step-persistence.md`). We will open a follow-up for `media_threshold_bytes` -- not something to widen this PR with.

Docs parity is satisfied as-is: neither `docs/step-persistence.md` nor the capability README carries a YAML spec block, and their single `from_spec` mention stays true. If item 1 lands, the `directory`/`database` restriction is worth a line in both -- `docs/planning.md` and `docs/spend.md` are the two house styles.

## One thing this PR should not carry

While reviewing, we swept every capability in the repo for the same defect class. It is not two capabilities, it is **9 of 33**, across four distinct root causes -- variadic `from_spec` is only the first. A `TYPE_CHECKING`-only name in an `__init__` annotation makes core fall back to the variadic default silently; a field type with no JSON schema deletes the entire entry rather than raising; and `ToolOutputLimits` fails schema generation outright, so any spec including it cannot be validated at all.

None of that is yours to fix. We are landing a repo-wide guard and per-cause issues separately, and `agent_docs/` will finally document the rule -- it currently mentions `from_spec` zero times, which is why this recurred across different authors in the first place. Flagging it only so the scope of this PR stays exactly where it is.

Solaris-star:
Addressed the requested review points in `b324866`.

- `directory` and `database` are now `str | Path | None`; defaults resolve only in their matching backend branch.
- Mismatched backend options raise explicit `ValueError` instead of being silently dropped.
- Runtime-only `store` gets a dedicated `UserError` explaining that it must be constructed in Python.
- Removed the inaccurate coverage pragma and exercised the invalid-backend path through `Agent.from_spec`.
- Removed the test cast and imported `CombinedCapability` from the public `pydantic_ai.capabilities` namespace.
- Moved schema/spec-loading tests into `TestFromSpecBackendValidation`.
- Expanded the `from_spec` docstring to explain the explicit signature and runtime-only fields.
- Used an explicit `metadata is not None` check.

Verification after the update:

- `make lint`
- `make typecheck`
- `make test` — 4482 passed, 47 skipped
- `make testcov` — 100% coverage

Solaris-star:
@dsfaccini All requested changes are now on `b324866`: mismatched backend options raise, runtime-only `store` has a dedicated error, the inaccurate coverage pragma and test cast are gone, and the signature/docstring/tests are updated. The full suite and 100% coverage pass. Please re-review when convenient.

Issue #571, opened 2026-08-07
Lowest-version CI fails when StackOne initializes FastMCP


dsfaccini:
> This issue was posted by Codex Desktop using gpt-5.6-terra on behalf of David.

The lowest-version CI jobs now fail before StackOne tests execute. Constructing `FastMCP('stackone-fake')` raises `pydantic_settings.exceptions.IncompleteFieldDefinitionWarning` for its unresolved `lifespan` annotation, which CI treats as an error.

The failure is independent of the Planning cache-point change in #570: all five lowest-version jobs fail in StackOne setup while the normal all-extras and slim matrices pass.

Example run: https://github.com/pydantic/pydantic-ai-harness/actions/runs/31222632425/job/93010256991

Decide and pin the compatible FastMCP / pydantic-settings combination, or make the StackOne fixture compatible with the declared lowest versions.

DouweM:
Closed by #573

PR #576, opened 2026-08-09
Resolve `FastMCP` settings before StackOne test fixtures

Neallin-917:
## Summary

Resolve FastMCP's `Settings` model before shared StackOne fixtures construct local servers. With FastMCP 3.4.5 and pydantic-settings 2.15.0, the unresolved `lifespan` annotation becomes an error under the suite warnings policy.

This is test-fixture-only compatibility work. Runtime code, dependency constraints, and the lockfile are unchanged.

## Linked Issue

Fixes #571

## Checklist

- [x] Linked issue exists and is referenced above
- [x] Tests updated for the compatibility behavior
- [x] `make lint && make typecheck && make test` passes locally
- [x] `make testcov` passes locally with 100% coverage
- [x] `pyproject.toml` and `uv.lock` are unchanged
- [x] Docstrings use single backticks (not RST double backticks)

## AI assistance

Prepared with AI assistance. The reproduction, diff, and validation reported above were verified in the checked-out repository.

PR #579, opened 2026-08-10
Recognize `ToolAvailabilityDeltaPart` in the compaction estimator and conversation-search index

### LINKED ISSUE #577: `SummarizingCompaction` crashes on `ToolAvailabilityDeltaPart` after a deferred capability loads
# Title

`SummarizingCompaction` crashes on `ToolAvailabilityDeltaPart` after a deferred capability loads

# Capability

Compaction

# Bug Description

Combining token-triggered compaction with an on-demand capability fails after the capability loads. Pydantic AI records the newly available tools in a `ToolAvailabilityDeltaPart`; on the next request, Harness's shared token estimator falls through to `str(part.content)`, but that part has `tools_added` and no `content` attribute.

The failure happens while checking the compaction threshold, even when the history is far below it. I expected a supported Pydantic AI request part to be accepted by the estimator and the run to continue.

This reproduces with the latest releases (`pydantic-ai-harness==0.18.0`, `pydantic-ai==2.27.0`) and against Harness `main` at `5e180850511dec469cc50aa9853675a8031d1f19`. It also reproduces with Pydantic AI 2.23.0 through 2.27.0. `ToolAvailabilityDeltaPart` was not present in 2.22.0.

The interaction appears to be within the documented contracts:

- Harness says its estimator counts every part sent: https://github.com/pydantic/pydantic-ai-harness/blob/v0.18.0/docs/compaction.md#what-counts-toward-the-fraction
- Pydantic AI documents `ToolAvailabilityDeltaPart` as the persisted representation of function-tool reveals: https://github.com/pydantic/pydantic-ai/blob/v2.27.0/docs/capabilities/on-demand.md#resumable-across-runs

Relevant failure:

```text
File "pydantic_ai_harness/compaction/_shared.py", line 80, in _collect_text
    segments.append(str(part.content))
                        ^^^^^^^^^^^^
AttributeError: 'ToolAvailabilityDeltaPart' object has no attribute 'content'
```

Making the capability eager (`defer_loading=False`) avoids the exception, but loses on-demand loading.

# Minimal Reproduction

```python
from pydantic_ai.messages import ModelRequest, ToolAvailabilityDeltaPart
from pydantic_ai_harness.compaction import estimate_token_count

messages = [
    ModelRequest(
        parts=[ToolAvailabilityDeltaPart(tools_added=['demo_tool'])],
    )
]
print(estimate_token_count(messages))
```

# pydantic-ai-harness version

0.18.0

# pydantic-ai version

2.27.0

### LINKED ISSUE #578: `test on latest pydantic-ai` fails on every PR since pydantic-ai 2.26.0
Two coupled problems:

1. pydantic-ai 2.26.0 (released 2026-08-07) changed how an unrevealed deferred tool reaches `TestModel`: it is now sent in `function_tools` flagged `defer_loading=True` instead of being dropped behind a synthetic `search_tools`. `tests/code_mode/test_code_mode.py::TestCodeMode::test_deferred_capability_loader_stays_native_with_tools_all` asserts the old form, so the `test on latest pydantic-ai` job is red on every PR since then. CodeMode itself handles both forms; only the assertion is stale. Bisected: 2.25.0 passes, 2.26.0 fails. First observed on PR #576 (run 31363985519), which is unrelated.

2. `uv.lock` pins pydantic-ai-slim/pydantic-graph 2.22.0, five releases behind. 2.27 adds `ToolAvailabilityDeltaPart` to `ModelRequestPart`, which two harness call sites (compaction token estimator, conversation-search indexer) do not handle, so a plain lock refresh fails `make typecheck`.

Fix: update the stale assertion to accept both wire forms, bump the lock to 2.27.0, and skip the bookkeeping part at both call sites. The `>=2.22.0` floor stays; `test-floor` keeps it honest.

adtyavrdhn:
## Summary

Combining compaction with an on-demand capability crashes: pydantic-ai records newly available tools in a `ToolAvailabilityDeltaPart`, and that is the one `ModelRequestPart` carrying no `content`, so Harness code that walks request parts raises `AttributeError` rather than quietly ignoring it. Reported in #577 with a two-line reproduction.

Two call sites are affected, each with its own crash:

- `compaction/_shared.py` — the token estimator's `else` branch does `str(part.content)` → `AttributeError: ... has no attribute 'content'`. This fires while *checking* the compaction threshold, so it breaks runs whose history is nowhere near the limit.
- `conversation_search/_toolset.py` — the indexer's `RetryPromptPart` fallback reads `part.tool_name` → `AttributeError: ... has no attribute 'tool_name'`. Not in the original report; found by the same audit.

Both now recognise the part and skip it: the names it records travel in the request's tool definitions, which the estimator already leaves out, and tool-list bookkeeping is not conversation content. `_format_message` grew past the complexity cap, so its request branch moved to `_format_request_part` unchanged.

The rest of the part vocabulary was audited and needs nothing: every other type added since 2.22 subclasses one these functions already handle (`ToolSearchCallPart`/`LoadCapabilityCallPart` ⊂ `ToolCallPart`, the search and capability-load returns ⊂ `ToolReturnPart`, `NativeToolSearch*` ⊂ their native bases). `ToolAvailabilityDeltaPart` is the sole exception.

## Exhaustiveness

Skipping one part type would have left the same shape of bug waiting for the next one, so both walks now name every type they handle and end in an `assert_never` guard: a part added upstream fails `make typecheck` instead of a user's run. Verified by simulating an unhandled part — pyright reports `Argument of type "RetryPromptPart" cannot be assigned to parameter "arg" of type "Never"`.

The guard sits under `if TYPE_CHECKING:` rather than in the branch body, which departs from how this is usually written here (`spend/_budget.py`, `guardrails/`). Those guard unions we own, where unreachable is genuinely true. These two walk pydantic-ai's union against whatever release the *user* installed, and the floor has no upper bound — so the usual runtime-raising form would raise on a newer release, which is exactly the failure #577 reported. Static tripwire, no runtime cliff.

The explicit `ToolAvailabilityDeltaPart` branch is redundant against today's union and carries a narrow `reportUnnecessaryIsInstance` ignore. Without it the fallthrough is dead code, and no future part would ever reach it.

## Floor and lock

The floor moves `>=2.22.0` → `>=2.23.0`, and `uv.lock` moves `pydantic-ai-slim`/`pydantic-graph` 2.22.0 → 2.27.0.

An earlier revision of this PR left the floor alone on the grounds that nothing needed the newer release at runtime. That was wrong: recognising this part *is* a runtime requirement, because anyone on 2.23.0–2.27.0 who combines compaction with a deferred capability hits the crash regardless of what we lock. 2.23.0 is where `ToolAvailabilityDeltaPart` was introduced (pydantic-ai#6793, first tagged v2.23.0) — so that is the minimum honest floor, and `test-floor` keeps it honest. The floor is not raised to 2.27 merely because that is what we lock.

With the class guaranteed to import, both checks are `isinstance` rather than a `part_kind` string comparison, and the two new tests drop their version guards and inline imports.

## What changed around this PR

#573 landed while this was open and fixed the `code_mode` deferred-tool assertion, so that part is dropped rather than duplicated — main asserts on `revealed_tool_names`, one branchless assertion holding on both sides of the visibility split. That overlap was the entire cause of the merge conflict.

#573 also silenced pydantic-settings 2.15's `IncompleteFieldDefinitionWarning` with a message-matched `filterwarnings` entry. **#576 (@Neallin-917) is the better fix and is now folded in here**, with `bb649d3` removing the suppression: the entry silenced that warning for every model in the suite, ours included, to route around one unresolved forward reference in a dependency, whereas #576 resolves the annotation where it originates and leaves the warning armed everywhere else.

It is sufficient and load-bearing, measured at the resolution that actually reproduces the failure — `lowest-direct`, which pairs pydantic-settings 2.15.0 with mcp 1.26.0 (the lock carries a newer `mcp`, which is why this does not reproduce against the lock):

| at `lowest-direct` | result |
|---|---|
| #576's fix, no suppression | 48 passed |
| neither | 15 errors |

Its commits are cherry-picked with authorship intact (`1d098db` @Neallin-917, `d1b008a` @adtyavrdhn), so **#576 should be closed as landed here**, not as an empty diff.

## Verification

Run at **both ends of the supported range** — the floor (`lowest-direct`, pydantic-ai 2.23.0) and the lock (2.27.0): 4598 passed, 47 skipped, 100% branch coverage; lint and pyright clean; `uv lock --check` consistent.

Both new tests were mutation-checked at the floor: disabling either `isinstance` branch reproduces its own original `AttributeError` (`.content` in the estimator, `.tool_name` in the indexer).

Testing at the floor was not incidental — an earlier revision passed at the lock and failed all five floor jobs, because `ToolAvailabilityDeltaPart`'s field was named `added` until #7104 renamed it to `tools_added` in v2.26.0. The tests now build the part with no arguments, which is valid from 2.23.0 onward and exercises the same two failures, since neither call site reads what the part carries.

## Linked Issue

Fixes #577
Fixes #578

## Checklist

- [ ] `pyproject.toml` and `uv.lock` are unchanged, or a maintainer added `dependencies:approved` to the current head

### DIFF
--- a/pydantic_ai_harness/compaction/_shared.py
+++ b/pydantic_ai_harness/compaction/_shared.py
@@ -73,6 +73,11 @@ def _collect_text(messages: Sequence[ModelMessage]) -> list[str]:
                     segments.append(_user_prompt_text_for_counting(part))
                 elif isinstance(part, SystemPromptPart):
                     segments.append(part.content)
+                elif part.part_kind == 'tool-availability-delta':
+                    # Tool-list bookkeeping (pydantic-ai 2.27+): the tool names ride in the
+                    # request's tool definitions, which this estimator already leaves out.
+                    # Matched by `part_kind`, not `isinstance`: the 2.22 floor lacks the class.
+                    pass
                 else:
                     # Everything else a request can carry is a retry prompt or a tool return
                     # of some kind -- the tool-search and capability-load returns subclass
--- a/pydantic_ai_harness/compaction/_shared.py
+++ b/pydantic_ai_harness/compaction/_shared.py
@@ -20,6 +20,7 @@ from pydantic_ai.messages import (
     ModelResponsePart,
     NativeToolCallPart,
     NativeToolReturnPart,
+    RetryPromptPart,
     SystemPromptPart,
     TextContent,
     TextPart,
@@ -30,7 +31,7 @@ from pydantic_ai.messages import (
     UserPromptPart,
 )
 from pydantic_ai.tools import RunContext
-from typing_extensions import Self
+from typing_extensions import Self, assert_never
 
 from pydantic_ai_harness.compaction._context_window import DEFAULT_CONTEXT_WINDOW, resolve_context_window
 from pydantic_ai_harness.compaction._pinning import is_pinned
@@ -74,17 +75,34 @@ def _collect_text(messages: Sequence[ModelMessage]) -> list[str]:
                     segments.append(_user_prompt_text_for_counting(part))
                 elif isinstance(part, SystemPromptPart):
                     segments.append(part.content)
-                elif isinstance(part, ToolAvailabilityDeltaPart):
-                    # Tool-list bookkeeping, not content: the names it records ride in the
-                    # request's tool definitions, which this estimator already leaves out.
-                    # It is also the one request part with no `content`, so the `else` below
-                    # would raise rather than over-count it.
+                elif isinstance(part, (ToolReturnPart, RetryPromptPart)):
+                    # Both are sent in full. The tool-search and capability-load returns
+                    # subclass `ToolReturnPart`, so they arrive here too.
+                    segments.append(str(part.content))
+                # Control bookkeeping rather than message text: it records which tools became
+                # available, and the schemas themselves travel in the request's tool
+                # definitions. Those schemas are not free -- this estimator counts no tool
+                # definitions at all, so revealing tools mid-run costs context it does not see
+                # (tracked separately); skipping the part is not a claim that the reveal was
+                # free. Redundant against the union as it stands today -- every other member is
+                # handled above -- but kept explicit so the `else` stays a real branch at
+                # runtime rather than dead code.
+                elif isinstance(part, ToolAvailabilityDeltaPart):  # pyright: ignore[reportUnnecessaryIsInstance]
                     pass
                 else:
-                    # Everything else a request can carry is a retry prompt or a tool return
-                    # of some kind -- the tool-search and capability-load returns subclass
-                    # `ToolReturnPart` -- and all of them are sent.
-                    segments.append(str(part.content))
+                    # A part pydantic-ai added after this was written. Skip rather than guess at
+                    # its payload: this runs on every request to decide whether to compact, so an
+                    # unrecognised part must not take the run down -- which is exactly what the
+                    # old `str(part.content)` fallback did once a part without `content` existed
+                    # (#577).
+                    #
+                    # `assert_never` sits under `TYPE_CHECKING` rather than in the branch body on
+                    # purpose. It still makes the chain exhaustive at type-check time, so a new
+                    # upstream part fails `make typecheck` and becomes a decision we make; but
+                    # unlike the usual runtime form it cannot re-crash a user who upgrades
+                    # pydantic-ai ahead of us, which is the failure this change exists to remove.
+                    if TYPE_CHECKING:
+                        assert_never(part)
         else:
             for part in msg.parts:
                 if isinstance(part, TextPart):
--- a/pydantic_ai_harness/compaction/_shared.py
+++ b/pydantic_ai_harness/compaction/_shared.py
@@ -24,6 +24,7 @@ from pydantic_ai.messages import (
     TextContent,
     TextPart,
     ThinkingPart,
+    ToolAvailabilityDeltaPart,
     ToolCallPart,
     ToolReturnPart,
     UserPromptPart,
@@ -73,10 +74,11 @@ def _collect_text(messages: Sequence[ModelMessage]) -> list[str]:
                     segments.append(_user_prompt_text_for_counting(part))
                 elif isinstance(part, SystemPromptPart):
                     segments.append(part.content)
-                elif part.part_kind == 'tool-availability-delta':
-                    # Tool-list bookkeeping (pydantic-ai 2.27+): the tool names ride in the
+                elif isinstance(part, ToolAvailabilityDeltaPart):
+                    # Tool-list bookkeeping, not content: the names it records ride in the
                     # request's tool definitions, which this estimator already leaves out.
-                    # Matched by `part_kind`, not `isinstance`: the 2.22 floor lacks the class.
+                    # It is also the one request part with no `content`, so the `else` below
+                    # would raise rather than over-count it.
                     pass
                 else:
                     # Everything else a request can carry is a retry prompt or a tool return
--- a/tests/compaction/test_context_budget.py
+++ b/tests/compaction/test_context_budget.py
@@ -645,12 +645,17 @@ def test_tool_availability_delta_adds_nothing_to_the_estimate():
     It is also the only `ModelRequestPart` without a `content` attribute, so before this
     was handled the estimator raised `AttributeError` on any history in which a deferred
     capability had loaded (#577).
+
+    Built with no arguments deliberately: the estimator rejects the part on its type and
+    never reads the names it carries, and the field holding them was renamed
+    (`added` -> `tools_added`) in pydantic-ai 2.26. Naming it here would pin the test to a
+    release later than the floor the runtime actually needs.
     """
     messages = _history(1)
     first = messages[0]
     assert isinstance(first, ModelRequest)
     augmented: list[ModelMessage] = [
-        ModelRequest(parts=[*first.parts, ToolAvailabilityDeltaPart(tools_added=['secret_tool'])]),
+        ModelRequest(parts=[*first.parts, ToolAvailabilityDeltaPart()]),
         *messages[1:],
     ]
 
--- a/tests/compaction/test_context_budget.py
+++ b/tests/compaction/test_context_budget.py
@@ -7,6 +7,7 @@ from collections.abc import Callable
 from typing import Any
 from unittest.mock import AsyncMock, patch
 
+import pydantic_ai.messages
 import pytest
 from opentelemetry.trace import NoOpTracer, Tracer
 from pydantic_ai import Agent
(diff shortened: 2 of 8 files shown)


pydantic_ai_harness/compaction/_shared.py
```diff
@@ -73,6 +73,11 @@ def _collect_text(messages: Sequence[ModelMessage]) -> list[str]:
                     segments.append(_user_prompt_text_for_counting(part))
                 elif isinstance(part, SystemPromptPart):
                     segments.append(part.content)
+                elif part.part_kind == 'tool-availability-delta':
```
DouweM: Why not isinstance?

DouweM: Related to https://github.com/pydantic/pydantic-ai-harness/issues/577

pydantic_ai_harness/compaction/_shared.py
```diff
@@ -73,6 +73,11 @@ def _collect_text(messages: Sequence[ModelMessage]) -> list[str]:
...
+                    # Tool-list bookkeeping (pydantic-ai 2.27+): the tool names ride in the
+                    # request's tool definitions, which this estimator already leaves out.
+                    # Matched by `part_kind`, not `isinstance`: the 2.22 floor lacks the class.
```
DouweM: Let's bump the floor then :)

DouweM: From PR description:
> The >=2.22.0 floor in pyproject.toml is deliberately untouched: nothing needs 2.27 at runtime, and test-floor keeps the floor honest. 

I'm not sure this is true considering the changes we made recently around tool availability deltas

DouweM:
Took this over — all three of your comments are addressed in `8b337e4`, and the branch is un-conflicted.

**"Why not isinstance?" / "Let's bump the floor then :)"** — done, both. The floor moves to `>=2.23.0` and both checks are `isinstance`.

**"I'm not sure this is true considering the changes we made recently around tool availability deltas"** — you were right, and the original rationale had it backwards. Recognising `ToolAvailabilityDeltaPart` *is* a runtime requirement: anyone on 2.23.0–2.27.0 combining compaction with a deferred capability crashes, whatever we lock. But the correct floor is 2.23.0, not 2.27.0 — that is where the part was introduced (pydantic-ai#6793, `git tag --contains` → first `v2.23.0`, which matches the reporter's "reproduces on 2.23.0 through 2.27.0"). Raising it to 2.27 just because that is what we lock would be stricter than the code actually needs.

Two things worth flagging beyond the review:

**There was a second crash, not in the report.** The conversation-search indexer falls through to a `RetryPromptPart` branch that reads `part.tool_name`, so it raised `AttributeError: ... has no attribute 'tool_name'` on the same part. Same root cause, different call site; both are covered and mutation-checked (disabling either `isinstance` reproduces its own `AttributeError`).

**The rest of the vocabulary is fine.** I enumerated `ModelRequestPart`/`ModelResponsePart` and checked subclassing: every type added since 2.22 subclasses one these functions already handle — `ToolSearchCallPart`/`LoadCapabilityCallPart` ⊂ `ToolCallPart`, the search and capability-load returns ⊂ `ToolReturnPart`, `NativeToolSearch*` ⊂ their native bases. `ToolAvailabilityDeltaPart` is the only `ModelRequestPart` that is *not* a `ToolReturnPart` and has no `content`, which is exactly why it is the only one that broke. So this is the whole bug surface, not the first of a series.

**Dropped from the PR:** #573 landed while this was open and already fixed the `code_mode` assertion and the pydantic-settings warning, so both are removed rather than duplicated — that overlap was the entire cause of the conflict. `Fixes #571` came off the description accordingly, since #573 fixed it.

**One decision left for you:** #576 (@Neallin-917) is no longer needed to make CI green, but it is arguably the *better* fix — it resolves the `mcp` `Settings` forward references at the root, where main now suppresses the warning with a fairly broad `filterwarnings` match. I did try to determine empirically whether the root-cause fix alone suffices, but couldn't reproduce the failure locally: it needs the `lowest-versions` combination of old `mcp` with new pydantic-settings, and layering pydantic-settings 2.15 alone doesn't trigger it. So I left both alone rather than guess. Worth judging #576 on its merits — and if you take it, dropping my suppression with it — rather than closing it as an empty diff.

DouweM:
Settled #576 empirically and folded it in, so this is now the single PR to merge.

You were right that CI has the combination — and so does `--resolution lowest-direct` locally, which is why my earlier attempt failed: layering pydantic-settings 2.15 onto the *lock* keeps the lock's newer `mcp`. The floor resolution pairs pydantic-settings 2.15.0 with mcp 1.26.0, and that reproduces immediately.

| at `lowest-direct` (pydantic-settings 2.15.0 + mcp 1.26.0) | result |
|---|---|
| root-cause fix, no suppression | **48 passed** |
| neither | **15 errors** |

So @Neallin-917's fix is both sufficient and load-bearing — not a no-op that was riding on the suppression. It is in as `1d098db` (Neal) and `d1b008a` (@adtyavrdhn), cherry-picked with authorship intact, and `bb649d3` removes the `filterwarnings` entry #573 added.

That is the better end state: the suppression silenced `IncompleteFieldDefinitionWarning` for *every* model in the suite, ours included, to route around one unresolved forward reference in a dependency. Now the annotation is resolved where it originates and the warning stays armed everywhere else.

**#576 can be closed as landed here** rather than as an empty diff — its commits are in this branch verbatim.

Verified at both ends after the change: floor (pydantic-ai 2.23.0) and lock (2.27.0) suites green — 4598 passed, 47 skipped — 100% branch coverage, lint and pyright clean, `uv lock --check` consistent.

Note the push reset `dependencies:approved` again; the dependency diff is unchanged apart from `pyproject.toml` losing the suppression block.

pydantic_ai_harness/compaction/_shared.py
```diff
@@ -73,6 +74,12 @@ def _collect_text(messages: Sequence[ModelMessage]) -> list[str]:
...
                 elif isinstance(part, SystemPromptPart):
                     segments.append(part.content)
+                elif isinstance(part, ToolAvailabilityDeltaPart):
+                    # Tool-list bookkeeping, not content: the names it records ride in the
+                    # request's tool definitions, which this estimator already leaves out.
+                    # It is also the one request part with no `content`, so the `else` below
+                    # would raise rather than over-count it.
+                    pass
                 else:
                     # Everything else a request can carry is a retry prompt or a tool return
```
DouweM: This is clearly a bad assumption to be making -- let's test those directly. As a general rule, we should consume pydantic-ai in a forward-compatible way.

tests/compaction/test_context_budget.py
```diff
@@ -637,6 +638,30 @@ async def test_one_token_over_the_trigger_compacts(self, monkeypatch: pytest.Mon
         assert len(request_context.messages) < 6
 
 
+def test_tool_availability_delta_adds_nothing_to_the_estimate():
+    """The part is tool-list bookkeeping: its names travel in the request's tool
+    definitions, which the estimator deliberately leaves out of the count.
+
+    It is also the only `ModelRequestPart` without a `content` attribute, so before this
+    was handled the estimator raised `AttributeError` on any history in which a deferred
+    capability had loaded (#577).
```
DouweM, line 650: This section is overconfident -- adding tools with schemas mid-run definitely has a context window cost. But we can track that as an issue instead of fix it now.

Issue #61, opened 2026-03-20
Tool Error Recovery capability


adtyavrdhn:
## Summary

A capability that catches unhandled tool execution errors and returns them as structured error messages to the LLM instead of crashing the agent run, enabling the agent to self-correct.

## Motivation

When a tool throws an unexpected exception, the default behavior crashes the agent loop. A production coding harness needs graceful error handling — the error should be returned to the LLM as context so it can adjust its approach. Both Mastra and Open SWE implement this pattern.

### Open SWE — ToolErrorMiddleware
```python
# agent/middleware/tool_error_handler.py:64-104
class ToolErrorMiddleware(AgentMiddleware):
    """Normalize tool execution errors into predictable payloads.
    
    Catches any exception during a tool call and converts it into
    a ToolMessage with status="error" so the LLM can see the failure
    and self-correct, rather than crashing the agent run.
    """
    async def awrap_tool_call(self, request, handler):
        try:
            return await handler(request)
        except Exception as e:
            data = _to_error_payload(e, request)
            return ToolMessage(
                content=json.dumps(data),
                tool_call_id=_get_tool_call_id(request),
                status="error",
            )
```

### Mastra — Error recovery with retry
```typescript
// packages/core/src/harness/harness.ts:1336-1373
// On "Tool X not found" error:
// 1. Emits retryable error event
// 2. Queues follow-up message telling agent to retry with correct tool name
// Error event includes: retryable: boolean, retryDelay: number
```

## pydantic-ai design sketch

This is best implemented as a hook or middleware on the toolset rather than a separate capability:

```python
class ToolErrorRecovery(AbstractCapability[AgentDepsT]):
    """Catch tool errors and return them to LLM for self-correction."""
    
    max_retries: int = 3
    retry_on: tuple[type[Exception], ...] = (Exception,)
    
    async def after_tool_call(self, ctx, tool_name, result, error=None):
        if error is not None:
            # Return error as tool result instead of raising
            return ToolReturnValue(
                content=f"Error in {tool_name}: {error}. Try again.",
                is_error=True,
            )
```

Note: pydantic-ai PR #4642 already implements tool lifecycle hooks with `ToolValidateError` and `ToolRetryError` short-circuit exceptions. This may already be partially covered.

## Who has this

| Framework | Feature | Details |
|-----------|---------|---------|
| Open SWE | `ToolErrorMiddleware` | Wraps all tool calls, returns errors as ToolMessage |
[...]

PR #581, opened 2026-08-11
Add `ToolErrorRecovery` capability

jondieaix:
Adds a `ToolErrorRecovery` capability that catches errors raised during tool calls and decides what to do with each one, instead of crashing the run.

A `classify` callable maps each error to one of four outcomes: `retry`(transparent, bounded), `inform` (report it to the model as a `ToolFailed` result), `fallback` (return a substitute value), or `propagate` (re-raise). By default, programming errors propagate, connection and timeout errors retry, and everything else is reported to the model. `str(error)` is kept out of the model-facing text by default, so messages can't leak through the model.

This grew out of #171 and started as a standalone package: https://github.com/aixbrain/pydantic-ai-error-recovery. 
Compared to the draft in #171, it uses a `classify` callable instead of a per-tool strategy map, a single `wrap_tool_execute` hook, and keeps `str(error)` out of the model-facing text by default. Addresses #61.

ruff, pyright strict, and the tests pass locally (100% branch coverage).

### DIFF
--- a/pydantic_ai_harness/tool_error_recovery/README.md
+++ b/pydantic_ai_harness/tool_error_recovery/README.md
@@ -1,0 +1,88 @@
+# Tool Error Recovery
+
+Recover from tool-call errors -- retry, report to the model, fall back, or propagate -- instead of crashing the agent run.
+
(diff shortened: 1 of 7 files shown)


jondieaix:
Hi @adtyavrdhn, let me know if there's anything I can do to help move this along. Feel free to ask if you have any questions.

PR #336, opened 2026-07-08
feat(loop_detection): detect and interrupt repeated-action loops

DouweM:
## Summary

Adds `LoopDetection`, a private/experimental capability that detects when an agent is stuck in a repeated-action loop and intervenes. An autonomous agent that gets stuck does not stop -- it repeats the same action until it exhausts its step or token budget. Five major coding harnesses (Gemini CLI, OpenHands, Roo, Crush, goose) each grew some form of loop detection; this is that guardrail as a composable capability. Recommendation 4 of the [broad CLI-harness survey](https://github.com/pydantic/pydantic-ai-notes/blob/main/features/harness-comparison/2026-07-06%20broad%20survey%20-%20cli%20coding%20harnesses.md).

## Linked Issue

Closes #71, supersedes #186.

## What

Four detection tiers, each mapped to a signal the surveyed harnesses use:

| Tier | Signal | Default |
|------|--------|---------|
| Exact repetition | Same `(tool_name, canonical_args)` fingerprint occurs `repeat_threshold` times within a sliding `window` of recent calls. | 5 within 10 |
| Error cycle | Same tool call returns a byte-identical result on `error_cycle_threshold` consecutive executions. | 3 |
| Alternation | Two distinct call fingerprints alternate A-B-A-B for `alternation_cycles` full cycles. | 3 cycles |
| Monologue | `monologue_threshold` consecutive model responses carry no tool call and near-identical text (normalized prefix match). | 3 |

On detection, `on_loop` runs one of:
- `'nudge'` (default): enqueue a harness-marked message the model sees on its next request ("You appear to be repeating the same action ...").
- `'error'`: raise `LoopDetectedError`.
- a sync/async callable receiving a structured `LoopDetected` (tier, tool_name, count, window, fingerprints, message).

## How

- `for_run` returns a fresh copy holding a per-run `_LoopState` (sliding window + counters), so concurrent runs never share state.
- `after_model_request` fingerprints the response's tool calls (tiers 1 and 2b) or its text (tier 2c); `after_tool_execute` tracks results (tier 2a).
- The nudge uses Pydantic AI's pending-message queue (`RunContext.enqueue(..., priority='asap')`), so it is delivered as a real user turn on the next model request (and redirects a would-be-terminating run into one more request). This is the mechanism the repo exposes to capabilities, so no new plumbing.
- Arguments are canonicalized (JSON, sorted keys) before fingerprinting; after a tier fires its counters reset so the same loop must rebuild before firing again.
[...]

PR #591, opened 2026-08-12
Front pages + first harnesses: Coder and Researcher, one capability index across core and Harness

DouweM:
**Draft, iterating with Douwe.** Front-page rewrite + the first **harnesses** (`Coder`, `Researcher`): complete combined capabilities you can run directly, plus their examples. Companion PR: pydantic-ai#7421 (the two repos' front pages tell one story).

## The harnesses

`Coder` and `Researcher` are regular `CombinedCapability`s that bundle a complete, opinionated stack — `Agent('anthropic:claude-fable-5', capabilities=[Coder()])` + `to_cli_sync()` is a working coding agent, and every page shows the blown-out equivalent so it's clear you can take the stack apart. Composition is informed by the coding agents we run in production daily, which is why `Coder` includes a read-only explorer sub-agent and tool-result-clearing compaction rather than a bare sliding window.

Each also exports a lowercase **module-level agent instance**, so CLIs that accept a module path can run them directly, today:

```
uvx --with pydantic-ai-harness clai -a pydantic_ai_harness.coder:coder_agent
```

(The instances are model-less and lazily constructed; `clai` supplies its default model and `-m` overrides — verified. Same mechanism a future `--mode` CLI or the gh-aw engine can point at.)

## Examples

Two flagship examples (`coding_agent`, `research_agent`) on the harnesses, plus `*_from_blocks` variants keeping the explicit compositions runnable — construction-tested with TestModel, no network. The other four agents from #402 (thanks @strawgate — code review, atlas, data analysis, support) intentionally wait for a follow-up: top-level harnesses are the ones we stand behind as directly runnable; broader show-what-you-can-build examples come next.

## Front pages

Per Samuel's 8/10 rules (understandable in 30 seconds, snippet above the fold, one-line comments):

- Quick start: 5-line Python + the one-line CLI incantation.
- "No magic" section: the blown-out equivalent, so composability stays the story.
- **One capability index across core + Harness** with a Ships-in column (Web Search, MCP, Tool Search, Thinking etc. appear alongside Harness capabilities); categories regrouped — Code Mode moved to *Context efficiency* (it changes how the agent executes, not where) with its benefit actually stated; knowledge split from history-rewriting; ACP under an interfaces note. The old ~50-row matrix, the stale docs table (missing 4 shipped capabilities), the "in development" teasers, and the six-community-package showcase are gone.
- `docs/nav.json` grouped to match (parity test now parses nested groups — that was the red CI). @strawgate: does the unified-docs mount render nested nav groups?
- Models: `claude-fable-5` / `gpt-5.6-sol`.

Quality: 4,665 tests passing, 100% branch coverage, pyright strict clean, docs parity + snippet tests green.

## Known follow-ups (deliberate, not blockers)

RepoContext incremental attach-on-first-touch (#594) · loop/tool-error guards (#336/#171 — prioritized by this work) · option flags so pydanty and the gh-aw shim can run on `Coder` with configuration · remaining #402 examples.


DouweM:
Pushed the response to the first review pass (`a9931ac`…`be9973b`):

**Code**
- `FileSystem` gains an intrinsic `read_only=True` mode; the Coder's `explorer` now mounts `FileSystem(read_only=True)` + `RepoContext` as capabilities (no more toolset juggling or hardcoded read-tool list). `READ_ONLY_TOOL_NAMES` is public.
- `instructions` is now a real default value on `Coder`/`Researcher`: pass your own to replace, `instructions=None` to disable entirely.
- `Researcher` uses `WebSearch(local=True)` — native provider search when the model supports it, DuckDuckGo fallback when it doesn't (`native=False` was forfeiting native search for nothing).
- `coder_agent`/`researcher_agent` are named (`name='coder'`/`'researcher'`).
- Top-level importability is now the norm: all public capabilities (and the default constants) lazily export from `pydantic_ai_harness`.
- Shortened the harness class docstrings so they don't enumerate a composition that will drift.

**Tests**
- New recorded end-to-end test: `Coder` fixes a seeded bug in a tmp workspace with `claude-fable-5` live (asserts on the actual file fix + filesystem/shell tool use), cassette replays keyless. Re-record-on-composition-change rule added to the review checklist.

**Examples**
- Blocks-only now: the `Coder`-wrapper example is gone (the `uvx … clai -a` one-liner *is* that), and both remaining examples are fully self-contained — instructions and allowlist written out, no constant imports — with keep-in-sync markers.

**Docs**
- Coder/Researcher pages: cross-links into the pydantic-ai docs (combined capability, capabilities, CLI custom agents, `to_web`), all of run/CLI/web shown, exported agents linked to their API reference, instructions override documented, 'Not included by default' / 'Making it more powerful' sections, blown-out equivalents written out literally, standard stability blockquote.
- examples page/README rewritten for a growing example set with honest `PYDANTIC_AI_MODEL` framing.
- agent_docs: three-way sync contract (implementation ↔ docs blown-out ↔ example).

**Filed from the review**: #606 (workspace arg naming), #607 (`agent_folders` default), #608 (swap/disable components + upgrade warnings + Coder-on-Modal), #609 (Researcher composition: CodeMode value, structured-output fan-out), #610 (grow Coder deliberately).

Open per Douwe: what `DEFAULT_CODER_INSTRUCTIONS` should shrink to (Pi ships ~2 behavior-only bullets and per-tool guidance — which our capabilities already contribute via `get_instructions()`), and whether unified-docs wants `navigation.yml` here like pydantic-ai#7416 (checking with Bill).

DouweM:
CI fix for the new integration test: the repo never depended on pydantic-ai's `anthropic` extra, and `browser-use` pins `anthropic==0.76.0`, so CI resolved an SDK too old for the provider (every job failed on that one test; local venvs masked it). `de3bb3f` adds `pydantic-ai-slim[anthropic]` to the dev group plus an `anthropic>=0.108.0` uv override (resolves to 0.122.0; browser-use tests still pass against it). This is the dependency change riding the `dependencies:approved` gate — pyproject + lock only.

DouweM:
Navigation finding (investigated per Douwe): since [unified-docs#179](https://github.com/pydantic/unified-docs/pull/179) (merged 8/12), production renders the harness sidebar from **pydantic-ai's `docs/navigation.yml`** (`source: "harness"` entries) — this repo's `docs/nav.json` has **zero production consumers** now (pydantic-ai's own nav.json was deleted in pydantic-ai#7416; the cutover is complete).

Consequences handled:
- The new Coder/Researcher/Examples pages + this PR's grouped sidebar structure have been ported to pydantic-ai's `navigation.yml` on the pydantic-ai#7421 branch (commit `e471735`) — same categories as this PR's nav.json, all existing slugs/aliases unchanged, so nothing moves URL. They go live once this PR merges, the harness releases, and #7421 merges — which is already the planned order.
- `agent_docs/docs-conventions.md` now documents the cross-repo contract: nav.json stays as the parity-enforced local ledger, but a page addition/rename isn't live until the companion `navigation.yml` entry lands in pydantic-ai.

Open policy call for @DouweM: keep nav.json as the ledger (status quo, documented), or delete it and retarget the nav half of `test_docs_parity.py`. Kept it for now — deleting is a one-commit follow-up if preferred.

DouweM:
Second response batch pushed (`a0e8811`…`d15cfb5`), executing the follow-up rulings:

**Coder ships with no default instructions** (`1d27e61`). `DEFAULT_CODER_INSTRUCTIONS` is deleted; `instructions=` (default `None`) adds your own. Backed by research: Anthropic removed >80% of Claude Code's system prompt for Claude-5-gen models with no eval loss, deleting exactly the categories our six lines were (step-by-step, read-before-edit, run-tests-after); what they kept is product identity + per-tool guidance — which for us is the user's `Agent` and each capability's own `get_instructions()`. Pi ships two behavior bullets; Code Puppy is persona-first. Researcher's citation instructions stay (they're the product contract, not coaching).
**Empirical check**: the cassette was re-recorded per the new checklist rule, and the instruction-less Coder completed the same seeded-bug task on its first live attempt — 8 model turns vs 7 with instructions (one wrong-filename guess, self-recovered). Keyless replay green.

**Also in this batch**: export-importability test tolerates extra-gated capabilities (fixes slim + 3.10 CI); branchless API-key setup in the integration test (fixes the one-sided branch coverage); top-level-export norm documented in agent_docs + the two stale 'no top-level re-export' notes fixed (DynamicWorkflow, LocalStack); both capability READMEs brought to parity with their reworked docs pages.

Open on @DouweM: (1) instrument-by-default for the packaged agents — recommended together with changing #7447's banner condition from 'not instrumented' to 'no exporter configured', otherwise pre-instrumented agents never show the banner; (2) nav.json keep-as-ledger vs delete (see navigation comment above); (3) optional one-line identity on `coder_agent` only, mirroring Claude Code's surviving identity line.

DouweM:
Composition change per review: **`Researcher` drops `CodeMode` and gains core `WebFetch(local=True)`** (ba89c29 + docs). Rationale: native web search bypasses `run_code` entirely, and with `local=True` the local tool is removed on models that support native search — so CodeMode had nothing to wrap on the models we showcase. Meanwhile the instructions say "read the sources" but the DuckDuckGo fallback only returns snippets — WebFetch closes that hole. New `[researcher]` extra supplies the local search+fetch fallbacks (`pydantic-ai-slim[duckduckgo,web-fetch]`). The DynamicWorkflow fan-out question for Researcher stays open in #609.

Also in this batch (216c50b): review-pass fixes on the front pages — the from-blocks example no longer crashes (`WebSearch(native=False)` without a local tool raises `UserError`), the blown-out snippet now passes `cwd` + `DEFAULT_ALLOWED_COMMANDS` so its 'allowlisted' comment is true (bare `Shell()` is denylist-mode), the Memory snippet imports `Agent` and uses top-level imports, 'sandboxed file access' reworded to workspace-rooted (FileSystem is path-safe, not a sandbox), `google:gemini-3-pro` → `gemini-3-pro-preview` (the bare id isn't a known model), and docs/index.md gains the Composing-from-blocks section so the page shows a run + output.

pydantic_ai_harness/coder/_capability.py
```diff
@@ -0,0 +1,93 @@
...
+
+    It ships with no default instructions: modern models don't need procedural coaching, and the composed capabilities
+    contribute their own tool guidance. Pass `instructions=` to add your own.
+    """
+
+    def __init__(
+        self,
+        workspace: str | Path = '.',
+        *,
+        allowed_commands: Sequence[str] | None = None,
```
DouweM, line 69: We have to make it very clear, here, and in the Coder docs, and in the Shell docs, that this is NOT a security boundary, if the LLM wants to work around us it can.

cc @adtyavrdhn

pydantic_ai_harness/coder/README.md
```diff
@@ -0,0 +1,90 @@
...
+```bash
+uvx --with pydantic-ai-harness clai -a pydantic_ai_harness.coder:coder_agent
+```
+
+`clai` supplies its default model when `-m` is omitted; pass `-m provider:model` to select or override it.
+
+It is literally these capabilities combined, in this order:
+
+- `FileSystem` — read, write, edit, and search tools rooted at the workspace, path-traversal and symlink safe
+- `Shell` — allowlisted commands rooted at the workspace (`DEFAULT_ALLOWED_COMMANDS` is the default allowlist)
```
DouweM: As mentioned above, we should not state this security aspect so confidently. Discuss with me how we should frame the situation.

pydantic_ai_harness/coder/README.md
```diff
@@ -0,0 +1,90 @@
...
+- `RepoContext` — repository instructions and structure
+- `Planning` — a plan the agent creates and keeps current during multi-step work
+- `SubAgents` — delegation, with a read-only `explorer` sub-agent by default
+- `ClearToolResults` — clears stale tool results at 70% of the model context window
+- `WarnNearLimits` — warns the agent at 90% of the model context window
+- `ToolOutputLimits` — bounds how much context any single tool result can consume
```
DouweM: All of these capability mentions obviously need to be links.

pydantic_ai_harness/coder/README.md
```diff
@@ -0,0 +1,90 @@
...
+from pydantic_ai import Agent
+from pydantic_ai_harness import Coder
+
+agent = Agent('anthropic:claude-fable-5', capabilities=[Coder('.')])
+
+result = agent.run_sync('Find out why tests/test_parser.py fails and fix the bug it caught.')
+print(result.output)
+```
+
+The same agent works with every Pydantic AI interface: `agent.to_cli_sync()` starts an interactive chat in your terminal, and `agent.to_web()` serves a browser chat UI.
```
DouweM: All the links you added to the docs version of this, we need here too. Keep them in sync!!!

pydantic_ai_harness/researcher/README.md
```diff
@@ -0,0 +1,67 @@
...
+```
+
+`clai` supplies its default model when `-m` is omitted; pass `-m provider:model` to select or override it.
+
+It is literally these capabilities combined, in this order:
+
+- Concise default research instructions (`DEFAULT_RESEARCHER_INSTRUCTIONS`) — pass `instructions='...'` to replace them, or `instructions=None` to disable
+- Core `WebSearch(local=True)` — the provider's native web search when the model supports it, with a local DuckDuckGo fallback when it doesn't
+- Core `WebFetch(local=True)` — read the pages behind the results, native where supported with a local fallback, so claims can be checked against their sources
+- `ToolOutputLimits` — bounds how much context any single tool result can consume
```
DouweM: Link to all the docs! Also for WebSearch, WebFetch

docs/coder.md
```diff
@@ -0,0 +1,155 @@
...
+print(result.output)
+#> Found it: `parse()` returned None on empty input instead of raising. Fixed in src/parser.py; tests pass now.
+```
+
+The same agent works with every Pydantic AI interface: [`agent.to_cli_sync()`](https://pydantic.dev/docs/ai/cli/) starts an interactive chat in your terminal, and [`agent.to_web()`](https://pydantic.dev/docs/ai/web/) serves a browser chat UI.
+
+Or skip the file entirely and run the exported [`coder_agent`](#api-reference) with [`clai`](https://pydantic.dev/docs/ai/cli/#custom-agents) (the Pydantic AI CLI), via [`uvx`](https://docs.astral.sh/uv/guides/tools/):
+
+```bash
+uvx --with pydantic-ai-harness clai -a pydantic_ai_harness.coder:coder_agent -m anthropic:claude-fable-5
```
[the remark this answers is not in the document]
DouweM: Verified empirically in a clean directory (per Douwe). The diagnosis isn't right: `uvx --with pydantic-ai-harness clai ...` **does** install `clai` as the primary tool package, and `clai==2.30.0` depends on the full `pydantic-ai` package, which bundles the `anthropic` and `cli` extras — resolution installs 102 packages with nothing missing. The command currently fails for a different reason: PyPI's `pydantic-ai-harness==0.20.0` predates the `coder` module, so the agent target doesn't exist yet. Running the same command with the merged (unreleased) harness code via `--with <local checkout>`, `coder_agent` loads and the run proceeds to the expected `ANTHROPIC_API_KEY` prompt. The suggested `--from 'pydantic-ai[cli,anthropic]'` variant would fail identically today for the same reason. Fix is the pending harness release, not a docs change.

PR #604, opened 2026-08-13
Require every capability to decide what OpenTelemetry it emits

dsfaccini:

David specified the rule and chose its placement and strength; the wording is not yet reviewed by him.

## What

Writes down a rule that until now existed only as unwritten practice: a capability decides
what it emits to OpenTelemetry as part of its design, and documents that decision --
including a decision to emit nothing.

## Why

An operator watching a run can only see a capability's own decision points if the
capability reports them: a guard refusing a value, a budget stopping a run, a history
being rewritten. Core's spans cover the model request, the tool call and the agent run,
and nothing else.

Four capabilities already emit spans and settled on a consistent pattern -- `guardrails`
(`guardrail blocked/redacted/deferred`, content behind `trace_include_content`), `spend`
(`spend budget exhausted`), `memory` (`memory.inject`), and `compaction`
(`compact_messages`, with a pluggable tracer defaulting to `NoOpTracer`). None of it was
recorded anywhere an author reads before building, so a new capability could ship blind to
it. Two capabilities also document why they emit nothing (`step_persistence`, and `spend`
for accrual), which is the model this rule generalizes.

## Changes

- `AGENTS.md`: new `Telemetry` section stating the rule and pointing at the pattern.
- `agent_docs/capability-authoring.md`: new `Telemetry` section with the house pattern --
  spans on `ctx.tracer`, `NoOpTracer` fallback outside a `RunContext`, attributes under
  `span.is_recording()`, static low-cardinality span names, capability-prefixed attributes,
  GenAI semantic conventions where one fits, and content behind `trace_include_content`.
- `agent_docs/review-checklist.md`: an Implementation bullet so review gates on it.
- `agent_docs/index.md`: a Task Routing entry.

Documentation only; no runtime change and no new telemetry is added to any capability by
this PR. Retrofitting spans into the capabilities that currently emit none is per-capability
work, not part of this change.

`.agents/capability-philosophy.md` carries the same rule, but it is deployed into worktrees
from `pydantic-ai-tooling/harness-config/` rather than tracked here, so that edit is not in
this diff.


mpfaffenberger (approved):
Verified every factual claim against the tree: the spend README and docs/spend.md do document that accrual emits nothing with the stated reason, the step_persistence README documents its no-span decision citing core's `Instrumentation` coverage, and `compaction/_manual.py` has the `tracer: Tracer | None = None` + `NoOpTracer` seam the house pattern describes. So this codifies real existing practice, and 'undocumented silence is indistinguishable from an oversight' is the right framing for the checklist gate. Docs-only, CI green.

PR #620, opened 2026-08-14
Let `SummarizingCompaction` stream nested summary requests

### LINKED ISSUE #619: SummarizingCompaction nested summary request cannot use stream-only providers
### Capability

Compaction (`SummarizingCompaction`)

### Bug Description

`SummarizingCompaction._summarize()` starts a nested `Agent` and calls `agent.run(prompt, usage=ctx.usage)` without an `event_stream_handler`. Pydantic AI therefore takes the non-streaming request path even when the summary model or provider requires streaming.

Providers that reject `stream=false` fail compaction with HTTP 400 instead of returning a summary. The parent agent's own streaming mode does not help: compaction cannot see the outer `event_stream_handler` from `before_model_request`, and a non-streaming parent can still need a streaming summary request.

Expected: an opt-in on `SummarizingCompaction` that streams only the nested summary request, drains those events internally, and still returns one completed summary to the parent run.

### Minimal Reproduction

```python
from pydantic_ai import Agent
from pydantic_ai_harness.compaction import SummarizingCompaction

# Any summary model/provider that rejects non-streaming requests.
agent = Agent(
    'openai:gpt-4o',
    capabilities=[
        SummarizingCompaction(
            model='provider:stream-only-summary-model',
            max_messages=2,
            keep_messages=1,
        )
    ],
)

# Compaction's nested Agent.run(...) uses the non-streaming path and the
# provider rejects the request.
await agent.run('go', message_history=[...])
```

main (after pydantic-ai 2.28)

### pydantic-ai version

2.28.0

JayGhiya:
> This pull request was posted by Codex Desktop using gpt-5.6-sol on behalf of David.

## Summary

`SummarizingCompaction` made non-streaming nested summary requests, so compaction failed for providers that require streaming.

Add keyword-only `event_stream_handler: EventStreamHandler[object] | None = None` to configure the nested summary run, matching the handler API used by `SubAgents`. Supplying a handler selects streaming and lets callers observe summary events. The exported `drain_summary_events` helper consumes those events when only the streaming transport is needed:

```python
from pydantic_ai_harness.compaction import SummarizingCompaction, drain_summary_events

SummarizingCompaction(max_messages=60, event_stream_handler=drain_summary_events)
```

The default remains non-streaming. The handler receives the nested run's context; the outer run's handler is not inherited. Shared usage accounting, reserved usage limits, instructions, model settings, and durable summary execution are preserved.

## Validation

Ruff format/lint, focused Pyright, and compaction tests pass. Hosted CI, including the test matrix and coverage gate, passes on `b42e629c`. Tests cover the default request path, draining streamed events, and custom handler delivery. Both compaction documentation surfaces describe the API.

## Linked Issue

Fixes #619.

The corresponding `ToolOutputLimits` streaming support is tracked separately in #687.


JayGhiya:
@DouweM can we please fast rack this I am stuck for multiple agents as part of my OSS due to me using codex sub (they only support streaming)

DouweM:
@JayGhiya Sorry for the delay here, @dsfaccini will have a look.

dsfaccini (dismissed):
Requesting changes for one blocking integration issue:

- **net-positive -> do (blocking): rebase onto current `main` and preserve the current nested-run configuration while resolving the conflicts.** `git merge-tree --write-tree origin/main origin/pr-620` reports content conflicts in `docs/compaction.md` and `_summarizing_compaction.py`. More importantly, current `main` now constructs the summarizer with configurable `instructions=self.instructions` and calls the nested run with `usage_limits=reserved_usage_limits(ctx.usage_limits)`. The PR branch predates both changes and currently replaces that older call site. Please integrate `event_stream_handler=_drain_summary_events if self.stream else None` into the current call rather than dropping either the instructions or reserved usage limits, and update the merged tests/docs as needed.

I found no defect in the drain-handler approach itself. On the PR head, the focused compaction suite passes (376 tests), and Ruff format/check pass for the touched Python files.

dsfaccini:

Merged and reshaped your PR in place. CI is green on the current head. Sorry for the wait.

Your commit `4115b213` is untouched, same sha. We merged `main` into your branch rather than
rebasing, specifically so your commit was not rewritten. The diagnosis in
https://github.com/pydantic/pydantic-ai-harness/issues/619 was right, and the drain-handler
approach was right. We verified the premise independently before touching anything:
`https://chatgpt.com/backend-api/codex/responses` returns HTTP 400 `"Stream must be set to true"`
on `stream:false`, and no other seam reaches it, since `ModelSettings` has no `stream` key and the
nested summarizer `Agent` takes no `capabilities=`.

What we changed, and why:

`stream: bool` became `event_stream_handler: EventStreamHandler[object] | None`. Reason is
consistency. Harness already answers this exact question elsewhere: `SubAgents.event_stream_handler`
in `pydantic_ai_harness/subagents/_capability.py` is a capability driving a nested `Agent.run` and
it takes a handler, not a flag. Two fields for one idea would have been the wrong shape to ship.

The handler also does strictly more than the boolean could. A boolean can only mean "stream and
discard". A handler lets a caller watch the summary as it is written, which is useful for a progress
indicator on a long compaction. That path is covered by a new test.

For your case, nothing gets harder. `drain_summary_events` is exported from
`pydantic_ai_harness.compaction` and is the drop-in:

```python
from pydantic_ai_harness.compaction import SummarizingCompaction, drain_summary_events

SummarizingCompaction(max_messages=60, event_stream_handler=drain_summary_events)
```

One thing your issue surfaced that we did not fix here. `ToolOutputLimits._summarize` runs the same
nested `agent.run` with no stream seam, and `_summarize_action` catches `Exception` and falls back
to truncation, so the same provider fails silently there instead of erroring. Out of scope for this
PR, tracked at https://github.com/pydantic/pydantic-ai-harness/issues/687.

JayGhiya:
thankyou @dsfaccini for educating me on this and following up

PR #628, opened 2026-08-15
Stop `FileSystem` tools from leaking absolute host paths to the model

### LINKED ISSUE #616: Filesystem tool errors leak the absolute host path into `ModelRetry` messages

`_recoverable` in `pydantic_ai_harness/filesystem/_toolset.py` converts the
exceptions in `_RECOVERABLE_ERRORS` to `ModelRetry(str(e))`. For an `OSError`
subclass raised by the OS rather than by the toolset, `str(e)` renders as
`[Errno N] message: '<absolute host path>'`, so the model receives the host's
filesystem layout outside the root directory it was given.

`FileSystemToolset` raises most of these itself with a root-relative message,
which is why the leak is easy to miss. The ones that reach the model are the
OS-raised instances.

## Reproduction

```python
import asyncio, pathlib, tempfile
from pydantic_ai import ModelRetry
from pydantic_ai_harness.filesystem import FileSystemToolset

async def main():
    root = pathlib.Path(tempfile.mkdtemp())
    (root / 'hello.txt').write_text('hi')
    ts = FileSystemToolset(
        root_dir=root, allowed_patterns=['**/*'], denied_patterns=[],
        protected_patterns=[], max_read_lines=100,
        max_search_results=10, max_find_results=10,
    )
    try:
        await ts.read_file('hello.txt/nested')
    except ModelRetry as e:
        print(e)

asyncio.run(main())
```

Prints the absolute host path:

```
[Errno 20] Not a directory: '/private/var/folders/.../T/tmpXXXX/hello.txt/nested'
```

Confirmed on 3.10, 3.12 and 3.14.

## Scope

All four `OSError` subclasses in `_RECOVERABLE_ERRORS` are affected --
`PermissionError`, `FileNotFoundError`, `NotADirectoryError`, `IsADirectoryError`
-- across every operation the decorator wraps. `ValueError`, the fifth entry, is
raised only by the toolset's own code and carries no path.

This predates #602 / #612. That PR closed the two instances it touched by
catching at the `create_directory` call site (`FileExistsError`, and
`NotADirectoryError` for the parent-collision case) and by keeping the
`_RECOVERABLE_ERRNOS` messages path-free. The general case is untouched.

## Suggested direction

Render OS-raised errors with the model-supplied root-relative path instead of
`str(e)`, rather than adding a call-site catch per operation. The containment
guarantee documented in `pydantic_ai_harness/filesystem/README.md` ("Security
model" -> "Containment") is about which paths the agent may *reach*; a path it
cannot reach still being *named* in an error message is the gap here.

ada-zl5825:
## Summary

`str(OSError)` embeds the absolute host filename. Two model-visible surfaces forwarded it verbatim:

- `_recoverable` passed `str(e)` into `ModelRetry`, so OS-raised errors named the host filesystem layout. OS errors now keep `errno` and `strerror`, with the path rewritten relative to the configured root, or `<outside-workspace>` when it cannot be represented there.
- `file_info` returned the raw `os.readlink` target, so a symlink stored with an absolute target showed the host path in the tool result. Targets now go through the same helper.

Tool-authored error messages are unchanged. No changes to path authorization, containment, public APIs, or dependencies.

## Linked Issue

Closes #616.

## Checklist

### DIFF
--- a/pydantic_ai_harness/filesystem/_toolset.py
+++ b/pydantic_ai_harness/filesystem/_toolset.py
@@ -579,6 +579,7 @@ class FileSystemToolset(FunctionToolset[AgentDepsT]):
                 parts.append(f'hash: {_content_hash(text)}')
 
         if is_link:
-            parts.append(f'symlink_target: {os.readlink(original)}')
+            target = _model_safe_filename(os.readlink(original), self._real_root)
+            parts.append(f'symlink_target: {target}')
 
         return '\n'.join(parts)
--- a/pydantic_ai_harness/filesystem/_toolset.py
+++ b/pydantic_ai_harness/filesystem/_toolset.py
@@ -54,19 +54,14 @@ def _model_safe_filename(filename: str | bytes, real_root: Path) -> str:
 def _sanitize_recoverable_error(error: BaseException, real_root: Path) -> str:
     """Render a recoverable error without exposing absolute host paths.
 
-    Non-`OSError` exceptions and `OSError` instances with no `filename` are
-    returned as `str(error)`, preserving tool-authored messages. OS-raised
-    errors keep `errno` and `strerror`, and rewrite `filename`/`filename2` to
-    workspace-relative paths. Paths that cannot be represented relative to
-    `real_root` become `_OUTSIDE_WORKSPACE`.
+    Errors without an OS-supplied `filename` keep their original message.
+    OS errors keep `errno` and `strerror`, and the path is rewritten relative
+    to `real_root`, or replaced with `_OUTSIDE_WORKSPACE` when it cannot be.
     """
     if not isinstance(error, OSError) or error.filename is None:
         return str(error)
 
     filename = _model_safe_filename(error.filename, real_root)
-    if error.filename2 is not None:
-        filename2 = _model_safe_filename(error.filename2, real_root)
-        return f'[Errno {error.errno}] {error.strerror}: {filename!r} -> {filename2!r}'
     return f'[Errno {error.errno}] {error.strerror}: {filename!r}'
 
 
@@ -80,8 +75,7 @@ def _recoverable(
         try:
             return await fn(self, *args, **kwargs)
         except _RECOVERABLE_ERRORS as e:
-            # `_recoverable` is module-level, so `self._real_root` trips reportPrivateUsage.
-            real_root: Path = getattr(self, '_real_root')
+            real_root = self._real_root  # pyright: ignore[reportPrivateUsage]
             raise ModelRetry(_sanitize_recoverable_error(e, real_root)) from e
 
     return wrapper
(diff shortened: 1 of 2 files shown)


dsfaccini:

Reviewed alongside #612, which touches the same `_recoverable` seam. Three notes for whoever rebases -- none blocking, and the direction of this PR looks right.

**1. Two comments in `create_directory` go stale when this lands.** #612 added call-site catches there whose stated reason is exactly the leak this PR removes:

```python
except FileExistsError as e:
    # `exist_ok` only suppresses the error when the existing path is a
    # directory. Re-raising `str(e)` would leak the absolute host path.
```

The catches should survive -- `Path 'x' exists and is not a directory.` stays more actionable for a model than `[Errno 17] File exists: 'x'` -- but their justification changes from "avoids a leak" to "better message". Same for the `NotADirectoryError` catch below it and for `# str(e) embeds the absolute host path; the reason alone doesn't.` on the errno branch in `_recoverable`.

**2. Private helpers imported into tests.** `tests/filesystem/test_filesystem.py` now imports `_sanitize_recoverable_error`, `_recoverable` and `_OUTSIDE_WORKSPACE` from `_toolset`. `CLAUDE.md` ("Testing patterns") asks tests to drive behavior through the capability's public surface, and to mark `# pragma: no cover` on branches only reachable by calling a private helper directly. `test_write_through_file_hides_host_path` already demonstrates the public route works for the main case. The file has pre-existing `_content_hash`/`_format_lines` imports, so this is a convention point rather than a new violation.

**3. `getattr` drops the attribute out of type checking.**

```python
# `_recoverable` is module-level, so `self._real_root` trips reportPrivateUsage.
real_root: Path = getattr(self, '_real_root')
```

The annotation is doing the work the checker would otherwise do, which is the same objection `CLAUDE.md` raises against `cast()` under "Coding standards". The repo's existing answer to this is an explicit suppression -- see `pydantic_ai_harness/exa/_agent.py` lines 20-23, which use `# pyright: ignore[reportPrivateUsage]` for private access across the same kind of boundary. `self._real_root  # pyright: ignore[reportPrivateUsage]` keeps the real type and documents the exception in one line.

**Merge order.** #612 is green and ready, so it will likely land first. Test-merging the two produces 3 small conflict regions (the `_RECOVERABLE_ERRNOS` block adjacent to your sanitizer, the `_recoverable` body, and the test import block) -- all mechanical.

ada-zl5825:
Thanks for the review — this is exactly the rebase plan I wanted.

I'll wait for #612 to land, then rebase and address the three notes:

1. Keep the `create_directory` call-site catches (and the errno-branch
   message) for the more actionable model-facing text, and update the
   comments so they no longer justify those catches as leak prevention.
2. Keep the public `write_file` regression as the main coverage, and
   tighten the private-helper tests toward the public surface where I
   can do that without losing the filename2 / outside-root cases.
3. Replace the `getattr(self, '_real_root')` workaround with
   `self._real_root  # pyright: ignore[reportPrivateUsage]`.

The three conflict regions you listed look mechanical; I'll resolve
them on the rebase rather than stacking ahead of #612.

adtyavrdhn (approved):

PR #638, opened 2026-08-17
Add a `serializer` seam and `indented_json`/`json_lines` presets so `ToolOutputLimits` spills page by line

### LINKED ISSUE #614: Add optional JSON indentation for spilled structured results
### Capability Name

Indented structured spill output

### Description

`ToolOutputLimits` serializes structured tool results as compact JSON. Compact JSON keeps the payload small, but it puts a large list on one line. `read_tool_result` reads line ranges. It cannot page through items when the full list is on one line.

Add an optional `json_indent` setting to `ToolOutputLimits`. Apply it when the capability encodes a structured value for spill storage. Use `None` by default so the current compact output stays unchanged.

The setting must not alter bytes that a tool returns directly. `OverflowStore` must keep its byte contract. Text and binary inputs must stay byte-faithful. Only structured values that the capability encodes can use indentation.

### Use Case

A tool returns a large list of records. Compact JSON stores the full list on one line. The line-based read tool cannot read the list in smaller pages. With indentation, list items use separate lines, so the caller can read bounded ranges. Users who want the smallest payload can keep the default `None` value.

- `after_tool_execute`: encode structured return values before `Spill` stores them.
- `read_tool_result`: keep its current line-based API.
- `OverflowStore`: keep its current API and exact byte behavior.

### Prior Art / References

- Relates to https://github.com/pydantic/pydantic-ai-harness/pull/332, which adds more spill query options and markers.
- Relates to https://github.com/pydantic/pydantic-ai/issues/4352, which proposes bounded reads for large tool results.

This request does not close either item.

adtyavrdhn:
## Summary

Thanks to @andranik-sahakyan-mx for reporting #614 and the first fix in #615 — this PR takes that work forward (credited as co-author on the branch).

Spilled structured results are stored as compact JSON, all on one line, so the line-based `read_tool_result` tool can't page them. This adds a `serializer` field to `ToolOutputLimits` with two presets:

```python
ToolOutputLimits(serializer=indented_json)  # one field per line
ToolOutputLimits(serializer=json_lines)  # one record per line
```

Any `(value) -> str` callable works — same pattern as `tokenizer`, so we don't bake a format knob like `json_indent` into the API. The serialized text is the single representation that gets measured, previewed, spilled, and read back.

Details worth knowing:

- A serializer that raises or returns non-text warns and falls back to compact JSON, so a broken one never loses tool output.
- Both presets escape U+0085/U+2028/U+2029 (raw in JSON strings, but line breaks to `str.splitlines`), so read-back offsets always land on whole items.
- Strings and binary payloads never pass through the serializer; below-threshold returns pass through untouched.

Complementary to #332: line-shaped payloads make `grep_tool_result` matches meaningful for structured data.

Fixes #614

## Checklist

### DIFF
--- a/docs/tool-output-limits.md
+++ b/docs/tool-output-limits.md
@@ -155,17 +155,19 @@ ToolOutputLimits(serializer=json_lines)  # top-level list: one record per line
 ```
 
 `indented_json` renders indent-2 JSON, so fields land on separate lines and
-`read_tool_result` can slice and `pattern`-filter them. `json_lines` renders a top-level list
-as one compact JSON object per line (JSON Lines), so line offsets map directly to records; a
-value that is not a list falls back to `indented_json`.
+`read_tool_result` can slice and `pattern`-filter them. `json_lines` renders a sequence
+(other than a string or a byte payload) as one compact JSON object per line (JSON Lines), so
+line offsets map directly to items; other values fall back to `indented_json`. Both presets
+escape the Unicode line separators that `str.splitlines` breaks on, so a line read back is
+always a whole item or field.
 
-Any `(value) -> str` callable works; `Serializer` is the exported alias. The capability
-serializes each structured return once: measurement, band selection, previews, stored bytes,
-and read-back all use the same text, which means an indented layout measures larger than
-compact JSON and can reach a band the compact form would not. The callable must be
-deterministic, so an identical return reduces to identical bytes and keeps the prompt prefix
-stable. Strings and binary payloads never pass through the serializer, and a structured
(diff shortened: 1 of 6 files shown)


adtyavrdhn:
@andranik-sahakyan-mx Would love your thoughts on if you like this API

andranik-sahakyan-mx:
> @andranik-sahakyan-mx Would love your thoughts on if you like this API

This looks good, thank you for working on it! JSON lines is great, the original goal was to be able to have an orchestrator agent use code mode/dynamic workflows to divide/conquer a large list.

PR #522, opened 2026-07-31
Route harness integrations through `ctx.workspace`

adtyavrdhn:
## Summary

Use `ctx.workspace` for repository instruction discovery, asset inventory, and workspace-local Pydantic AI docs. Pass the exact workspace facade to delegated agents and accept `workspace=` in ACP session configuration, preserving application wrappers such as `ReadOnlyWorkspace`.

Local applications pass `LocalWorkspace(root=...)`. The bundled `coder_agent` supplies its current checkout so its existing CLI entry point continues to work. `Shell` and `FileSystem` retain their existing host execution behavior.

Workspace acquisition and lifecycle remain with the supplied backend and application. This layer adds no dispatcher or durable workspace operations. Documentation caches and repository instruction state are scoped to each run.

## Linked Issue

Implements the harness integration for the workspace design in pydantic/pydantic-ai#6492. Stacks on main; the predecessor #710 has merged. The temporary core dependency is pinned to `db57de423ca76e422037a43521ca7a3075cd6b4d` and must be replaced with a containing release before merge.

## Validation

Focused tests cover workspace reads, run-local state, filesystem events, ACP sessions, subagent forwarding, the bundled Coder, and existing event integrations. The Coder recorded integration was replayed; its cassette is unchanged from main and the observed instruction snapshot was updated. Documentation snippets and paired pages pass their checks. Ruff and explicit-file Pyright pass. CI runs the complete test and type-check matrix.

## Checklist

- [ ] Linked issue exists and is referenced above
- [x] Tests added/updated for new behavior
- [ ] `make lint && make typecheck && make test` passes locally (full-repository type checks and tests are CI-only)
- [ ] `pyproject.toml` and `uv.lock` are unchanged, or a maintainer added `dependencies:approved` to the current head
- [x] Docstrings use single backticks (not RST double backticks)

PR #504, opened 2026-07-29
Add E2B sandbox capability with Logfire tracing

vsokolovskii:
## Summary

- add `E2BSandbox` and `E2BSandboxSession` under the public `pydantic_ai_harness.e2b_sandbox` module
- support owned, attached, and caller-managed E2B sandbox lifecycles
- expose bounded command and file tools with completed-command tail capture and bounded timeout diagnostics
- emit content-free OpenTelemetry spans through the active Pydantic AI tracer so operations appear in Logfire
- add the optional `e2b>=2.34.0` package extra, package README, and unified documentation
- keep the E2B implementation self-contained; this PR has no changes under `pydantic_ai_harness/modal_sandbox`
- reject Temporal, DBOS, and Prefect durability composition at construction and for per-run additions because the run-scoped client lifecycle cannot safely cross replay boundaries

## Linked Issue

Fixes #509.

Refs #265. Supersedes #502, which was recreated from current upstream `main` as this PR to eliminate stale branch conflicts.

## Boundary notes

- **Downstream contract:** verified against E2B Python SDK 2.35.0 and the declared 2.34.0 floor. E2B launches commands through `/bin/bash -l -c`; the capture wrapper depends on Bash, `setsid`, `mkfifo`, `cat`, `dd`, `wc`, `tee`, and GNU `tail` in the standard `base` template.
- **Command/output boundary:** the wrapper redirects its inherited streams away from the SDK, captures command stdout/stderr inside the sandbox, and reads capture, prefix, and count artifacts through bounded streaming reads. Completed commands retain a bounded tail; timed-out commands retain an incrementally flushed bounded prefix. A focused POSIX reproduction covers short stderr followed by a timeout.
- **Known upstream limitation:** Bash login startup output occurs before command-level redirection, so `max_output_bytes` is not a hard SDK-memory ceiling if a reused sandbox has noisy or modified login files. Both docs call this out and recommend silent template login profiles until E2B exposes raw/no-buffer execution.
- **File boundary:** `read_file` checks metadata and then streams at most `max_read_bytes + 1`; the final model-facing command, file, and listing values are bounded again by byte and line limits.
- **Lifecycle:** owned sandboxes are killed on exit; already-gone sandboxes count as cleaned up; other owned cleanup failures warn with the sandbox id and retain state so `close()` can retry. Timeout cancellation makes best-effort process-group and command-handle cleanup. Attached and injected lifecycles are never killed by Harness.
[...]

PR #640, opened 2026-08-18
Port the Modal and E2B sandbox capabilities onto the pydantic-ai sandbox abstraction

adtyavrdhn:
## Summary

**Exploratory draft — not for merge as-is.** This ports the harness's two real sandbox providers onto the sandbox abstraction from pydantic/pydantic-ai#6492 (`pydantic-ai-slim` is pinned to the `sandbox-concept` branch via `[tool.uv.sources]`), to answer: does the abstraction fit implementations that weren't written for it, and what does the harness gain or lose? `ModalSandbox` comes from `main`; `E2BSandbox` is taken from #504 and reworked here. Sibling of #522, which ports the consuming capabilities; this PR is the provider side.

Shape per provider: the session machinery becomes a `SandboxBackend` implementation (`_backend.py`; `SupportsFilesystem` + `SupportsStart`, honestly-supportable subset only), the capability becomes the lifecycle supplier via `create_sandbox`/`get_sandbox`/`destroy_sandbox` with a run-scoped backend cache, and the toolset consumes `ctx.sandbox` — it now runs unchanged against `LocalSandbox` in tests.

### What the abstraction bought
- Both toolsets lost their lifecycle plumbing (`for_run` clones, session injection, `_require_session`) — Modal's toolset shrank ~40%.
- `Sandbox.resolve()` + `working_dir()` replaced both bespoke path layers.
- E2B's 140-line in-sandbox capture wrapper is deleted — and with it #504's open background-child reaping bug (`setsid` broke the group kill), which is now structurally impossible on the harness side.
- #504's durability rejection (`_validate_runtime_capabilities`, `for_agent`) is deleted: a serializable `SandboxRef` + reconnect in `get_sandbox` is exactly the mechanism it lacked. Removes the harness's only use of that private core hook.
- The protocol's cancellation contract forced a real improvement on E2B: client-owned deadline (`anyio.move_on_after` + `commands.kill`) instead of the SDK's abandon-on-timeout.

### Friction worth core's attention (both providers hit these independently)
1. **Timeout with partial output has no home**: `run()` is result-or-raise, so both ports invented a provider-specific `...CommandTimeoutError(TimeoutError)` carrying `stdout`/`stderr`, and both toolsets special-case it to show the model partial output. Standardizing `partial_stdout`/`partial_stderr` on a core timeout error would delete both.
2. **Client-side output bounding is inexpressible**: both old implementations bounded output before it reached (Modal) or left (E2B) the sandbox; the protocol's deliver-everything contract deleted both mechanisms. `| tail -c` is now the documented advice.
3. **Every supplier re-invents the run-scoped `(run_id, sandbox_id) → backend` cache** (`get_sandbox`/`destroy_sandbox` receive only a ref), and keying by sandbox alone is subtly wrong when two runs attach to one sandbox.
4. **The `pwd`-probe `working_dir()`** (probe + single-flight lock + validation) is duplicated character-for-character; neither platform reports a default cwd.
5. Smaller: the concrete `FileEntry` carrier isn't exported (both re-declare it); no error-taxonomy story, so each toolset's terminal-vs-retry ladder degrades to "unknown" for foreign backends; both platforms needed a post-hoc liveness probe (`is_alive()` candidate); E2B's `connect()` resumes paused sandboxes — a state the ref vocabulary can't express.

### Flagged decisions (maintainer input wanted)
- #504's `e2b.*` OTel spans are **removed**: their attributes came from the session the toolset owned; with `ctx.sandbox` being potentially any backend they'd be dishonest. If sandbox telemetry is wanted, it belongs in core around the `Sandbox` facade.
- E2B `workdir` default `/home/user` → `None` (probed); `secure=True` passed explicitly so an SDK default flip can't change behavior silently; `E2BSandbox` deliberately not re-exported from the package root yet.
- Modal floor `1.5.0` → `1.5.2` (`Sandbox.filesystem`).

## Boundary notes

- **Process lifecycle**: E2B commands are killed on deadline/cancellation via `commands.kill(pid)` (SIGKILL, command's own pid — children backgrounded *by the model's command* are not reached; documented and pinned by a live test; the owned-sandbox teardown reaps them at run end). Modal has no per-command kill: `kill()` raises `NotImplementedError`, deadlines are platform-owned, and the cancellation limitation is documented per the protocol's escape hatch.
- **Output limits**: no client/in-sandbox bounding remains (see friction 2); presentation-layer truncation in `_tool_output.py` is unchanged.
- **Cleanup failure**: teardown RPCs are time-bounded and shielded; `destroy_sandbox` failures propagate rather than being swallowed.
- Live tests for both providers are rewritten but **not run** (need real credentials; markers `modal_live`/`e2b_live`).

## Linked Issue

None — exploratory draft demonstrating pydantic/pydantic-ai#6492 against real providers. Related: #522 (consumer-side port), #504 (E2B origin), pydantic/pydantic-ai#6492.

## Checklist

- [ ] Linked issue exists and is referenced above (exploratory draft — see above)
- [x] Tests added/updated for new behavior (Modal 345, E2B 331; full suite 5031 passed; touched-module coverage 100%)
- [ ] `pyproject.toml` and `uv.lock` are unchanged, or a maintainer added `dependencies:approved` to the current head — **changed**: the sandbox-concept source pin, the `e2b` extra, and two test-infra lines (`e2b_live` marker, coverage omit); needs the label once on the current head


DouweM:
Superseded by #735 and #734

PR #661, opened 2026-08-21
Document dated removal guidance for deprecated APIs

### LINKED ISSUE #556: No convention in `agent_docs` for deprecating a released API with a dated removal

`agent_docs/` has no convention for deprecating a released public API with a dated removal, and
#548 is the first time the repo has needed one.

## Why it came up

`SpendStore` shipped in `v0.17.0` and had to change shape days later. #548 keeps the released
singular protocol working through a deprecating adapter and targets removal at `0.20.0`. Several
decisions in that PR had to be reasoned from scratch, and every one of them recurs:

- **New method names are forced, not stylistic.** `SpendStore` is `@runtime_checkable`, and runtime
  protocol checks test method *presence*, not signatures. Re-signaturing an existing method leaves
  no way to tell an old implementation from a new one and therefore no dispatch point on which to
  warn. Any future protocol deprecation in this repo hits the same constraint.
- **What the warning has to admit.** The compatibility adapter cannot deliver the fix -- it loops,
  which is the exact defect the new shape exists to remove -- so the warning says so rather than
  implying migration is cosmetic.
- **Data outlives the API.** A key-format change reaches counters no shim can touch, so #548 also
  carries a read-fallback and a dated point at which stale data stops counting. An API deprecation
  policy that only covers signatures misses this.
- **Where the removal date lives** so it is findable at removal time: the warning text, the
  docstrings, and the release note.

## What would settle it

A short subsection in `agent_docs/capability-authoring.md`, plus a line in
`agent_docs/review-checklist.md` for reviewing a deprecation. `README.md` and `docs/index.md`
already state the *policy* (0.x minor releases may break, with release-note migration guidance);
what is missing is the *mechanics* for authors.

Its own issue rather than part of #548 because contributor-doc changes do not ride in feature PRs,
and because the convention should be written from more than one example -- #548 is one data point,
not a policy.

Refs #548.

seanxuu:
Fixes #556

Adds repository guidance for deprecating released APIs with a named replacement, a dated removal release, warning/doc/release-note consistency, protocol compatibility, and persisted-data fallback testing.

Validation: `make lint`, `make typecheck`, `uv run codespell agent_docs/capability-authoring.md agent_docs/review-checklist.md`, and `uv run pytest tests/test_docs_parity.py -q` (277 passed).


mpfaffenberger (approved):
Reviewed against #556. The guidance matches what the codebase actually does: `HarnessDeprecationWarning` exists in `_warn.py`, and #747 (spend store deprecation) already follows this exact pattern including the dated removal, adapter, and warning tests, so this is codifying proven practice rather than aspiration. The review-checklist line gives reviewers a concrete gate. Docs-only, CI green.

PR #669, opened 2026-08-22
Expose the summarizer agent's `instructions` on `SummarizingCompaction`

### LINKED ISSUE #668: `SummarizingCompaction` hardcodes the summarizer agent's instructions
## Problem

`SummarizingCompaction` runs its nested summary request through an internal `Agent` whose `instructions` are hardcoded:

https://github.com/pydantic/pydantic-ai-harness/blob/main/pydantic_ai_harness/compaction/_summarizing_compaction.py#L625-L628

```python
agent: Agent[None, str] = Agent(
    cast('Model[Any] | str', model),
    instructions='You are a context summarization assistant. Extract the most important information from conversations.',
)
```

`summary_prompt` customizes the user turn of the summary request, but the system prompt is out of reach. Every other prompt surface of the strategy is a field; this one is not.

## Motivating case

Claude Code OAuth endpoints reject requests whose first system block is not the fixed string `You are Claude Code, Anthropic's official CLI for Claude.` (the endpoint fingerprints the leading system prompt; a mismatch surfaces as `rate_limit_error` 429s). A harness consumer that points `SummarizingCompaction(model=...)` at such an endpoint has no way to satisfy that requirement: the hardcoded instructions become the request's system prompt and the summary call fails, which in turn defeats compaction for the run (modulo `FallbackCompaction`).

Code Puppy hit exactly this after moving its compaction to the harness. Current workaround is a transport-level body rewrite that force-prepends the fingerprint block on the OAuth client ([mpfaffenberger/code_puppy@149faefe](https://github.com/mpfaffenberger/code_puppy/commit/149faefe)), which works but fixes at the wrong layer: the strategy should let the consumer say what the summarizer agent's system prompt is.

## Proposal

Add an `instructions` field to `SummarizingCompaction`, defaulting to the current string so behavior is unchanged:

```python
instructions: str = _DEFAULT_INSTRUCTIONS
"""Instructions for the internal agent that writes the summary.

`summary_prompt` shapes the user turn of the summary request; this shapes its
system prompt. Override it when the summarizer endpoint constrains the system
prompt, e.g. Claude Code OAuth endpoints require requests to open with a fixed
instruction string.
"""
```

and pass `instructions=self.instructions` in `_summarize`. Consumers who want to keep the default text after their required preamble can compose: `instructions=f'{required_prefix}\n\n{SummarizingCompaction.instructions}'` (or `dataclasses.replace`).

This mirrors the existing `summary_prompt: str = _DEFAULT_SUMMARY_PROMPT` pattern, adds no new branch, and is spec-expressible as a plain string field.

PR incoming.

mpfaffenberger:
Fixes #668.

`SummarizingCompaction` runs its nested summary request through an internal `Agent` whose `instructions` were hardcoded. `summary_prompt` customizes the user turn of the summary request, but the system prompt was out of reach.

That matters when the summarizer endpoint constrains the system prompt: Claude Code OAuth endpoints reject requests whose first system block is not a fixed instruction string, so pointing `model=` at one made every summary call fail (and with it, compaction for the run, modulo `FallbackCompaction`).

## Changes

- New `instructions: str` field on `SummarizingCompaction`, defaulting to the previous hardcoded string (`_DEFAULT_INSTRUCTIONS`), mirroring the `summary_prompt` / `_DEFAULT_SUMMARY_PROMPT` pattern. Behavior is unchanged unless the field is set.
- `_summarize` passes `instructions=self.instructions` to the internal agent.
- Docs parity: `pydantic_ai_harness/compaction/README.md` and `docs/compaction.md` both document the two prompt surfaces of the summary request.
- Tests: the default reaches the internal agent; an override reaches the internal agent (asserted through the same `patch('pydantic_ai.Agent')` seam the surrounding `TestSummarizingCompactionModel` tests use).

No new branches; a plain string field, spec-expressible as-is.

`make lint`, `pyright` (0 errors), and the full test suite (4460 passed) are green locally.

### DIFF
--- a/docs/compaction.md
+++ b/docs/compaction.md
@@ -341,6 +341,8 @@ agent = Agent(
 
 `model` accepts a model name or a `Model`; when left `None` it inherits the running agent's model. Its nested summary run inherits the parent usage limits and reserves one request from a finite request limit for the pending parent request. By default `incremental=True` updates the newest existing summary as an anchor. This changes the summary-call prompt from earlier releases; set `incremental=False` to retain the prior regeneration behavior.
 
+Both prompt surfaces of the summary request are fields: `summary_prompt` is the user-turn template (it must contain a `{messages}` placeholder) and `instructions` is the summarizer agent's system prompt. Override `instructions` when the summarizer endpoint constrains the system prompt, e.g. Claude Code OAuth endpoints require requests to open with a fixed instruction string.
+
 ### Usage accounting
 
 The summary call is a real request to the model, so its full usage -- tokens **and** the request itself -- is folded into the run's `ctx.usage`. This is deliberate: it keeps cost honest, keeps the request count consistent (a model request that did not count as one would be the surprise), and lets a `UsageLimits` request limit catch a runaway compaction. The nested run receives the other parent limits unchanged; the finite request limit is reduced by one so it cannot spend the slot already approved for the parent request. A run-request or iteration limiter will therefore see compaction calls among its requests.
--- a/docs/compaction.md
+++ b/docs/compaction.md
@@ -341,7 +341,7 @@ agent = Agent(
 
 `model` accepts a model name or a `Model`; when left `None` it inherits the running agent's model. Its nested summary run inherits the parent usage limits and reserves one request from a finite request limit for the pending parent request. By default `incremental=True` updates the newest existing summary as an anchor. This changes the summary-call prompt from earlier releases; set `incremental=False` to retain the prior regeneration behavior.
 
-Both prompt surfaces of the summary request are fields: `summary_prompt` is the user-turn template (it must contain a `{messages}` placeholder) and `instructions` is the summarizer agent's system prompt. Override `instructions` when the summarizer endpoint constrains the system prompt, e.g. Claude Code OAuth endpoints require requests to open with a fixed instruction string.
+Both prompt surfaces of the summary request are fields: `summary_prompt` is the user-turn template (it must contain a `{messages}` placeholder), and `instructions` sets the internal agent's static instructions, which Pydantic AI sends in the request's system prompt. Override `instructions` when the summarizer endpoint requires a fixed leading instruction.
 
 ### Usage accounting
 
--- a/pydantic_ai_harness/compaction/_summarizing_compaction.py
+++ b/pydantic_ai_harness/compaction/_summarizing_compaction.py
@@ -86,6 +86,10 @@ preamble, no markdown fences.
 </messages>\
 """
 
+_DEFAULT_INSTRUCTIONS = (
+    'You are a context summarization assistant. Extract the most important information from conversations.'
+)
+
 _SUMMARY_PREFIX = 'Summary of previous conversation:\n\n'
 
 # Anchored-incremental update instruction (opencode mechanism): the previous summary is fed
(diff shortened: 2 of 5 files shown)


pydantic_ai_harness/compaction/_summarizing_compaction.py
```diff
@@ -312,6 +316,14 @@ class SummarizingCompaction(AbstractCapability[AgentDepsT]):
     Must contain a ``{messages}`` placeholder.
     """
 
+    instructions: str = _DEFAULT_INSTRUCTIONS
```
dsfaccini: [P1] Adding this as a positional dataclass field shifts the existing `tokenizer` argument and every positional field after it. A previously valid call such as `SummarizingCompaction(None, 1, None, 0, None, '{messages}', len)` now binds `len` to `instructions`; the nested Agent then fails with `TypeError: object of type 'RunContext' has no len()`. Define this as `field(default=_DEFAULT_INSTRUCTIONS, kw_only=True)` and add it to `TestPositionalCompatibility`.

docs/compaction.md
```diff
@@ -341,6 +341,8 @@ agent = Agent(
 
 `model` accepts a model name or a `Model`; when left `None` it inherits the running agent's model. Its nested summary run inherits the parent usage limits and reserves one request from a finite request limit for the pending parent request. By default `incremental=True` updates the newest existing summary as an anchor. This changes the summary-call prompt from earlier releases; set `incremental=False` to retain the prior regeneration behavior.
 
+Both prompt surfaces of the summary request are fields: `summary_prompt` is the user-turn template (it must contain a `{messages}` placeholder) and `instructions` is the summarizer agent's system prompt. Override `instructions` when the summarizer endpoint constrains the system prompt, e.g. Claude Code OAuth endpoints require requests to open with a fixed instruction string.
```
dsfaccini: [P2] Please qualify this OAuth example. The available evidence describes an undocumented, model-dependent requirement, with Haiku explicitly reported as exempt, rather than a constraint shared by every Claude Code OAuth endpoint. Use a generic constrained-endpoint example or say that some endpoint/model combinations require a fixed leading instruction, and keep the README and class docstring wording aligned. Evidence: https://github.com/anthropics/claude-code/issues/40515

dsfaccini (changes requested):
> This review was posted by Codex Desktop using gpt-5.6-sol on behalf of David.

The issue is valid and the implementation exposes the required prompt surface, but two changes are needed before merge:

1. Preserve the existing positional constructor bindings by making `instructions` keyword-only and add the corresponding compatibility regression test.
2. Narrow the Claude Code OAuth documentation claim: the available public evidence describes an undocumented, model-dependent constraint rather than a requirement shared by every endpoint/model combination.

Focused compaction tests, lint, type checking, spec loading/schema, and current CI pass.

pydantic_ai_harness/compaction/_summarizing_compaction.py
dsfaccini: > This reply was posted by Codex Desktop using gpt-5.6-sol on behalf of David.

Fixed in 0b4549b6: `instructions` is keyword-only, and the positional compatibility test now passes a sentinel tokenizer through the legacy positional slot.

docs/compaction.md
dsfaccini: > This reply was posted by Codex Desktop using gpt-5.6-sol on behalf of David.

Fixed in 0b4549b6: the source, README, unified docs, and test now use a generic constrained-endpoint example and describe `instructions` as the internal agent static instructions sent in the request system prompt.

dsfaccini (approved):
> This review was posted by Codex Desktop using gpt-5.6-sol on behalf of David.

The requested changes are fixed in `0b4549b6`: constructor compatibility is preserved, the regression test covers the legacy positional tokenizer slot, and the endpoint wording is qualified consistently. CI is green and all review threads are resolved.

PR #677, opened 2026-08-24
Expose base fields in capability `from_spec` schemas

### LINKED ISSUE #554: Five capabilities' `from_spec` omits the `AbstractCapability` base fields, so a spec cannot set `id`/`description`/`defer_loading`

Review level: found by the repo-wide sweep in #546 and verified per capability against pydantic-ai 2.22.0. David directed the work and approved filing.

One of four root causes behind incomplete `AgentSpec` schemas. The guard in #546 xfails these at
layer 3; this issue is what that xfail points at.

## The defect

`id`, `description` and `defer_loading` are declared on `AbstractCapability` and reach `cls()`
through `**kwargs`, so they keep working at runtime. They never appear in the schema, so **a spec
cannot set them on these capabilities today** and an editor following the `$schema` line rejects
them.

## Affected

`BrowserUse`, `ExaAgent`, `ExaSearch`, `Memory`, `Planning`.

Each fix is naming the three parameters in `from_spec` and passing them through --
`SpendLimits.from_spec` and `StepPersistence.from_spec` (#543) are the reference implementations.

`Skills` fails the same check for a different reason and belongs to #553, not here.

Refs #537, #546.

mimimind:
## Summary
- expose `id`, `description`, and `defer_loading` in generated `AgentSpec` schemas for `BrowserUse`, `ExaAgent`, `ExaSearch`, `Memory`, and `Planning`
- forward values through each `from_spec()` constructor
- add schema and round-trip regression coverage

Fixes #554.

## Boundary notes
- the default remains `defer_loading=false`, so existing specs keep the same visible tools, instructions, and runtime behavior
- an explicit `defer_loading=true` now reaches the existing `AbstractCapability` loading gate instead of being impossible to express through `AgentSpec`
- focused schema and round-trip tests cover both inherited field exposure and constructor forwarding for all five affected capabilities
- this change introduces no new command, parser, process, network, cleanup, resource-limit, dependency, or CI boundary

## Validation
- 13 focused spec tests passed
- affected module suites: 539 passed; two unrelated Windows-only environment failures (path length and symlink privilege)
- `ruff format --check .`
- `ruff check .`
- Pyright on changed source/tests: 0 errors

## Checklist

### DIFF
--- a/pydantic_ai_harness/browser_use/_capability.py
+++ b/pydantic_ai_harness/browser_use/_capability.py
@@ -352,6 +352,9 @@ class BrowserUse(AbstractCapability[AgentDepsT]):
         session_scope: Literal['call', 'agent'] = 'call',
         cdp_url: str | None = None,
         guidance: str | None = None,
+        id: str | None = None,
+        description: str | None = None,
+        defer_loading: bool = False,
     ) -> BrowserUse[AgentDepsT]:
         """Construct the capability from serializable spec options.
 
@@ -371,4 +374,7 @@ class BrowserUse(AbstractCapability[AgentDepsT]):
             session_scope=session_scope,
             cdp_url=cdp_url,
             guidance=guidance,
+            id=id,
+            description=description,
+            defer_loading=defer_loading,
         )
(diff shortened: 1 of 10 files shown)


mpfaffenberger (approved):
Reviewed against #554: all five capabilities named there (BrowserUse, ExaAgent, ExaSearch, Memory, Planning) now accept and forward `id`/`description`/`defer_loading`, and the schema tests were strengthened from string-containment to real property assertions, which is a nice upgrade. Verified locally: exa/memory/planning suites pass (307 passed) and the branch merges clean with current main. The browser_use module collection requires the extra, which CI's all-extras jobs cover and they are green.

PR #678, opened 2026-08-24
Add stable IDs to one-off capabilities

### LINKED ISSUE #564: Give one-off harness capabilities a default `id` and plumb it through to their toolsets

David reviewed the summary of this draft and approved filing it (light review).

Harness capabilities that contribute a leaf toolset don't reliably carry an `id`, so an agent that combines one with durable execution (Temporal, DBOS, Prefect) fails at construction with:

> Toolsets that are 'leaves' (i.e. those that implement their own tool listing and calling) need to have a unique `id` in order to be used with Temporal.

That fires for a capability the user never parameterized. Pydantic AI is fixing the same gap for its own one-off capabilities (`WebSearch`, `WebFetch`, `ImageGeneration`, and `XSearch` get hardcoded default ids). This issue is the harness half.

Beyond durable execution: agent-control (Logfire remote override) targets capabilities, toolsets, and the instructions that inherit their id individually, so these ids become a public, stable override surface. Pick names we are willing to keep.

## Two separate defects

Harness capabilities fall into three states today.

**1. No id anywhere.** The capability has no default `id`, and its toolset is constructed without one, so even a user-set `id=` would not help. Durable execution raises. Verified: `Shell`, `FileSystem`, `SubAgents`, `ModalSandbox`, `ExaSearch`, `ExaAgent`, `CapabilityCreation`, `BrowserUse`, `RepoContext`, `PydanticAIDocs`, `LocalStack`, `Macroscope`, `ToolOutputLimits`, and `DynamicWorkflow` (which forwards `toolset_id=self.id`, defaulting to `None`).

**2. Toolset id hardcoded, capability `id` ignored.** Works under durable execution, but silently drops a user-set `id=`, and diverges from the Pydantic AI contract that a capability stamps its own toolset. Verified: `Memory` (`id='memory'`), `Planning` (`id='planning'`), `SpendLimits` (`id='spend'`).

**3. Already correct.** `StackOne` uses `id: str | None = 'stackone'` and forwards it as `id=self.id or _DEFAULT_ID`. That is the pattern to copy.

## Ask

For every one-off harness capability (one fixed built-in identity), do both:

1. give the capability a default `id` on its constructor, overridable by a user-passed `id=`;
2. forward that id to the toolset it builds, replacing any hardcoded toolset-level id.

Capabilities whose identity depends on what the user puts in them (guardrails, anything wrapping a user-supplied toolset, agent, or store where more than one instance per agent is normal) keep `id=None`. That is the same carve-out Pydantic AI is applying to `Capability(tools=...)`, `Toolset`, `DynamicCapability`, and `NativeTool`.

Naming: Pydantic AI uses `to_snake(ClassName)`, which is also what the run already derives when `id is None`, so nothing shifts in message history or Logfire attributes. Suggest the harness match (`shell`, `file_system`, `sub_agents`, and so on) unless a shorter established name already exists in the docs.

## Note

Two instances of the same one-off capability in one agent will start raising `Capability id '...' is used by multiple capabilities` at construction, where today they auto-suffix. That is the intended outcome, but worth checking that the harness has no capability where two instances is a normal composition. Candidates to double check: `SubAgents`, `ModalSandbox`, `ExaSearch`. Any that qualify belong in the carve-out.

## References

- Pydantic AI prerequisite PR (the same change on the Pydantic AI side): https://github.com/pydantic/pydantic-ai/pull/7248
- Pydantic AI PR that surfaced the gap: https://github.com/pydantic/pydantic-ai/pull/7144
- Original report: https://github.com/pydantic/pydantic-ai/issues/7109

The inventory above was read from the harness `main` checkout on 2026-08-06. Treat it as a starting audit, not a closed list.

mimimind:
## Summary
- assign stable default ids to fixed one-off capabilities and forward them to their leaf toolsets
- include `PlaywrightBrowser`, `YouSearch`, and `YouResearch`, which landed after the issue's starting inventory
- preserve custom ids across run-scoped toolset clones and established toolset ids when `id=None`
- keep `Memory` capability ids auto-suffixed because multiple scoped memories are a supported composition, while forwarding a custom id to its toolset

Fixes #564.

## Validation
- 28 focused capability/toolset id regression tests passed
- 866 selected affected tests passed, 2 skipped, 4 known Windows-only tests deselected
- BrowserUse: 159 passed
- Playwright and You.com: 417 passed
- Ruff formatting and lint checks passed
- Pyright on changed files: 0 new errors

## Checklist

### DIFF
--- a/pydantic_ai_harness/browser_use/_capability.py
+++ b/pydantic_ai_harness/browser_use/_capability.py
@@ -89,6 +89,9 @@ class BrowserUse(AbstractCapability[AgentDepsT]):
     judgement about unknown pages, not for scripted flows.
     """
 
+    id: str | None = field(default='browser_use', kw_only=True)
+    """Stable capability and toolset ID."""
+
     llm: ChatModelInput | None = None
     """The chat model driving the sub-agent.
 
@@ -294,6 +297,7 @@ class BrowserUse(AbstractCapability[AgentDepsT]):
             if sensitive_data is not None and browser_profile is not None:
                 browser_profile = browser_profile.model_copy(deep=True)
             self._toolset = BrowserUseToolset[AgentDepsT](
+                id=self.id,
                 browser_agent=self.browser_agent if self.browser_agent is not None else default_browser_agent,
                 llm=resolve_chat_model(self.llm),
                 browser_profile=browser_profile,
--- a/pydantic_ai_harness/browser_use/_toolset.py
+++ b/pydantic_ai_harness/browser_use/_toolset.py
@@ -413,6 +413,7 @@ class BrowserUseToolset(FunctionToolset[AgentDepsT]):
     def __init__(
         self,
         *,
+        id: str | None = None,
         browser_agent: BrowserAgentFactory,
         llm: BaseChatModel | None,
         browser_profile: BrowserProfile | None,
@@ -428,7 +429,7 @@ class BrowserUseToolset(FunctionToolset[AgentDepsT]):
         session_scope: Literal['call', 'agent'],
         cdp_url: str | None,
     ) -> None:
-        super().__init__()
+        super().__init__(id=id)
         self._browser_agent = browser_agent
         self._llm = llm
         self._browser_profile = browser_profile
--- a/pydantic_ai_harness/capability_creation/_capability.py
+++ b/pydantic_ai_harness/capability_creation/_capability.py
@@ -2,7 +2,7 @@
 
 from __future__ import annotations
 
-from dataclasses import dataclass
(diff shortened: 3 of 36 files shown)


dsfaccini:

Could you turn on "Allow edits by maintainers" on this PR?

We have the CI fix ready and would rather push it to your branch than open a second PR.

Your diagnosis is right and the fix works: 16 of 21 one-off capabilities raise under Temporal on harness main, and all 21 construct after your diff.

The id strings are on hold pending a maintainer decision about duplicate instances, so nothing else is waiting on you.

PR #682, opened 2026-08-24
Allow `SummarizingCompaction` to stream summary requests

### LINKED ISSUE #619: SummarizingCompaction nested summary request cannot use stream-only providers

AyobamiH:
## Summary

Add an opt-in `stream_summary` flag to `SummarizingCompaction`. When enabled, only the nested summarizer request uses Pydantic AI's existing streaming path; its events are consumed internally and the completed summary contract is unchanged. The default remains non-streaming.

The regression coverage uses a stream-only `FunctionModel` and a non-streaming-only `FunctionModel` through real `Agent` runs so selecting the wrong transport fails at the model boundary. The new option is keyword-only to preserve constructor compatibility.

Both compaction documentation surfaces are updated. No dependency or lockfile changes.

### Boundary notes

This change uses the public `Agent.run(event_stream_handler=...)` primitive. It does not add provider wire handling, model-specific behavior, or alternate agent-loop semantics in Harness.

## Linked Issue

Fixes #619

## Checklist

- [ ] `make lint && make typecheck && make test` passes locally (not run locally in this environment; the repository's full CI matrix has passed on the fork and will be re-run for any code follow-up)

### DIFF
--- a/docs/compaction.md
+++ b/docs/compaction.md
@@ -343,6 +343,8 @@ agent = Agent(
 
 Both prompt surfaces of the summary request are fields: `summary_prompt` is the user-turn template (it must contain a `{messages}` placeholder), and `instructions` sets the internal agent's static instructions, which Pydantic AI sends in the request's system prompt. Override `instructions` when the summarizer endpoint requires a fixed leading instruction.
 
+If the summary endpoint only accepts streaming requests, set `stream_summary=True`. Only the nested summary request is streamed; its events are consumed internally and the parent run still receives one completed summary. The default remains non-streaming.
+
 ### Usage accounting
 
 The summary call is a real request to the model, so its full usage -- tokens **and** the request itself -- is folded into the run's `ctx.usage`. This is deliberate: it keeps cost honest, keeps the request count consistent (a model request that did not count as one would be the surprise), and lets a `UsageLimits` request limit catch a runaway compaction. The nested run receives the other parent limits unchanged; the finite request limit is reduced by one so it cannot spend the slot already approved for the parent request. A run-request or iteration limiter will therefore see compaction calls among its requests.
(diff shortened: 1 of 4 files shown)


dsfaccini:

Closing in favour of #620, which opened first with the same approach.

We're taking #620 forward now — rebasing it onto main and reshaping the opt-in to an `event_stream_handler` field, to match the precedent SubAgents already sets. Thanks for the report either way; the underlying bug is real and it's getting fixed.

PR #703, opened 2026-08-26
Add the `final-review` skill for PRs that have already been iterated on

dsfaccini:

David approved the design and the two structural calls. He has not read the file line by line.

Adds `.agents/skills/final-review/SKILL.md`, a deliberately-triggered review pass for a PR that has already been iterated on. It is committed rather than kept local so contributors and maintainers can run it too.

## Why this shape

The design is measured, not assumed. Two studies fed it.

**What actually produced findings.** Across four harness PRs reviewed in one session, every substantive finding came from running two versions of the code against the same input and printing a table. Findings produced by reading alone were doc items or duplicates of something a probe had already produced.

**What the wider sample said about that premise.** Measured against the last 25 merged PRs in this repo, a probe-only design covers roughly 40 to 45 percent of the findings that actually changed a diff. It cannot reach the largest single class, which is prose claiming more than the code delivers: 4 of the 8 PRs that received any review finding at all. `tests/test_docs_parity.py` enforces that both surfaces exist and match each other, and can never check whether either one is true.

So the skill has two lanes. Lane A charters run a probe. Lane B charters read, and are not asked to pretend otherwise. Both feed one adjudication pass.

## What it deliberately does not do

Macroscope posts per-line correctness with an approvability verdict, and Veria posts security findings, on every PR without being asked. This skill does not re-report either. Its subject is what those two cannot see: the iteration itself.

## Scope decisions a reviewer could mistake for oversights

- `.agents/` is excluded wholesale by `.gitignore`, so this file needed `git add -f`. That is deliberate: `docs-parity-reviewer.md` is tracked the same way.
- The concern lists in Lane A and Lane B are ranked by how often each concern changed a merged PR in this repo, not by category tidiness. CI and workflow-graph semantics leads Lane B because it produced more diff-changing findings in the sample than any other single concern.
- Durability composition and cassette re-recording appear in Lane A because `agent_docs/review-checklist.md` demands both. Neither appeared in the 25-PR sample. Either they never bite or nobody checks, and that is not determined.

## Checklist

- [x] Linked issue exists and is referenced above, or the change needs none

### DIFF
--- a/.agents/skills/final-review/SKILL.md
+++ b/.agents/skills/final-review/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: final-review
-description: Run a final review pass on a harness PR that has already been iterated on - bots have posted, threads have been resolved, and commits have landed since the first review. Use it before asking a maintainer to merge, or when a PR has changed a lot since it was last read. Do NOT use it as a first read of a fresh PR: it audits what earlier rounds claimed, so with no earlier rounds most of it has nothing to work on. Do NOT use it to re-report per-line correctness or security findings - Macroscope and Veria already post those on every PR.
+description: Run a final review pass on a harness PR that has already been iterated on - bots have posted, threads have been resolved, and commits have landed since the first review. Use it before asking a maintainer to merge, or when a PR has changed a lot since it was last read. Do NOT use it as a first read of a fresh PR: its highest-value charters audit what earlier rounds claimed, so with no earlier rounds most of it has nothing to work on. Do NOT use it to re-report per-line correctness or security findings - Macroscope and Veria already post those on every PR.
 ---
 
 # Final review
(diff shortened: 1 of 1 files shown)


dsfaccini:

Ran this skill against harness PR #589 before asking anyone to merge it. It found a defect in itself, so the skill has been rewritten. What changed and why:

**The yield was inverted.** Three Lane A charters, roughly 320k tokens, zero findings. Two Lane B charters, roughly 240k tokens, four findings. Every finding that landed came from reading, not from running. The first draft spent eight bullets and five ranked primitives on Lane A and gave Lane B four bullets and no method. Lane B now gets method, and the claim charters are stated as the reason to run the pass at all.

**A concurrency defect.** Phase 2 said "dispatch one agent per charter, in parallel" while two of the five probe primitives mutate the shared working tree. Three of four charters independently reported tree corruption, and one warned that an uncommitted mutation it saw "would neuter the entire fix if committed". Mutating probes are now isolated-tree or serial, and isolation by `git worktree` or `git archive` export is the top-ranked primitive. The primitive the draft ranked strongest was used by nobody.

**A missing provenance assertion.** One charter ran `python -m pytest` from the wrong directory; `-m` puts the current directory on `sys.path` ahead of `PYTHONPATH`, so its first measurement silently read a concurrently-mutating tree. Every probe now asserts its resolved module path and revision before reporting a number.

**The vacuity rule was wrong and would have deleted a good test.** Revert-and-rerun cannot separate a vacuous test from the positive-control half of a boundary pair. The charter now escalates to a targeted mutation, and distinguishes ERROR from FAIL.

**Phase 3 is in-session by default.** Dispatching it cost two infrastructure deaths and produced nothing; adjudicating in-session took a fraction of the time, because the caller has just read every charter result in full.

Also corrected: "give each agent its charter and nothing else" (withhold conclusions, never environment constraints), two pre-pass checks that produced false positives or passed vacuously, a Phase 4 instruction that contradicted the skill's own constraints, and a RESCOPED rule that assumed write access. Added a cost note, a resume-before-failing rule, and two charter types the run invented on its own.

Full critique is in the branch notes for #589.

Issue #690, opened 2026-08-25
Doc snippets are checked for parse and imports but not for undefined names, so two `spend` snippets fail on the 3.10 floor


dsfaccini:
> This issue was posted by Claude Code using claude-opus-5 on behalf of David.

`agent_docs/review-checklist.md` requires that "every snippet in both docs is runnable: all imports
present". `tests/test_doc_snippets.py` does not check that, and says so in its own docstring: it
parses each block and resolves the harness symbols it imports, and "deliberately does not execute
the snippets". So a snippet missing a non-import name passes CI on every matrix leg.

Two snippets in `docs/spend.md` are not runnable on the declared 3.10 floor because of it. Both
annotate a parameter with `SpendLimits[None]` without importing `SpendLimits`:

- `docs/spend.md` and `pydantic_ai_harness/spend/README.md`, the `status()` section:
  `async def report(limits: SpendLimits[None]) -> None:`
- the same pair, the workflow-admission section:
  `async def start_if_funded(limits: SpendLimits[None], tenant_id: str) -> None:`

Verified by running the first block on its own:

```
$ uv run --no-project --python 3.10 python snip.py
NameError: name 'SpendLimits' is not defined
$ uv run --no-project --python 3.14 python snip.py
(clean)
```

3.14 lets it pass because PEP 649 defers annotation evaluation; 3.10 evaluates them at `def`. The
second snippet also calls an undefined `workflow_handle`, which is illustrative rather than
copy-pasteable.

## Why this was not fixed in #548

Neither snippet is in a section #548 touches, and three open PRs (#547, #548, #549) all edit
`docs/spend.md`. Reaching into an untouched section to fix a pre-existing snippet buys a conflict.

## What this needs

Two things, and the second is the one that stops it recurring:

- The missing imports added to both snippets, in both surfaces, and a decision on
  `workflow_handle` (define it or fence the block).
- A snippet check that catches an undefined name. `pytest-examples` (already a dependency, already
  used here) can lint each block against the names it defines and imports; that is cheaper than
  executing them and would have caught both. A repo-wide sweep of the other 33 pages under whatever
  check lands is worth doing in the same PR.

Refs #548.

dsfaccini:
> This comment was posted by Claude Code using claude-opus-5 on behalf of David.

The two broken snippets are fixed in #706. This issue stays open for the test-side half, which is bigger than the description implies -- so here are numbers for whoever picks it up.

[...]

PR #706, opened 2026-08-26
Import `SpendLimits` in the two `spend` doc snippets that annotate with it

dsfaccini:

Two snippets in the `spend` docs annotate a parameter with `SpendLimits[None]` without importing it, so they raise `NameError` on the declared 3.10 floor:

```
$ python3.10 snip.py
NameError: name 'SpendLimits' is not defined
```

3.14 hides it -- PEP 649 defers annotation evaluation, and the dev interpreter is 3.14. Both blocks appear on both surfaces (`docs/spend.md` and `pydantic_ai_harness/spend/README.md`), so four blocks change.

The admission block also awaited an undefined `workflow_handle`. It now takes the starter as a parameter, which keeps what the block teaches -- check the budget before starting the workflow -- and makes it runnable.

Verified on 3.10 against a stub package: the previous form raises `NameError`, both new forms execute cleanly. `ruff check --select F821` is clean on both.

## What this does not do

Closes the concrete half of #690. The test-side half stays open, because it is bigger than the issue implies. `tests/test_doc_snippets.py` parses each block and resolves the harness symbols it imports, and deliberately does not execute -- so any missing non-import name passes on every matrix leg.

I measured what turning that check on would cost, across `docs/`, every capability `README.md`, and the root `README.md`:

| scope | snippets | fail `F821` |
|---|---|---|
| all non-skipped | 449 | 99 |
| only those carrying their own imports | 392 | 42 |

Most of those are continuation blocks that deliberately reuse `agent`, `store` or a name bound in the block above, which is an idiom the docs use throughout -- not bugs. Even the narrower heuristic leaves 42 needing a per-snippet judgement. So it is a triage job with a design decision attached (an opt-out directive, snippet grouping, or executing them), not a flag flip. The numbers are on #690 for whoever takes it.

### DIFF
--- a/docs/spend.md
+++ b/docs/spend.md
@@ -107,6 +107,9 @@ SpendLimits(budgets=[Budget(usd=Decimal('100'))], on_spend=show)
 `status()` reads the same numbers without a run, which is what a cost display in a UI wants:
 
 ```python
+from pydantic_ai_harness.spend import SpendLimits
+
+
 async def report(limits: SpendLimits[None]) -> None:
     for status in await limits.status(scope='acme'):
         print(status.budget.name, status.spent.usd, status.exhausted)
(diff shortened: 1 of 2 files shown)


dsfaccini:

Closing as a duplicate. #701 was opened 18 hours before this one and fixes the same two snippets; I did not check for an open PR against #690 before starting, which is on me.

#701 is also the better change of the two:

- It adds a regression test, which this PR does not.
- Its import form is the correct one. It uses `from pydantic_ai_harness import SpendLimits`, matching the five other snippets on that page; I used `from pydantic_ai_harness.spend import SpendLimits`, which would have been the only submodule-path import on the page.

The rest converged independently -- both replace the undefined `workflow_handle` with an injected `Callable[[], Awaitable[None]]` starter.

Nothing here needs carrying over. The sizing of the wider snippet-test gap is already on #690 as a comment and is independent of either PR.

PR #714, opened 2026-08-27
Deprecate `on_spend`, `on_usage`, `on_fire`, and `PlanEventEmitter` toward capability event subscriptions

### LINKED ISSUE #711: Adopt core's capability event family: filesystem events, compaction lifecycle events, and callback deprecations
pydantic/pydantic-ai#6258 gives core two typed event families (`CustomEvent`, `CapabilityEvent`) with `@on_event` subscriptions, inline-dispatched decision events, and live stream delivery; pydantic/pydantic-ai#7801 adds the shared compaction lifecycle family (`CompactionStartEvent` — cancellable — and `CompactionEndEvent` in `pydantic_ai.capabilities.compaction`). It's the core-native answer to the [unified domain events proposal](https://github.com/pydantic/pydantic-ai-notes/pull/15)'s audit, and Harness is the first consumer:

- [ ] **FileSystem emits traversal events** (`FileReadEvent`, `DirectoryListedEvent`, `FileWrittenEvent` in a `file_system` namespace, carrying the normalized authorized path), and **RepoContext subscribes with `@on_event`** for nested-`AGENTS.md` surfacing, deprecating its `traversal_tool_names`/`traversal_path_arg` tool-name sniffing.
- [ ] **Compaction strategies emit the core lifecycle family**: every strategy attempt emits the cancellable `CompactionStartEvent` (a capability mid-activity, or app code via `hooks.on.event`, can hold compaction for a turn) and a changed-only `CompactionEndEvent`; a cancelled fallback candidate is a deliberate skip, not a failure that advances the chain.
- [ ] **Deprecate the ad-hoc constructor callbacks** (`on_spend`, `on_usage`, `on_fire`, …) toward event subscriptions.

Requires a pydantic-ai release carrying the two core PRs before any of this can merge.

DouweM:
## Summary

The ad-hoc constructor callbacks are deprecated toward [capability event](https://github.com/pydantic/pydantic-ai/pull/6258) subscriptions — one typed stream instead of four bespoke callback shapes. Each event is always emitted; each legacy callback keeps working and warns with `HarnessDeprecationWarning` at construction:

| Deprecated | Replacement event | Kind |
|---|---|---|
| `SpendLimits.on_spend` | `SpendRecordedEvent` (flat response fields + serializable `SpendBudgetStatus` entries) | `spend.recorded` |
| `ReportContextUsage.on_usage` (now optional) | `ContextUsageEvent` (`used_tokens`, `window_tokens`, `resolved`, `fraction`) | `context_usage.measured` |
| `SystemReminders.on_fire` | `ReminderFiredEvent` (rendered text, committed only once the reminder is actually appended) | `system_reminders.fired` |
| `PlanEventEmitter` / `EventCallback` / stores' emitter wiring | `PlanCreatedEvent`, `PlanUpdatedEvent`, `PlanStatusChangedEvent`, `PlanCompletedEvent`, `PlanDeletedEvent`, derived from before/after store snapshots on the Planning tool paths | `plan.created` … |

Plan events are emitted where a `RunContext` exists (the planning tools); direct store mutations from application code have no run and emit no run events — documented, and the deprecated emitter still covers that path for those who need it.

Docs and READMEs for spend, context usage, system reminders, and planning now show subscriptions as the primary API with a short migration note each.

Requires an unreleased pydantic-ai (pydantic/pydantic-ai#6258); don't merge before a core release carries it.

## Linked Issue

Closes #711 (the filesystem and compaction items land via #712 and #713).

## Checklist

- [x] `make lint && make typecheck && make test` passes locally (against pydantic-ai `emit-custom-events`)
---

**Unblocked.** [pydantic-ai 2.38.0](https://github.com/pydantic/pydantic-ai/releases/tag/v2.38.0) carries the event API ([#6258](https://github.com/pydantic/pydantic-ai/pull/6258)), so this no longer needs an unreleased core.

The floor bump itself is #780, which is green. Until it merges this branch carries its commit so CI here can run; once #780 lands, merging `main` in is a no-op on those two files.

Verified against the released `pydantic-ai-slim==2.38.0` from PyPI (not a local checkout): **5,602 passed, 63 skipped**, plus ruff lint, ruff format (333 files) and pyright at 0 errors / 0 warnings.


mpfaffenberger:
Hey Douwe, merging this one on my call since CI is green and it's clean against main. The six open bot threads read as non-blocking to me: three are `_emit_changes` edge cases with custom `PlanStore`s and concurrent store mutations, and the two Veria notes are about the payloads being visible on `run_stream_events`, which is the intended surface. Happy to follow up on any of them if you see it differently.

This unblocks the harness-native CLI, which subscribes to `ContextUsageEvent`, `SpendRecordedEvent`, and the plan events for its status line. I have sub-agents rebasing #712 and #713 and working through the Macroscope threads there next.

DouweM:
Took you up on the offer and checked the four Macroscope threads against merged main. Two hold up, one doesn't, one I'd leave. Filed the survivors as #824.

**Doesn't reproduce — r3868323580** (a subscriber mutating `event.item.status` suppresses `PlanStatusChangedEvent` and `PlanCompletedEvent`). The aliasing it describes is real: `PlanUpdatedEvent` is a dataclass, so `event.item is item`, and writing through the event does write the snapshot. But no listener runs in that window. All five plan events carry `event_dispatch='stream'`, and `dispatch_event_immediate` returns early on anything that isn't `'immediate'`, so `_emit_changes` has emitted all three before the first reaches a listener. Both ways on a real agent run:

```
no mutation  : ['created', 'updated', 'status_changed', 'completed']
with mutation: ['created', 'updated', 'status_changed', 'completed']
```

Worth remembering if a plan event is ever switched to immediate dispatch, because then it becomes live.

**Holds — r3868323561** (reorder emits nothing). `id` is in `PlanItem`'s schema, so the model resends ids, and a `write_plan` that only reorders produces no per-item diff:

```
after first write : [('0385c228', 'A'), ('92b58dc8', 'B')]   emitted: ['PlanCreatedEvent', 'PlanCreatedEvent']
after reorder     : [('92b58dc8', 'B'), ('0385c228', 'A')]   emitted: []
```

The store takes the new order and a subscriber rebuilding from events keeps the old one. Reprioritizing is ordinary model behavior, so a status line driven off these events can show a stale order.

**Holds as a contract gap — r3868323576.** `_emit_changes` depends on `get_items()` returning detached copies, and the `PlanStore` protocol never says so — the docstring is just "Return every step in insertion order". Both built-in stores comply, so nothing is broken today; a third-party store returning its backing list stops emitting silently.

**Would leave — r3919467895.** Out-of-band mutation racing the two snapshots. Inherent to diffing, and the docs already say a direct store mutation has no run context and reaches no listener.

One more that no bot caught, and the reason I went looking: `PlanStore.set_items`'s docstring still says the bulk replacement "does not emit `PlanEvent`s -- so the `write_plan` tool, which calls it, is event-silent", and tells applications to read the plan after a run or steer the model toward the granular tools. True of the deprecated `PlanEvent`; `write_plan` now emits capability events, as the repro above shows. Anyone migrating onto the new events is being told to work around a limitation that no longer applies to them.

Agree with you on the two Veria notes — those payloads on the run stream are the intended surface.

PR #716, opened 2026-08-27
Declare the `httpx[brotli]` decoder the VCR cassettes need, unbreaking `test-floor`

dsfaccini:
> This pull request was posted by Claude Code using claude-fable-5 on behalf of David.

David approved this fix direction and approved adding this dependency in his own words. He has not reviewed the diff.

## Summary

The five `test on N (lowest-versions)` jobs are failing on `main` and on every open branch:

```
FAILED tests/coder/test_coder_integration.py::test_coder_completes_task
FAILED tests/researcher/test_researcher_integration.py::test_researcher_completes_task
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xc4 in position 4: invalid continuation byte
```

`brotli` was not declared anywhere in this repository. It reached the environment only through
`ddgs`, which required `httpx[brotli,http2,socks]` up to 9.14.4 and dropped its `httpx`
requirement in 9.16.0. `test-floor` deviates from the lock on purpose, so it resolves `ddgs`
fresh, picks up 9.16.0, and the decoder disappears. 25 recorded responses across those two
cassettes declare `content-encoding: br`, so httpx hands `json.loads` a still-compressed body.

## Why a test dependency and not a runtime one

httpx builds its `Accept-Encoding` header from the decoders that are actually installed
(`ACCEPT_ENCODING = ", ".join(key for key in SUPPORTED_DECODERS ...)`). Without the decoder it
advertises `gzip, deflate` and no compliant server sends brotli, so a live run never needs it.
Replay is the asymmetric case: a cassette replays what the server said when the response was
recorded, whatever the replaying client advertises. Nothing under `pydantic_ai_harness/`
imports brotli, and the only brotli-encoded bodies in the tree are in cassettes.

`httpx[brotli]` rather than a bare `brotli` because the extra is httpx's own name for this
capability, and it resolves to `brotlicffi` on non-CPython.

The entry carries a comment. Nothing imports the package, so without one it reads as an unused
dependency and is a natural candidate for deletion.

## Verification

Run in a worktree at this branch, using the `test-floor` job's own setup command:

```
UV_LOCKED=0 uv sync --resolution lowest-direct --group dev --all-extras
uv run --no-sync pytest
```

- Before: `2 failed, 5337 passed, 62 skipped`
- After: `5527 passed, 62 skipped`

The resolved environment has `ddgs 9.16.0` (the release that triggered this) with `brotli 1.2.0`
present, now from this declaration rather than transitively. `uv lock --check` passes, and the
lock diff is the two edges for the new dev-group edge and nothing else -- no version churn.

Root-cause analysis and the original reproduction on a clean `origin/main` worktree are in
[this comment on #708](https://github.com/pydantic/pydantic-ai-harness/pull/708#issuecomment-5431933075),
which is where the failure was diagnosed.

## Dependency approval

This changes `pyproject.toml` and `uv.lock`, so `dependency-approval.yml` requires the
`dependencies:approved` label. David approved the addition and the label was applied on his
behalf from his account. Note that pushing again to this branch revokes it, by design -- it
will need reapplying after any further commit.

## Checklist

- [x] Linked issue exists and is referenced above. There is no issue for this; the failure is
      diagnosed in the linked comment on #708. Happy to open one if maintainers prefer.
- [x] Tests added/updated for new behavior. No new tests: this restores two existing tests that
      the environment had broken, verified by the before/after run above.
- [x] `pre-commit` passes locally over all files, against a `uv sync --locked --all-groups
      --all-extras` environment.
- [ ] `pyproject.toml` and `uv.lock` are unchanged -- they are changed; that is the fix, and the
      label covers it.


dsfaccini:
> This comment was posted by Claude Code using claude-fable-5 on behalf of David.

Closing in favor of #715, which fixes the same test-floor break structurally: with the cassette bodies decoded and the guard test in place, no brotli decoder is needed anywhere. The analysis here supports that — httpx only advertises encodings it has decoders for, so runtime never receives br; only cassette replay ever depended on one.

PR #718, opened 2026-08-27
Support `AdvisorTool.provider_settings` across Pydantic AI versions

### LINKED ISSUE #717: Migrate `Advisor` to provider-specific native-tool settings
Pydantic AI is moving Anthropic-only `AdvisorTool.caching` into `provider_settings['anthropic']` in pydantic/pydantic-ai#7807.

`pydantic-ai-harness` should translate its existing `Advisor(caching=...)` convenience option to that nested native-tool setting when supported, while retaining compatibility with the currently supported released Pydantic AI versions. This keeps the harness API unchanged and prevents the new deprecation warning from failing the cross-repository compatibility suite.

adtyavrdhn:
## Summary

Translate the existing `Advisor(caching=...)` convenience option to `AdvisorTool.provider_settings['anthropic']` when the installed Pydantic AI version supports it. Fall back to the legacy `AdvisorTool.caching` field on currently supported releases.

This is the coordinated harness migration for pydantic/pydantic-ai#7807. The harness API and user-facing behavior do not change.

## Linked Issue

Fixes #717

## Verification

- the full advisor test module passes against the released Pydantic AI 2.33.0 floor
- the native-advisor test passes against pydantic/pydantic-ai#7807
- repository-wide Ruff and focused Pyright pass

## Checklist

- [x] Focused lint, typecheck, and tests pass locally

### DIFF
--- a/pydantic_ai_harness/advisor/_capability.py
+++ b/pydantic_ai_harness/advisor/_capability.py
@@ -6,6 +6,7 @@ from collections.abc import Awaitable, Callable
 from dataclasses import dataclass, field
 from typing import Literal
 
+import pydantic_ai.native_tools as native_tools
 from pydantic_ai import Agent
 from pydantic_ai.capabilities import ModelSelection, NativeOrLocalTool
 from pydantic_ai.exceptions import ModelRetry, UnexpectedModelBehavior, UserError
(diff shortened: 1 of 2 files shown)


mpfaffenberger (commented):
The implementation is sound and I verified the advisor suite passes locally on the released-version path (20 passed, merges clean with main). But I am holding approval on one gate from our own review checklist ('Stale Or Pre-Merge PRs'): this rides on pydantic/pydantic-ai#7807, which is still open. Until that merges, the new-version branch is dead code against every released Pydantic AI, and the `provider_settings['anthropic']['caching']` shape plus the `AnthropicAdvisorToolSettings` hasattr probe are assertions about an API that can still change in upstream review. Two asks: (1) re-confirm the shape and the feature-detection symbol against the merged upstream diff once #7807 lands, then ping for approval, which I expect to be quick; (2) non-blocking, consider probing for the actual gating attribute (e.g. `hasattr(AdvisorTool, 'provider_settings')` or a field check) rather than a sibling class name, so the detection cannot drift from the field it guards.

PR #724, opened 2026-08-28
Add `TrajectoryJudge`: review the live run with a second model on a cadence and steer it mid-run

mpfaffenberger:
Implements the narrow v1 of the background trajectory judges proposal (pydantic/pydantic-ai-notes#11), per the rulings from the 2026-08-25 pair review.

## What it does

`TrajectoryJudge` reviews the live run's most recent `window` tokens (rendered as a transcript) with a second model every `every` model requests, concurrently with the run. The judge delivers exactly one verdict per evaluation as its output type, `AllGood | Steer`; a `Steer` message is enqueued into the running conversation (`RunContext.enqueue`, `'asap'`) with attribution, so the agent course-corrects while recovery is still cheap.

```python
agent = Agent(
    'anthropic:claude-sonnet-4-6',
    capabilities=[
        TrajectoryJudge(
            model='anthropic:claude-haiku-4-5',
            instructions='Flag claims that lack evidence from files the agent actually read.',
            every=20,
        )
    ],
)
```

## Rulings baked in

- No `JudgeCouncil`: dispatch is internal; each judge is its own capability instance in `capabilities=[...]`.
- No goal-gate / `on_end` power: end-of-run verification stays with #605.
- No `allowed_actions`: v1 is a judge that watches every N requests over a sliding token window and can steer mid-run.
- Steer is the judge's output, a union verdict type (`AllGood | Steer`), not a tool the judge calls.
- Advanced customization: pass a full `Agent` (typed `Agent[None, AllGood | Steer]`, so the verdict contract is checker-enforced) instead of piling knobs onto the capability.

## Mechanics

- Cadence counted in `after_model_request`; the evaluation runs as a background task, so the run is never blocked on a judge. At most one evaluation per judge in flight; a due tick that finds one running is skipped.
- Judge usage threads onto the run's `usage` and runs under `usage_limits` minus one reserved request.
- An evaluation failure is raised on the run at the next cadence tick or run end (provider failures propagate as their original `ModelAPIError` subclasses); resilience policy belongs to a passed judge `agent` (e.g. `FallbackModel`), not to fields on the capability.
- Run end cancels an in-flight evaluation; a failing run discards the evaluation entirely so it cannot mask the run's own error.

## Verification

- `tests/trajectory_judge/`: 20 tests through the public `Agent` path plus direct hook tests for the lifecycle branches; 100% branch coverage on the new module and the test file.
[...]

PR #725, opened 2026-08-28
Re-export `MCPToolsetClient` from `pydantic_ai_harness.stackone._toolset` so CI lint passes again

mpfaffenberger:
CI's lint job runs the pre-commit pyright hook against the latest pydantic-ai rather than the lockfile, and it now flags `stackone/_capability.py`'s import of `MCPToolsetClient` from `_toolset` as `reportPrivateImportUsage`. That fails lint on every PR (first seen on #724; main at `6daf761` was green before the drift).

Declaring `MCPToolsetClient` in `_toolset.__all__` makes the re-export explicit, which satisfies pyright under both the old and new pydantic-ai. `_capability` keeps importing it from `_toolset`, so the guarded import there still turns a missing `mcp` extra into the install-hint `ImportError`.

No behavior change; one line plus a comment.

### DIFF
--- a/pydantic_ai_harness/stackone/_toolset.py
+++ b/pydantic_ai_harness/stackone/_toolset.py
@@ -39,9 +39,13 @@ except ImportError as _import_error:  # pragma: no cover
         'MCP support is required for the StackOne capability. Install it with: uv add "pydantic-ai-harness[stackone]"'
     ) from _import_error
 
+# `MCPToolsetClient` is re-exported for `_capability`, which must not import `pydantic_ai.mcp`
+# directly: this module's guarded import is what turns a missing `mcp` extra into the
+# install-hint `ImportError`.
 __all__ = (
     'STACKONE_API_KEY_ENV',
     'STACKONE_BASE_URL',
+    'MCPToolsetClient',
     'StackOneToolset',
     'ToolMode',
 )


adtyavrdhn (approved):
Alright with a reexport

dsfaccini:
I wouldn't merge this unless we can link to a place where this error is currently happening, it's not clear to me what this is fixing

aside from that, notes on variable naming and re-exporting symbols:
- no _private vars in _private modules
- we should try to maintain `__all__` in public modules, to prevent accidentally re-exporting variables imported from private modules

e.g.

```python
# harness/public_module.py
from harness._private import normal_variable

# if there's no __all__ people may be able to import `from harness.public_module import normal_variable`
```

IIRC this is in `agent_docs`

mpfaffenberger:
> I wouldn't merge this unless we can link to a place where this error is currently happening, it's not clear to me what this is fixing
> 
> aside from that, notes on variable naming and re-exporting symbols:
> 
>     * no _private vars in _private modules
> 
>     * we should try to maintain `__all__` in public modules, to prevent accidentally re-exporting variables imported from private modules
> 
> 
> e.g.
> 
> ```python
> # harness/public_module.py
> from harness._private import normal_variable
> 
> # if there's no __all__ people may be able to import `from harness.public_module import normal_variable`
> ```
> 
> IIRC this is in `agent_docs`

Hey @dsfaccini - Aditya is okay using `__ALL__` in private modules. You have strong feelings still?

dsfaccini:
no strong feelings on using `__all__` in private modules (I don't see the point of using it but I don't mind)

not sure if that misunderstands my previous message though, which is about taking care that we don't accidentally re-export originally private symbols from public modules

PR #727, opened 2026-08-28
Spill: preserve non-mapping ToolReturn.metadata instead of discarding it

### LINKED ISSUE #702: `Spill` replaces non-mapping `ToolReturn.metadata`
### Capability

ToolOutputLimits

### Bug Description

When a `Spill` succeeds, `ToolOutputLimits` replaces existing `ToolReturn.metadata` unless it is a `Mapping`. For example, a Pydantic model is discarded and the resulting `ToolReturnPart.metadata` contains only `overflow_handle` and `overflow_bytes`.

Pydantic AI types `ToolReturn.metadata` as `Any` and documents it as application-accessible data that is not sent to the model ([ref](https://pydantic.dev/docs/ai/api/pydantic-ai/messages/#pydantic_ai.messages.ToolReturn)).

The current implementation copies existing metadata only when it is a `Mapping`, and an existing unit test explicitly asserts replacement of non-mapping metadata. I could not find this restriction documented by `ToolOutputLimits`.

I expected a successful spill not to discard caller-owned application metadata given that it's not model-facing (i.e., that the original metadata value remains available rather than being silently replaced).

### Minimal Reproduction

```python
from pathlib import Path
from tempfile import TemporaryDirectory

from pydantic import BaseModel
from pydantic_ai import Agent, ToolReturn
from pydantic_ai.messages import ToolReturnPart
from pydantic_ai.models.test import TestModel

from pydantic_ai_harness import ToolOutputLimits
from pydantic_ai_harness.tool_output_limits import Band, LocalFileStore, Spill


class QueryStats(BaseModel):
    match_count: int


with TemporaryDirectory() as directory:
    agent = Agent(
        TestModel(call_tools=['run_query']),
        capabilities=[
            ToolOutputLimits(
                bands=[Band(over=10, action=Spill())],
                store=LocalFileStore(base_dir=Path(directory)),
            )
        ],
    )

    @agent.tool_plain
    def run_query() -> ToolReturn:
        return ToolReturn(
            return_value='x' * 100,
            metadata=QueryStats(match_count=12345),
        )

    result = agent.run_sync('go')
    part = next(
        part
        for message in result.all_messages()
        for part in message.parts
        if isinstance(part, ToolReturnPart) and part.tool_name == 'run_query'
    )

    print(part.metadata)
    # Actual: {'overflow_handle': '...', 'overflow_bytes': 100}
    # Expected: the QueryStats value remains available.
```

0.25.0

### pydantic-ai version

2.35.0

AmirF194:
## Summary

`_with_handles` rebuilds `ToolReturn.metadata` when a `Spill` fires, but it only copies the existing value in when it is a `Mapping`. `ToolReturn.metadata` is typed `Any` and documented as opaque, app-only data, so a non-mapping value (a Pydantic model, a plain string, any object) is silently replaced by `{'overflow_handle': ..., 'overflow_bytes': ...}` with no trace of the original.

Fix: when the existing metadata is present and not a mapping, nest it under `original_metadata` instead of dropping it. A mapping is still merged in directly (unchanged behavior); `None` still produces the bare overflow dict (unchanged behavior).

## Boundary notes (output-limit)

`ToolReturn.metadata` is never sent to the model (confirmed against pydantic-ai's own docstring: "Additional data accessible by the application but not sent to the LLM"), so nesting the original value under a new key changes only application-facing data, not anything model-visible. The spill preview text, `overflow_handle`/`overflow_bytes` byte accounting, and the read-back path (`read_tool_result`, which takes the handle as an explicit argument, never from metadata) are all unchanged; only the previously-discarded branch is affected.

## Linked Issue

Fixes #702

## How I verified

- Reproduced end-to-end against current `main` (`ToolReturn(return_value=..., metadata=<non-mapping>)` through `ToolOutputLimits` with a triggering `Spill` band): the returned metadata was `{'overflow_handle': ..., 'overflow_bytes': ...}`, the original value gone.
- Added a regression test at the same level as the existing `test_spill_merges_existing_metadata` (`test_spill_preserves_non_mapping_metadata`), plus a unit test on `_with_handles` itself; both fail on `main` and pass on this branch. Updated the one existing test that had pinned the old (buggy) behavior.
- Full `tests/tool_output_limits/` suite (101 tests) and the full repo suite (`uv run coverage run -m pytest`, 4588 passed / 71 skipped) pass; the changed lines are covered.
- `ruff format --check`, `ruff check` clean. `pyright` on the two changed files is clean (0 errors); a whole-repo `pyright` run reports errors, but the identical count and locations (all in `tests/youdotcom/`, unrelated to this change) also occur on an unmodified `main` checkout in the same environment, so I could not confirm the whole-repo pyright job is clean in CI's exact setup, only that this diff introduces no new pyright errors.


pydantic_ai_harness/tool_output_limits/_capability.py
```diff
@@ -502,6 +502,8 @@ def _with_handles(
     base: dict[str, object] = {}
     if _is_mapping(existing):
         base.update(_copy_mapping(existing))
+    elif existing is not None:
+        base['original_metadata'] = existing
```
mpfaffenberger, line 506: This makes `original_metadata` an application-facing key and the only way callers can recover opaque metadata after a spill. Please document the mapping and non-mapping behavior, including this key, in both `pydantic_ai_harness/tool_output_limits/README.md` and `docs/tool-output-limits.md`. Otherwise we are shipping a new contract that users can discover only from source or tests.

tests/tool_output_limits/test_tool_output_limits.py
```diff
@@ -558,6 +558,14 @@ async def test_spill_merges_existing_metadata(self, tmp_path: Path):
         assert out.metadata['orig'] is True
         assert 'overflow_handle' in out.metadata
 
+    async def test_spill_preserves_non_mapping_metadata(self, tmp_path: Path):
+        store = LocalFileStore(base_dir=tmp_path)
+        cap: ToolOutputLimits[object] = ToolOutputLimits(bands=[Band(over=5, action=Spill())], store=store)
+        out = await _run(cap, ToolReturn(return_value='a' * 100, metadata='app-request-id-123'))
```
mpfaffenberger, line 564: The reported contract is the `ToolReturnPart.metadata` applications observe after `Agent.run`, but `_run` stops at a direct `after_tool_execute` call. Please cover this through `Agent(..., capabilities=[cap])` under `TestAgentIntegration`, returning a `ToolReturn` with non-mapping metadata and asserting that the value survives under `original_metadata` in message history. I ran that exact public path locally and it passes, so this is a coverage change rather than an implementation change.

mpfaffenberger (dismissed):
Hey Amir, thank you for the focused fix. The implementation preserves the original object correctly, and I reproduced the full Agent path locally. I found two contract and coverage gaps to close before merge.

AmirF194:
Thanks for testing this so thoroughly. Pushed both: documented `original_metadata` next to the existing spill-key section in both READMEs, and added `TestAgentIntegration.test_spill_preserves_non_mapping_metadata_through_agent_run`, which drives `Agent.run` end to end and asserts the value survives in message history. Confirmed it fails against the pre-fix commit and passes on this branch.

AmirF194:
Pushed 7493e678 covering both gaps (original_metadata key documented, Agent.run path added to TestAgentIntegration) about 10 days ago. Anything else you'd want changed before another look, or is this ready to merge as is?

PR #729, opened 2026-08-29
Give `ToolOutputLimits` summaries a stream seam

### LINKED ISSUE #687: Give `ToolOutputLimits` summaries a stream seam and stop swallowing summarizer provider errors

### Capability

Tool Output Limits (`ToolOutputLimits`)

### Bug Description

`ToolOutputLimits._summarize` runs a nested `Agent.run(...)` with no way to select the streaming
request path, and `_summarize_action` turns the resulting provider error into a silent fallback.

Two things compound:

1. **No stream seam.** `_summarize` builds `Agent(model, instructions='You summarize oversized tool
   output.')` and calls `agent.run(prompt, usage=ctx.usage, usage_limits=...)` with no
   `event_stream_handler`, so core takes the non-streaming request path. `ModelSettings` carries no
   `stream` key, and the nested `Agent` is constructed with no `capabilities=`, so there is no
   supported way to change that from the outside.
2. **The failure is silent.** `_summarize_action` re-raises `UserError` (deliberately, for
   misconfiguration) but catches `Exception` and returns `self._fallback(...)`. A provider that
   rejects non-streaming requests therefore does not error -- the oversized output is truncated
   instead of summarized, on every call, with nothing surfaced to the caller.

An endpoint that requires `stream: true` is not hypothetical:
`https://chatgpt.com/backend-api/codex/responses` returns HTTP 400 `"Stream must be set to true"`
on `stream:false` ([openclaw/openclaw#67740](https://github.com/openclaw/openclaw/issues/67740)).
The inverse also exists -- some Azure deployments reject `stream:true` with `Unsupported value:
'stream' does not support 'true' with this model`
([Microsoft Q&A](https://learn.microsoft.com/en-us/answers/questions/2184105/bug-report-azure-open-ai-assistant-api-streaming-s))
-- so the transport has to stay a caller's choice, not a new default.

This is the same defect class as #619, which #620 fixes for `SummarizingCompaction` by adding an
`event_stream_handler` field to the capability (matching the existing `SubAgents.event_stream_handler`
precedent) plus a `drain_summary_events` handler for callers who need the transport but not the
events. `ToolOutputLimits` is the second nested-summarizer site and was left out of that PR's scope.

Suggested shape, for consistency with what #620 lands: an `event_stream_handler` on the `Summarize`
action, threaded into the nested `agent.run(...)`. Whether the silent-fallback half should also
change (for example, letting a permanent provider rejection surface rather than truncating on every
call) is worth deciding separately -- the fallback is intentional for transient summarizer failures.

Found while reviewing #620; filed rather than folded in, since it is pre-existing and out of that
PR's scope.

### Minimal Reproduction

```python
from pydantic_ai import Agent
from pydantic_ai_harness.tool_output_limits import Summarize, ToolOutputLimits

# Any summarizer model whose endpoint rejects non-streaming requests.
agent = Agent(
    'openai:gpt-5.6-terra',
    capabilities=[
        ToolOutputLimits(
            default=Summarize(model='provider:stream-only-summary-model'),
            over_tokens=1_000,
        )
    ],
)

# An oversized tool result reaches `_summarize`, the provider rejects the
# non-streaming request, `_summarize_action` catches it, and the output is
# truncated instead. No error reaches the caller.
await agent.run('call the tool that returns a large payload')
```

main (`ac12341c`)

### pydantic-ai version

2.28.0

xixi-box:
Fixes #687.

`ToolOutputLimits` summarizes oversized tool returns through a nested `Agent.run(...)`. That call never received an `event_stream_handler`, so the summary request always took the non-streaming request path, and there was no supported way to change that: `ModelSettings` carries no `stream` key and the nested `Agent` accepts no `capabilities=`. On a summarizer endpoint that rejects non-streaming requests (for example the Codex endpoint's HTTP 400 `Stream must be set to true`), `_summarize_action` catches the provider error and degrades to the `then` fallback, so the output is truncated on every call with nothing surfaced.

This is the same defect class as #619, which #620 fixes for `SummarizingCompaction`. This PR applies the same API decisions to the second nested-summarizer site:

- `Summarize` gains `event_stream_handler: EventStreamHandler[object] | None = None`, threaded into the nested `agent.run(...)`. The `object` parameter follows #620's resolution: the handler receives the summary run's own `RunContext`, which carries no caller deps. Supplying any handler selects the streaming request path; left `None` the request stays non-streaming, which is what endpoints that reject streaming requests need. The transport remains a caller's choice, not a new default.
- `drain_summary_events` ships as the ready-made handler for callers who need the streaming transport but not the events, exported from `pydantic_ai_harness.tool_output_limits`. Naming follows #620's helper.
- Ignored when a `summarize` callable is set, since a custom callable performs no model request.

The silent-fallback half of #687 (whether a permanent provider rejection should surface rather than truncate) is deliberately left unchanged: the issue flags it as a separate decision, and the fallback is intentional for transient summarizer failures.

## Tests

- A stream-only `FunctionModel` (no `function`, so a non-streaming request fails outright) falls back through `then` without a handler, and produces the summary with `drain_summary_events`.
- A caller-supplied handler observes the summary deltas as they stream.
- An end-to-end run through `Agent(..., capabilities=[ToolOutputLimits(...)])` on `TestModel`: an oversized tool return is summarized over the streaming path inside the parent run.
- The existing usage-limits test now also asserts the nested run receives `event_stream_handler=None` by default.

## Docs

`pydantic_ai_harness/tool_output_limits/README.md` and `docs/tool-output-limits.md` both gain a "Summary streaming" section, kept in parity.

## Coordination with #620

#620 is still open and this PR is independent of it. Both add a `drain_summary_events` to their own package (this one in `pydantic_ai_harness.tool_output_limits`, #620's in `pydantic_ai_harness.compaction`). If #620 lands first and a single shared home is preferred, happy to consolidate.

### DIFF
--- a/docs/tool-output-limits.md
+++ b/docs/tool-output-limits.md
@@ -240,6 +240,32 @@ instance to `Summarize(model=...)` to override, or a `summarize` callable to byp
 built-in prompt entirely. The `summary_prompt` template on the capability must contain both
 `{tool_name}` and `{output}` placeholders.
 
+## Summary streaming
+
+The summary request is non-streaming unless `event_stream_handler` is set on the `Summarize`
+action. Supply a handler to watch the summary as it is written, or pass `drain_summary_events`
+to take the streaming request path without handling the events -- which is what a summarizer
+endpoint that rejects non-streaming requests needs:
+
+```python
+from pydantic_ai import Agent
+from pydantic_ai_harness.tool_output_limits import Band, Summarize, ToolOutputLimits, drain_summary_events
+
+agent = Agent(
+    'openai:gpt-4o',
+    capabilities=[
+        ToolOutputLimits(
+            bands=[Band(over=20_000, action=Summarize(event_stream_handler=drain_summary_events))],
+        )
+    ],
+)
+```
+
+Neither transport works everywhere, which is why this is a choice rather than a default: some
+endpoints reject non-streaming requests and others reject streaming ones. The handler receives
+the summary run's own `RunContext` and event stream; the outer `Agent.run(...)` handler is not
+inherited and never sees the summary token deltas.
+
 ## Edge cases
 
 - Binary returns spill verbatim and are never stringify-truncated; `Truncate` / `Summarize`
--- a/pydantic_ai_harness/tool_output_limits/README.md
+++ b/pydantic_ai_harness/tool_output_limits/README.md
@@ -207,6 +207,32 @@ instance to `Summarize(model=...)` to override, or a `summarize` callable to byp
 built-in prompt entirely. The `summary_prompt` template on the capability must contain both
 `{tool_name}` and `{output}` placeholders.
 
+## Summary streaming
+
+The summary request is non-streaming unless `event_stream_handler` is set on the `Summarize`
+action. Supply a handler to watch the summary as it is written, or pass `drain_summary_events`
+to take the streaming request path without handling the events -- which is what a summarizer
+endpoint that rejects non-streaming requests needs:
+
+```python
+from pydantic_ai import Agent
+from pydantic_ai_harness.tool_output_limits import Band, Summarize, ToolOutputLimits, drain_summary_events
+
+agent = Agent(
+    'openai:gpt-4o',
+    capabilities=[
+        ToolOutputLimits(
+            bands=[Band(over=20_000, action=Summarize(event_stream_handler=drain_summary_events))],
+        )
+    ],
+)
+```
+
+Neither transport works everywhere, which is why this is a choice rather than a default: some
+endpoints reject non-streaming requests and others reject streaming ones. The handler receives
+the summary run's own `RunContext` and event stream; the outer `Agent.run(...)` handler is not
+inherited and never sees the summary token deltas.
+
 ## Edge cases
 
 - Binary returns spill verbatim and are never stringify-truncated; `Truncate` / `Summarize`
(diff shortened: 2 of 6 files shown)


xixi-box:
Merged current `main` into the branch and resolved the three conflicts:

- `_summarize` is now a journaled durable operation that recovers the `Summarize` action from its path on the worker side. `event_stream_handler=action.event_stream_handler` threads through that resolved action, so the seam works identically on the caller and durable-worker paths.
- The README and docs page keep both the new durability paragraphs and the "Summary streaming" section.
- The no-handler fallback test now uses the same 95-char truncation budget as the adjacent custom-summarizer test, since complete-tail-lines truncation no longer appends a marker at a 10-char budget.

On the red checks from the earlier run: `test on latest pydantic-ai` failed in `tests/coder/test_coder_integration.py::test_coder_completes_task` with `ImportError: Please install 'anthropic'` -- a module this PR never touches; current `main` passes that job (run 35004621384, 6388 passed). `check` is the alls-green gate echoing that failure.

One coordination note now that #620 has landed: `drain_summary_events` exists in both `pydantic_ai_harness.compaction` and `pydantic_ai_harness.tool_output_limits`, matching each package's own exports. Happy to consolidate into a shared home if preferred.

xixi-box:
A quick status update: this branch was brought up to date with current `main` on Sep 16 (the three conflicts noted above are resolved) and is ready for review. Happy to rebase again or adjust the implementation if the approach should change — just let me know.

PR #734, opened 2026-08-31
Add `ModalWorkspace` for agent workspace operations

adtyavrdhn:
## Summary

Adds `ModalWorkspace` and `ModalWorkspaceBackend` for agent tools using `ctx.workspace`. Construction performs no remote work. First use creates a workspace or attaches to the exact supplied reference, and caches the typed SDK handle. `await backend.workspace` exposes that `modal.Sandbox` for application-owned SDK operations.

Message history carries the reference into later runs. Runs without a reference create independent workspaces; names are creation options. The application owns durable coordination and remote cleanup. The existing shipped `ModalSandbox` integration remains unchanged.

Boundary notes: command startup respects the caller's timeout; Modal applies whole-second execution deadlines and has no per-command kill operation. Result collection retains captured output on timeout and drains reader tasks on errors or cancellation. Filesystem operations use Modal's native API and preserve workspace paths, including trailing whitespace.

Validation: focused new and existing Modal tests, scoped Pyright, full Ruff, and paired documentation/snippet checks pass. Live Modal checks passed for stdout/stderr/nonzero exit, filesystem roundtrip, execution timeout output, and native termination/detachment. Fake-backed tests cover lazy acquisition, exact-reference history reuse, concurrent operations, failure/cancellation retry, and error translation.

## Linked Issue

Implements the workspace design in pydantic/pydantic-ai#6492. Stacked on #522; review this layer's diff. The temporary core pin is `db57de423ca76e422037a43521ca7a3075cd6b4d` and must become a released dependency before merge. E2B #735 builds on this layer.

The harness page must reach harness `main` before the companion core navigation entry: the unified documentation build reads harness pages from `main`.

## Checklist

- [x] Linked design and predecessor are referenced above
- [x] Focused tests and type checks, full Ruff, and documentation checks pass locally; CI runs repository-wide checks
- [ ] A maintainer added `dependencies:approved` to the current head


adtyavrdhn:
@douwebot review

adtyavrdhn:
@douwebot review

pydantic_ai_harness/modal_sandbox/_backend.py
```diff
@@ -0,0 +1,947 @@
...
+        return exit_code == _SIGKILL_EXIT and elapsed >= self._deadline
+
+    async def kill(self) -> None:
+        raise NotImplementedError(
+            'Modal exposes no way to kill an individual command; start it with `timeout=` so the '
+            'platform kills it at the deadline, or terminate the whole sandbox.'
+        )
+
+
+class _ModalFilesystem:
```
adtyavrdhn: Why private

pydantic_ai_harness/modal_sandbox/_backend.py
```diff
@@ -0,0 +1,947 @@
...
+    """Modal's sandbox filesystem API behind the `SandboxFilesystem` protocol."""
+
+    def __init__(self, sandbox: modal.Sandbox, classify: Callable[[Exception], Awaitable[ModalSandboxError]]) -> None:
+        self._sandbox = sandbox
+        self._classify = classify
+
+    @asynccontextmanager
+    async def _translated(self, path: str) -> AsyncGenerator[None]:
```
adtyavrdhn: Again, why private. I did not expect us to need this, should we design around it in the protocol? Is this even necessary

pydantic_ai_harness/modal_sandbox/_backend.py
```diff
@@ -0,0 +1,947 @@
...
+
+def _absolute_path(name: str, value: str | None) -> str | None:
+    if value is None:
+        return None
+    if not posixpath.isabs(value):
+        raise ValueError(f'{name} must be an absolute sandbox path or None, got {value!r}.')
+    return value
+
+
+async def _create_on_asyncio(
```
adtyavrdhn: This feels like a complete mess, what is even the point of doing this. How does this help when there is an event loop why can't we just do it with a fail_after?

pydantic_ai_harness/modal_sandbox/_backend.py
```diff
@@ -0,0 +1,947 @@
...
+        arrival = itertools.count()
+
+        async def read_one(name: _Stream) -> tuple[int, _Stream, bytes | None]:
+            chunk = await anext(iterators[name], None)
+            # Stamped where the chunk lands rather than where it is collected: a wake-up that
+            # finds both streams ready hands them back as an unordered set, and this is what
+            # keeps the merge in the order Modal produced the output.
+            return next(arrival), name, chunk
+
+        pending = {asyncio.ensure_future(read_one(name)) for name in iterators}
```
adtyavrdhn: Do we have scenarios where this could be anything other than a coroutine? Seems unlikely

pydantic_ai_harness/modal_sandbox/_backend.py
```diff
@@ -0,0 +1,947 @@
...
+                    pending.add(asyncio.ensure_future(read_one(name)))
+                    text = decoders[name].decode(chunk)
+                    if text:
+                        yield _ModalOutputChunk(stream=name, data=text)
+        finally:
+            # Reaped, not just cancelled: an abandoned read would otherwise keep retrying
+            # against the worker long after the consumer walked away.
+            for task in pending:
+                task.cancel()
+            await asyncio.gather(*pending, return_exceptions=True)
```
adtyavrdhn: Why do we have cancel scopes in certain places but not all cleanups

pydantic_ai_harness/modal_sandbox/_backend.py
```diff
@@ -0,0 +1,947 @@
...
+            name: Optional Modal name, unique among running sandboxes in the app.
+        """
+        try:
+            import modal
+        except ImportError as e:
+            raise ModalSandboxError(_MISSING_MODAL) from e
+
+        workdir = _absolute_path('workdir', workdir)
+
+        async def create_sandbox() -> modal.Sandbox:
```
adtyavrdhn: We know modal provides a sync API to create the sandbox but it would block the event loop?

pydantic_ai_harness/modal_sandbox/_backend.py
```diff
@@ -0,0 +1,947 @@
...
+            await cls(created, working_dir=workdir, sandbox_timeout=sandbox_timeout).close(terminate=True)
+
+        sandbox: modal.Sandbox | None = None
+        try:
+            with anyio.CancelScope(shield=True):
+                try:
+                    asyncio.get_running_loop()
+                except RuntimeError:
+                    with anyio.move_on_after(_CREATE_TIMEOUT):
+                        sandbox = await create_sandbox()
```
adtyavrdhn: How can we reach this scenario? If there isn't an event loop in Pydantic AI app then we create one and run the app on loop.run_until_complete which should have already happened by now

pydantic_ai_harness/modal_sandbox/_backend.py
```diff
@@ -0,0 +1,947 @@
...
+                else:
+                    try:
+                        sandbox = await _create_on_asyncio(create_sandbox, cleanup_created)
```
adtyavrdhn: I don't like this?

pydantic_ai_harness/modal_sandbox/_backend.py
```diff
@@ -0,0 +1,947 @@
...
+        except modal.exception.Error as e:
+            raise ModalSandboxError(f'Could not start Modal sandbox: {e}') from e
+        if sandbox is None:
+            raise ModalSandboxError(
+                f'Modal sandbox creation did not complete within {_CREATE_TIMEOUT}s; '
+                'the Modal control plane may be unreachable.'
+            )
+        backend = cls(sandbox, working_dir=workdir, sandbox_timeout=sandbox_timeout)
+        try:
+            await anyio.lowlevel.checkpoint()
```
adtyavrdhn: It would be worthwhile to check why we need it, if we do at all there should be code comments explaining concisely what is the point of this?

pydantic_ai_harness/modal_sandbox/_backend.py
adtyavrdhn: Removed the filesystem wrapper. Native file operations now live directly on ModalSandboxBackend, which implements SupportsFilesystem.

pydantic_ai_harness/modal_sandbox/_backend.py
adtyavrdhn: The separate filesystem object and callback layer are gone. Core now discovers the backend's native filesystem methods directly through SupportsFilesystem.

pydantic_ai_harness/modal_sandbox/_backend.py
adtyavrdhn: Removed the async creation bridge. Creation now awaits Modal's .aio API inside anyio.fail_after; no extra event loop or synchronous fallback remains.

pydantic_ai_harness/modal_sandbox/_backend.py
adtyavrdhn: The ensure_future/streaming wrapper is gone. The one-shot run path awaits SDK coroutines directly and collects stdout, stderr, and exit status.

pydantic_ai_harness/modal_sandbox/_backend.py
adtyavrdhn: Provider teardown RPCs are bounded and shielded, including detach after a failed terminate. Command cancellation cancels and drains local reader tasks; Modal has no per-command kill, so the remote deadline limitation is documented. Cleanup retains the original SDK cause.

pydantic_ai_harness/modal_sandbox/_backend.py
adtyavrdhn: Creation uses await modal.Sandbox.create.aio(...). There is no synchronous creation call on the agent event loop.

pydantic_ai_harness/modal_sandbox/_backend.py
adtyavrdhn: Removed the no-running-loop fallback and the creation bridge. The backend performs I/O only from its async operations.

pydantic_ai_harness/modal_sandbox/_backend.py
adtyavrdhn: Removed the extra creation machinery and the private process wrapper. run() is now the direct one-shot command path.

pydantic_ai_harness/modal_sandbox/_backend.py
adtyavrdhn: Removed the post-create cancellation checkpoint and its extra cleanup path. The remaining bounded cleanup documents why shielding is needed.

DouweM:
## Review notes for the three provider PRs: this one, #735, #736 (Douwe, written up by Claude)

**Status (2026-09-22 morning):** items 1, 2 and 6 are done on all three PRs (`get_client()`, durable wording replaced by "Lifetimes and cleanup", `E2BSandbox`/`DaytonaSandbox` renames, and the `ModalSandbox` deprecate-in-place bridge — see the follow-up comment below). Item 3's "dead ref raises `WorkspaceUnavailableError`" and "ref after first op" are now covered by tests per backend; the live provider runs are still open. Items 4 and 5 are still open. CI is green except the known Windows job. One thing to settle: the Modal upgrade text tells users to add `Shell()`/`FileSystem()` for tools, which is only true once #761 lands, so #761 should merge before or with this stack.

I merged the updated base up the stack (`1c9af955` here, `d89c1288` on #735, `737fd1bb` on #736). The `coverage` red on all three is one gap inherited from my first #522 merge (already fixed there in `3c68b832`) and clears on the next base merge; the Windows image-paste failure is the core `LocalWorkspace` POSIX-only issue noted on #522.

1. **Native handle naming.** `await backend.workspace` mixes our "workspace" word with the provider's object, and an awaitable property reads badly. On Sep 8 we settled on a lazy `get_client()` method; please use that on all three (`await backend.get_client()`), returning the provider's typed object.
2. **"The application owns durable coordination" has to go** from the docs and PR text. That is not the expectation we set anywhere else; core is getting durable handling of `ctx.workspace` back (see pydantic/pydantic-ai#6492). What stays the application's job is deleting environments.
3. **Live runs.** #735 says no live E2B calls were run and #736 says no live Daytona provisioning. Before merge each provider needs a recorded or manually verified live pass: create, reattach from a `WorkspaceRef`, command + file round trip, timeout, and **reattach to a deleted/expired environment raising `WorkspaceUnavailableError`** rather than silently creating a fresh one (our Sep 3 ruling; I have not verified the three backends do this).
4. **Fresh-create races across processes.** The in-backend lock does not span Temporal activities, so two parallel activities with no ref would create two environments. Core will force one first use from the workflow, but a backend that can report a name-based ref before any I/O (Daytona's conversation-derived names nearly do) avoids the problem entirely; worth doing where the provider supports names.
5. **#736's `destroy()`/`stop()`/`pause()`/`disconnect()`**: fine as provider-specific extras on the concrete backend, but let's agree on which names we would want every provider to share before three backends each pick their own.
6. **Capability names.** On Sep 10 we settled that the concept is a workspace but the provider capabilities keep "sandbox" in the name (`ModalSandbox`, `E2BSandbox`, `DaytonaSandbox`), because an agent definition should say at a glance that the agent is confined, and `LocalWorkspace` correctly implies the opposite. These PRs ship `ModalWorkspace`/`E2BWorkspace`/`DaytonaWorkspace`. Same call: `ModalSandbox` is deprecated in place (same namespace, bridging old and new, warning on legacy-only constructor arguments) rather than left unchanged next to a new class. Unless something changed after Sep 10 that I've lost track of, these need to follow that.

DouweM:
Per the Sep 10 naming ruling (the concept is a workspace, a provider capability is a sandbox), the stack now ships `ModalSandbox` / `ModalSandboxBackend` here, and `E2BSandbox` / `E2BSandboxBackend` (#735) and `DaytonaSandbox` / `DaytonaSandboxBackend` (#736) above it, in packages `modal_sandbox`, `e2b_sandbox`, and `daytona_sandbox`.

Modal keeps its shipped name and namespace: the workspace-backed implementation takes over `ModalSandbox` in `pydantic_ai_harness.modal_sandbox` (`0f7c635f`), replacing the previous capability that bundled its own `run_command` / `read_file` / `write_file` / `list_directory` tools and terminated its sandbox. Passing one of that capability's constructor arguments (`sandbox_id`, `session`, `default_command_timeout`, `max_command_timeout`, `max_output_bytes`, `max_output_lines`, `max_read_bytes`, `instructions`) raises a `UserError` that says how to express each one now (`Shell` / `FileSystem` options, `workspace=WorkspaceRef(...)` on the run, `ModalSandboxBackend(workspace=...)`), and the docs page and README carry the same guidance under "Upgrading from the previous `ModalSandbox`". `ModalSandboxSession` and the `ModalSandbox*Error` classes are gone with it; a bare `ModalSandbox()` keeps working, now without tools of its own.

— Douwe, written up by Claude

PR #735, opened 2026-08-31
Add `E2BWorkspace` for agent workspace operations

adtyavrdhn:
## Summary

Adds `E2BWorkspace` and `E2BWorkspaceBackend` for agent tools using `ctx.workspace`. Construction performs no remote work. First use creates a workspace or attaches to the exact supplied reference and caches the typed SDK handle. Applications can await `backend.workspace` to use the native `e2b.AsyncSandbox` directly.

Message history carries the reference into later runs. Runs without a reference create independent workspaces. Metadata is ordinary creation metadata; the application owns durable coordination, reference persistence, and remote cleanup.

Boundary notes: E2B runs shell-quoted argument lists through its login shell. One command timeout covers acquisition, startup, and result collection. Timeout, cancellation, or result failure triggers bounded best-effort command PID cleanup while preserving SDK output and the original error. Native filesystem operations preserve canonical workspace paths, including trailing spaces. Connecting to a paused environment resumes it using E2B's default connection lifetime.

Validation: focused E2B, composition, and documentation tests pass, alongside scoped Pyright and full Ruff. Public tests cover lazy acquisition, exact-reference history reuse, concurrent operations, failure/cancellation retry, and preservation of a supplied read-only workspace. The three documentation wrappers have executed unchanged with the SDK fake and a deterministic model, covering history reattachment, explicit-reference file access, and native cleanup. No live E2B service calls have been run.

## Linked Issue

Implements the workspace design in pydantic/pydantic-ai#6492, related to #509. Stacked on #734; review this layer's diff. The temporary core pin is `db57de423ca76e422037a43521ca7a3075cd6b4d` and must become a released dependency before merge. Daytona #736 builds on this layer.

The harness page must reach harness `main` before the companion core navigation entry: the unified documentation build reads harness pages from `main`.

## Checklist

- [x] Linked design, issue, and predecessor are referenced above
- [x] Tests added/updated for new behavior
- [x] Focused tests and type checks, full Ruff, and documentation checks pass locally; CI runs repository-wide checks
- [ ] A maintainer added `dependencies:approved` to the current head
- [x] Docstrings use single backticks

PR #736, opened 2026-08-31
Add `DaytonaWorkspace` for agent workspace operations

### LINKED ISSUE #683: Add a Daytona sandbox capability
Pydantic AI agents need an optional Daytona-backed sandbox for bounded command and file work.

The capability should own fresh-sandbox cleanup, support caller-owned sessions, bound command output while streaming it, and expose a managed bidirectional process seam without leaking the raw SDK sandbox.

The Daytona dependency remains optional and pinned to the verified SDK range.

adtyavrdhn:
Stacked on #735 (E2B), above #734 (Modal) and #522. Review only this layer's diff.

Targets pydantic/pydantic-ai#6492 at `24b51c3f37dd7f4a9dacd2f33ad4a14c6b3722be`. The temporary core pin must be replaced by a released dependency before merge.

Adds `DaytonaSandbox` and `DaytonaSandboxBackend` for `ctx.sandbox`. Construction does no I/O; first use creates or attaches, and `await backend.sandbox` returns the native SDK sandbox. Conversation-derived names let follow-up runs reuse the workspace.

- Uses a dataclass for per-command session state and one deadline covering acquisition, session setup, output, and exit status.
- Applies the configured working directory to commands and resolves its canonical path inside Daytona.
- Preserves terminal errors and original SDK causes. Cleanup cancels local log tasks even when session deletion fails; sandbox deletion retries obtain a handle from the fresh client.
- Updates both documentation copies and adds SDK-fake Agent lifecycle coverage.

Runs do not perform cleanup. Created sandboxes stop after the configured idle interval and retain their disk instead of being deleted. `destroy()` explicitly deletes the sandbox, including attached references; `disconnect()` releases the backend-owned SDK client. `pause()` uses Daytona's supported VM suspension, and `stop()` preserves disk while rejecting ephemeral sandboxes that would be deleted on stop. Lifecycle calls do not create or resume a sandbox for cleanup; callers finish active operations first.

Validation uses SDK fakes and scoped Pyright. No live Daytona provisioning was performed.

Fixes #683.

The backend implements `create_or_attach()` and inherits core's `LazySandbox`: shared acquisition locking and caching, with `await self.sandbox` in operations. No provider-local acquisition lock or resolver is needed. Core tests cover concurrent first use, cached reuse, failure retry, cancelled acquisition, and cancelled waiters.


adtyavrdhn:
@douwebot review

adtyavrdhn:
@douwebot review

pydantic_ai_harness/daytona_sandbox/_backend.py
```diff
@@ -0,0 +1,606 @@
...
+    """Apply command-local settings that Daytona's session request cannot represent."""
+    if env:
+        assignments = ' '.join(shlex.quote(f'{name}={value}') for name, value in env.items())
+        command = f'env -- {assignments} sh -c {shlex.quote(command)}'
+    if cwd is not None:
+        command = f'cd -- {shlex.quote(cwd)} && {command}'
+    return command
+
+
+class _DaytonaProcess:
```
adtyavrdhn: Why is this not a dataclass?

pydantic_ai_harness/daytona_sandbox/_backend.py
```diff
@@ -0,0 +1,606 @@
...
+import anyio
+from pydantic_ai.sandboxes import CommandResult, FileEntry, SandboxTimeoutError, SandboxUnavailableError
+from typing_extensions import Self
+
+from pydantic_ai_harness._sandbox_provider import absolute_path, cleanup_call, raise_after_cleanup
+
+if TYPE_CHECKING:
+    from daytona import AsyncDaytona
+    from daytona._async.process import AsyncProcess
+    from daytona._async.sandbox import AsyncSandbox
```
adtyavrdhn: We should have import errors here?

pydantic_ai_harness/daytona_sandbox/_backend.py
```diff
@@ -0,0 +1,610 @@
...
+        self,
+        client: AsyncDaytona,
+        sandbox: AsyncSandbox,
+        *,
+        owned: bool,
+        working_dir: str | None = None,
+    ) -> None:
+        self._client = client
+        self.sandbox = sandbox
+        """The underlying Daytona sandbox, for provider-specific functionality."""
```
adtyavrdhn: Why is this here

adtyavrdhn: Hoist it up

pydantic_ai_harness/daytona_sandbox/_backend.py
adtyavrdhn: The per-command Daytona state is now a keyword-only dataclass; the hand-written forwarding constructor and redundant wait/_settle layer are removed.

pydantic_ai_harness/daytona_sandbox/_backend.py
adtyavrdhn: Daytona now has a guarded module-level import with an optional-extra installation hint. An isolated import test exercises the missing-package path.

pydantic_ai_harness/daytona_sandbox/_backend.py
adtyavrdhn: Hoisted the native sandbox access documentation onto async get_sandbox(). Both the README and docs page describe that accessor and the current lazy lifecycle.

pydantic_ai_harness/daytona_sandbox/_backend.py
```diff
@@ -0,0 +1,664 @@
...
+
+    async def run(
+        self,
+        command: SandboxCommand,
+        *,
+        shell: bool = False,
+        cwd: str | None = None,
+        env: Mapping[str, str] | None = None,
+        timeout: float | None = None,
+    ) -> CommandResult:
```
adtyavrdhn: create_or_attach() 

-> self.sandbox.run()

adtyavrdhn: Addressed in the workspace rewrite: the public `create_or_attach()` is gone. Acquisition now runs through the awaitable `workspace` property (`_get_workspace` -> `_create_or_attach`, under the lock), and `run()` plus every filesystem operation do `await self.workspace` first, so no operation can skip create-or-attach. This matches the Modal and E2B backends.

DouweM:
Review notes for this PR are in the shared comment on #734 (native handle naming → `get_client()`, drop "the application owns durable coordination", live provider runs incl. reattach-to-deleted raising `WorkspaceUnavailableError`, fresh-create races, lifecycle method names). — Douwe, written up by Claude

DouweM:
Following today's design decision (provider backends drop their own lifecycle helpers; provider lifecycle features go through the SDK object from `await backend.get_client()`), across this stack:

- #734 Modal: nothing to remove. `ModalSandboxBackend` already exposes only the workspace protocol plus `get_client()`; cleanup is documented as `await native.terminate.aio()` / `await native.detach.aio()` on the `modal.Sandbox` (Modal 1.5.2).
- #735 E2B: same. Cleanup is `await native.kill()` on the `e2b.AsyncSandbox` (e2b 2.51.0).
- #736 Daytona (this PR, 20665828): removed `DaytonaSandboxBackend.disconnect()` plus its test and fake hooks. The backend still closes an API client it created itself when acquisition fails, privately, so a retry starts clean. The docs and README example now own the client with `async with AsyncDaytona() as client:` passed as `client=`, and delete the sandbox with `await native.delete(wait=True)` on the `daytona.AsyncSandbox` (daytona 0.198.0). The PR body no longer describes `destroy()` / `stop()` / `pause()` / `disconnect()`; the first three had already left the code.

— Douwe, written up by Claude

DouweM:
Follow-up from today's session: an audit of `WorkspaceRef` semantics across core and the three provider backends (core side fixed in pydantic/pydantic-ai#6492 `b04547409`…`ada3ebce8`; the backend exception contract is now spelled out in `pydantic_ai.workspaces` and `docs/workspace.md` "Errors"). All three backends get the ref lifecycle right: `ref` is `None` until create returns and set right after under the lock, attach-or-`WorkspaceUnavailableError`, no name lookups that create. Daytona has exception-contract mismatches to fix here:

1. `daytona_sandbox/_backend.py:236-237` and `:292-293` map every `DaytonaNotFoundError` to `FileNotFoundError` / `exists() == False`, so a deleted sandbox looks like a missing file. Only a missing path may raise `FileNotFoundError`; a gone sandbox must raise `WorkspaceUnavailableError` (it ends the run, by contract).
2. `:351-352` `working_dir()` and `:251-255` the `write_bytes` mkdir omit `unavailable=True`, so a dead sandbox surfaces as a retryable `WorkspaceError`. `working_dir()` is the probe core recommends for "does this environment still exist", so it must raise `WorkspaceUnavailableError`.
3. `:316-318` and `:331-333` a create/attach RPC timeout raises `WorkspaceTimeoutError`, which is the *command* timeout type; Modal/E2B raise `WorkspaceError` there.
4. No `ref` / class docstring, unlike Modal and E2B.
5. `client.get` accepts a name and `:229` rewrites the ref to the real id; worth a docstring note.

Also (sandbox-migration / sandbox-toolsets): `coder/_agent.py:16-17` returns `None` for every ref, so a `local` ref in history is never rehydrated by the bundled coder; with core's `LocalWorkspace` capability landing, that agent could use it instead.

— Douwe, written up by Claude

PR #747, opened 2026-08-31
Warn on the deprecated single-key `get`/`add` of `InMemorySpendStore` and `RedisSpendStore`

### LINKED ISSUE #688: Deprecated single-key `get`/`add` on `InMemorySpendStore` and `RedisSpendStore` emit no warning

`InMemorySpendStore.get`, `InMemorySpendStore.add`, `RedisSpendStore.get` and `RedisSpendStore.add`
are deprecated by #548 and their docstrings name the release that removes them:

```
"""What `key` has accumulated. Deprecated in favour of `get_many`, removed in 0.28.0."""
```

None of the four warns. `SpendLimits` never calls them -- it drives `get_many` / `add_many` -- so the
only caller that reaches them is an application holding a concrete store and calling it by hand,
which is exactly the caller who gets no signal before the removal lands. Meanwhile every capability
page opens with

> deprecation warnings and release-note migration guidance tell you (or your agent) exactly how to
> upgrade

and #548 gives that warning only to the `SpendStore`-protocol adapter path (`as_batch_store`).

## Why it was not done in #548

`pyproject.toml` sets `filterwarnings = ['error']`, so adding the warnings turns roughly 43 existing
call sites across `tests/spend/test_spend.py` and `integration_tests/redis/test_live_redis.py` into
failures. The right fix for most of them is to call `get_many` / `add_many` instead, which is a
worthwhile sweep and a large diff in a PR that is already +1300 lines on a capability three other
open PRs also touch.

#548 states the deprecation in both docs surfaces instead, including that the four do not warn.

## What this needs

- A `HarnessDeprecationWarning` from each of the four, `stacklevel=2` so it lands on the caller.
- The test and integration-test call sites moved to `get_many` / `add_many`, except the ones whose
  subject is the deprecated pair itself.
- Both docs surfaces updated to drop the "they do not warn" clause once they do.

Refs #548.

xixi-box:
Fixes #688.

#548 deprecated the single-key `get`/`add` pair on `InMemorySpendStore` and `RedisSpendStore` (removed in 0.28.0) but none of the four warns. `SpendLimits` drives `get_many`/`add_many`, so the only caller that reaches them is an application holding a concrete store and calling it by hand -- the caller nothing else reaches before the removal lands.

## What changed

- Each of the four emits a `HarnessDeprecationWarning` naming its `get_many`/`add_many` replacement and the 0.28.0 removal, with `stacklevel=2` so the warning lands on the calling line.
- The test and integration-test call sites moved onto `get_many`/`add_many`, except the new tests whose subject is the deprecated pair itself: they pin the message, the still-working result, and the caller-pointing stacklevel for each of the four.
- Both docs surfaces drop the "a direct call to either does not warn" clause now that they do.

`_LegacyStore` in the test suite (a stand-in for a user's legacy `SpendStore`) keeps its single-key public shape but now delegates to the inner `InMemorySpendStore` through the batch pair, so the protocol-adapter tests keep testing what they tested without tripping the new warnings.

## Coordination

#714 also deprecates spend surface (`on_spend` and friends); it touches the event emitters rather than the stores, so no file overlap expected.


mpfaffenberger (approved):
Reviewed against #688. The warnings are emitted with `HarnessDeprecationWarning` and `stacklevel=2`, both maintained docs (spend README and docs/spend.md) carry the same migration text, the warnings themselves are tested via `pytest.warns`, and every internal caller including the Redis integration tests migrated to `get_many`/`add_many` so the suite stays warning-clean. Verified locally: 219 passed, merges clean with current main.

PR #752, opened 2026-09-01
Scrub ambient Logfire credentials in `tests/logfire_variables` so a local `.env` cannot reach the real Logfire API

### LINKED ISSUE #566: A local `.env` makes the test suite write to the real Logfire API
### Problem

`tests/guardrails/test_detectors.py` calls `load_dotenv()` at import time. `load_dotenv` mutates `os.environ` for the whole process, and `tests/guardrails/` sorts before `tests/logfire_variables/`, so on any checkout with a `.env` every later test in the run inherits whatever that file defines.

With a real Logfire token in scope, `logfire.configure()` resolves a `LogfireRemoteVariableProvider` instead of the `NoOpVariableProvider` the tests assume. The suite then makes **real POST requests to the production Logfire API** from background baseline-publish threads.

### Reproduction

On a checkout with no `.env`:

```
uv run --all-extras pytest -q -p no:randomly
4715 passed, 47 skipped
```

Symlink a `.env` containing a Logfire token into the same worktree, change nothing else, and re-run:

```
2 failed, 4710 passed, 47 skipped, 5 errors

FAILED tests/logfire_variables/test_auto_create.py::test_code_default_with_no_provider_does_not_create
FAILED tests/logfire_variables/test_auto_create.py::test_no_provider_makes_no_attempt
ERROR  tests/logfire_variables/test_managed_prompt.py::test_records_variable_resolution_span
ERROR  tests/logfire_variables/test_managed_prompt.py::test_provider_backed_resolution_uses_remote_value_and_label
ERROR  tests/logfire_variables/test_temporal.py::test_agent_control_skips_write_backs_in_temporal_workflow
```

`test_no_provider_makes_no_attempt` fails with `assert ['prompt__auto_no_provider'] == []` — auto-create fired against a provider the test believed was a no-op.

CI has no `.env`, so it is green there and this only bites locally. That is what makes it costly: it looks exactly like a regression in whatever you happen to be working on, and the real writes are invisible unless you go looking.

### Suggested fix

Any of:

- Don't call `load_dotenv()` from a test module. Load the values the detector tests need through a fixture scoped to those tests, or read them without mutating `os.environ`.
- If process-wide loading is genuinely needed, use a fixture that restores `os.environ` afterwards (`monkeypatch.setenv` per key, or `mock.patch.dict(os.environ, ...)`).
- Independently, make the logfire-variables tests assert their provider rather than assume it, so a leaked token fails loudly at the boundary instead of silently reaching the network.

The last one is worth doing regardless: a test that can reach the production API when an environment variable happens to be set should say so.

mpfaffenberger:
## Summary

An ambient Logfire credential (a shell export, or a `.env` loaded into `os.environ` as in the issue's reproduction) makes `logfire.configure()` lazily resolve a `LogfireRemoteVariableProvider` even with `send_to_logfire=False` (`LogfireConfig._lazy_init_variable_provider` reads `LOGFIRE_API_KEY` whenever `variables=` was not passed). The `tests/logfire_variables` tests assume the no-provider fallback, so with a credential in scope they wrote to the production Logfire API from background publish threads and failed on span contents far from the cause. Reproduced on current `main`: `LOGFIRE_API_KEY=<fake> uv run pytest tests/logfire_variables` gives 2 failed, 1 error.

The `load_dotenv()` call named in the issue no longer exists on `main` (it lives on the unmerged `capability/guardrails` branch, as do `test_auto_create.py` and `test_temporal.py` from the repro output), but the ambient-credential vector still applies to the surviving `test_managed_prompt.py`, including two tests from the issue's failure list.

Changes, matching the issue's suggested fixes:

- `tests/logfire_variables/conftest.py`: a package-scoped autouse fixture removes `LOGFIRE_TOKEN` and `LOGFIRE_API_KEY` for the duration of the package. Package scope orders it ahead of the module-scoped `logfire.configure()` fixture, which captures `api_key` at configure time.
- `test_no_variable_provider_is_configured`: asserts the resolved provider is a `NoOpVariableProvider`, so a leaked credential fails loudly at the boundary instead of reaching the network.

Test-only change, so no README or docs page updates apply.

Fixes #566

## How I verified

- `uv run --no-sync pytest -p no:cacheprovider tests/logfire_variables`: 23 passed
- `LOGFIRE_API_KEY=<fake> LOGFIRE_TOKEN=<fake> uv run --no-sync pytest -p no:cacheprovider tests/logfire_variables -p no:randomly`: 23 passed (before this change: 2 failed, 1 error, with real HTTP requests to the Logfire API)
- With `conftest.py` temporarily removed and a fake `LOGFIRE_API_KEY` set, `test_no_variable_provider_is_configured` fails, confirming the boundary assertion catches a leak
- `uv run --no-sync ruff format --check .`: 327 files already formatted
- `uv run --no-sync ruff check .`: all checks passed
- `PYRIGHT_PYTHON_IGNORE_WARNINGS=1 uv run --no-sync pyright tests/logfire_variables/conftest.py tests/logfire_variables/test_managed_prompt.py`: 0 errors, 0 warnings

### DIFF
--- a/tests/logfire_variables/conftest.py
+++ b/tests/logfire_variables/conftest.py
@@ -1,0 +1,31 @@
+"""Shared fixtures for the `logfire_variables` tests.
+
+These tests rely on the no-provider fallback: resolution uses code defaults
+because no Logfire variable provider is configured. Logfire lazily creates a
+`LogfireRemoteVariableProvider` that issues requests to the Logfire API
+whenever a credential such as `LOGFIRE_API_KEY` is present in `os.environ`,
+even with `send_to_logfire=False`. A credential inherited from the developer's
+shell, or from a `.env` file loaded into the process, would therefore make
+these tests write to the real Logfire API from background publish threads.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Iterator
+
+import pytest
+
+_LOGFIRE_CREDENTIAL_VARS = ('LOGFIRE_TOKEN', 'LOGFIRE_API_KEY')
+
+
+@pytest.fixture(autouse=True, scope='package')
+def _scrub_logfire_credentials() -> Iterator[None]:
+    """Remove ambient Logfire credentials while this package's tests run.
+
+    Package scope orders this ahead of the module-scoped `logfire.configure()`
+    fixtures in the test modules, which capture `api_key` at configure time.
+    """
+    with pytest.MonkeyPatch.context() as monkeypatch:
+        for name in _LOGFIRE_CREDENTIAL_VARS:
+            monkeypatch.delenv(name, raising=False)
+        yield
--- a/tests/logfire_variables/test_managed_prompt.py
+++ b/tests/logfire_variables/test_managed_prompt.py
@@ -26,6 +26,7 @@ import pytest
 from inline_snapshot import snapshot
 from logfire.testing import CaptureLogfire
 from logfire.variables import LabeledValue, Rollout, VariableConfig, VariablesConfig
+from logfire.variables.abstract import NoOpVariableProvider
 from opentelemetry.sdk.trace.export import SimpleSpanProcessor
 from pydantic_ai import Agent, RunContext
 from pydantic_ai.capabilities import Instrumentation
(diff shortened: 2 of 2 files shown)


mpfaffenberger:
@coderabbitai full review

mpfaffenberger:
@douwebot review

mpfaffenberger:
@codex review

mpfaffenberger:
@adtyavrdhn - this just prevents our test suite from picking up your local Logfire creds. Recommend merging.

PR #753, opened 2026-09-01
Publish explicit `StepPersistence.from_spec` fields to the `AgentSpec` schema

### LINKED ISSUE #537: `StepPersistence.from_spec` publishes no fields to the AgentSpec schema

mpfaffenberger:
## Summary

`StepPersistence.from_spec` was declared `(*args: Any, **kwargs: Any)`. Core reads that signature to build the capability's `AgentSpec` JSON schema (`pydantic_ai/agent/spec.py` `_get_schema_target`, then `pydantic_ai/_spec.py` `build_schema_types`), and `build_schema_types` drops variadic parameters, so the generated schema was the bare `{"const": "StepPersistence"}` with no `$defs` entry and no fields. A configured spec block loaded fine at runtime but failed validation against the schema its `# yaml-language-server: $schema=` line points at.

This names every spec-expressible parameter explicitly, keyword-only, following the `SpendLimits` fix for the same defect (#538) and folding in the review feedback from the earlier attempt (#543):

- 11 fields published: `backend`, `directory`, `database`, `max_snapshots_per_run`, `agent_name`, `run_id`, `parent_run_id`, `metadata`, plus the re-declared `AbstractCapability` fields `id`, `description`, `defer_loading` (which previously only reached `cls()` through `**kwargs` and would otherwise silently stop being expressible).
- `directory` and `database` are `str | Path | None = None`, with defaults resolved inside their matching backend branch, and a mismatched backend option raises `ValueError` instead of being dropped -- the `Memory.from_spec` guard shape -- so a spec cannot claim durability it does not deliver.
- `**unsupported: Any` stays as the catch-all: `store` (runtime-only, the field a user is most likely to try) gets a dedicated `UserError` pointing at Python construction; any other unknown field keeps the specific `has no spec field(s)` error. Core drops the variadic from the schema, so `additionalProperties: false` still holds.
- The unknown-backend branch is exercised through the public `Agent.from_spec` path (the `tests/memory` idiom), keeping the `Literal` annotation cast-free in tests.

The issue's reproduction is a regression test: `TestFromSpecSchema::test_agent_spec_schema_publishes_fields_and_excludes_runtime_store` asserts `spec_params_StepPersistence` exists in `$defs` with exactly the 11 fields, `additionalProperties: false`, the `backend` enum, and no `store`. Docs parity: the one behavior change visible to spec authors (the `directory`/`database` restriction) gets a line in both the capability README and `docs/step-persistence.md`, next to their existing `from_spec` mention.

Fixes #537

## How I verified

- `uv run --no-sync pytest -p no:cacheprovider tests/step_persistence` -- 253 passed
- `uv run --no-sync pytest -p no:cacheprovider tests/test_docs_parity.py` -- 277 passed
- `PYRIGHT_PYTHON_IGNORE_WARNINGS=1 uv run --no-sync pyright pydantic_ai_harness/step_persistence/_capability.py tests/step_persistence/test_step_persistence.py` -- 0 errors
- Manual schema check: `AgentSpec.model_json_schema_with_capabilities([StepPersistence])` now carries `spec_params_StepPersistence` with all 11 fields (was absent on `main`).


mpfaffenberger:
@coderabbitai full review

mpfaffenberger:
@douwebot review

mpfaffenberger:
@codex review

mpfaffenberger:
@dsfaccini - can you let me know more here? I had my Fable implement this issue, but I dont know if we want AgentSpec coupled to StepPersistence. What do you think?

pydantic_ai_harness/step_persistence/_capability.py
```diff
@@ -139,33 +141,90 @@ class StepPersistence(AbstractCapability[AgentDepsT]):
...
+        database: str | Path | None = None,
+        max_snapshots_per_run: int | None = None,
+        agent_name: str | None = None,
+        run_id: str | None = None,
+        parent_run_id: str | None = None,
+        metadata: dict[str, str] | None = None,
+        id: str | None = None,
+        description: str | None = None,
+        defer_loading: bool = False,
+        **unsupported: Any,
```
mpfaffenberger, line 158: remove

dsfaccini:
> @dsfaccini - can you let me know more here? I had my Fable implement this issue, but I don't know if we want AgentSpec coupled to StepPersistence. What do you think?

I don't know I'm sorry ❤️ will need to think about it when I get back

one thing that would be optimal would be to centralize the storage side of the args in a single arg, instead of having both "database" and "directory", since they serve the same purpose (storage)

but I'm too mentally out of the capability rn so yeah, I'll need to dig in when I get back

Issue #553, opened 2026-08-05
One field type with no JSON schema deletes a capability's whole `AgentSpec` entry (`CodeMode`, `ConversationSearch`, `ModalSandbox`, `Skills`)


dsfaccini:
This issue was posted by Claude Code using claude-opus-5 on behalf of David.

Review level: found by the repo-wide sweep in #546 and verified per capability against pydantic-ai 2.22.0. David directed the work and approved filing.

One of four root causes behind capabilities that publish no fields to the `AgentSpec` schema. The
guard in #546 xfails these; this issue is what that xfail points at.

## The mechanism

`pydantic_ai/_spec.py` `filter_serializable_type` strips only `TypeVar` and `Callable`, so an
arbitrary class survives into a TypedDict built with `arbitrary_types_allowed=True`. Pydantic then
drops the entire `spec_<Name>` member from the capabilities union rather than raising. **One
unrepresentable field erases every other field on the capability.**

## The rule, probed rather than assumed

- a bare `X` fails
- `X | None` fails (pydantic's nullable path)
- `X | <representable>` is benign -- only `X` is dropped

That distinction decides whether a given fix is a one-liner.

## Affected

- `CodeMode.os_access: AbstractOS | None`
- `ConversationSearch.source: HistorySource` -- bare, so the short form dies too
- `ModalSandbox.session: ModalSandboxSession | None`
- `Skills.include` / `Skills.exclude: Collection[str] | None` -- long form only; the short form
  survives, which is the only reason `Skills` passes layer 2

Fix shape in each case is a `from_spec` naming the spec-expressible parameters, following
`SpendLimits.from_spec`.

Refs #537, #546.

dsfaccini:
This comment was posted by Claude Code using claude-opus-5 on behalf of David. Unreviewed by him.

**Referenced from shipped code**, so the reference has a constraint attached.

[#546](https://github.com/pydantic/pydantic-ai-harness/pull/546) adds `tests/test_capability_specs.py`, whose `_TRACKERS['unrepresentable-field']` points at this issue, and whose module docstring documents this as cause 3.

**The constraint: fixing a capability here means deleting its entry from `_NO_SCHEMA_ENTRY` (or `_NO_BASE_FIELDS`, for `Skills`) in the same PR.** Those xfails are `strict=True`, so a capability that starts passing turns the build red until its line is removed.

Two things the tests encode that are easy to lose:

[...]

PR #756, opened 2026-09-01
Publish spec-schema fields for `Advisor`, `SummarizingCompaction`, `ManagedPrompt`, and `PyaiDocs`

### LINKED ISSUE #552: `Advisor`, `SummarizingCompaction`, `ManagedPrompt`, `PyaiDocs` publish no schema fields: a name in the read annotations does not resolve at runtime

Review level: found by the repo-wide sweep in #546 and verified per capability against pydantic-ai 2.22.0. David directed the work and approved filing.

One of four root causes behind capabilities that publish no fields to the `AgentSpec` schema. The
guard in #546 xfails these; this issue is what that xfail points at.

## The mechanism

`pydantic_ai/agent/spec.py` `_get_schema_target` reads `cls.__init__` when `from_spec` is not
overridden, wrapped in `except (NameError, TypeError, AttributeError)`, and then falls back to
`AbstractCapability`'s variadic `from_spec`. Variadic parameters are dropped by
`build_schema_types`, so the capability lands in exactly #537's state -- a bare
`{"const": "<Name>"}` with no fields -- purely because a name in its `__init__` annotations does
not resolve at runtime. It still loads fine; only schema validation fails.

## Affected, with verified fix depth

Each is different, so this is not one mechanical change:

- **`PyaiDocs`** -- needs only `from pathlib import Path` at runtime in `_deprecated.py`
  (`@dataclass` rebuilds `__init__` there).
- **`SummarizingCompaction`** -- needs `Model`/`KnownModelName` at runtime, and is then fully fixed.
- **`Advisor`** -- same fix reaches layer 2, but it still lacks the `AbstractCapability` base
  fields, so it also needs #554.
- **`ManagedPrompt`** -- cannot be fixed this way at all. Resolving the name exposes
  `logfire_instance: Logfire | None`, which then hits the nullable-drop in #553. Needs a
  `from_spec` override.

Note `Advisor` and `SummarizingCompaction` fail on core's `ModelSelection`, which is the *string*
`'Model | KnownModelName | str'` resolved against the harness module's globals.

Refs #537, #546.

mpfaffenberger:
## Summary

The agent-spec schema builder reads `__init__` annotations when `from_spec` is not overridden. When a name in those annotations does not resolve at runtime, it falls back to `AbstractCapability`'s variadic `from_spec`, and the capability's schema entry collapses to a bare `{"const": "<Name>"}` with no fields. Loading still worked; only schema validation was broken.

Each capability named in the issue needed a different depth of fix, matching the issue's verified map:

- **`PyaiDocs`** (`pydantic_ai_docs/_deprecated.py`): `@dataclass` rebuilds `__init__` in that module, so the inherited `local_docs_path: Path` annotation must resolve against its globals. Added a runtime `from pathlib import Path`.
- **`Advisor`** (`advisor/_capability.py`): core's `ModelSelection` alias is the string `'Model | KnownModelName | str'`, evaluated against the advisor module's globals. Added runtime imports of `Model` and `KnownModelName`. Its schema now publishes `model` (required), `mode`, `max_uses`, `max_tokens`, `caching`, and `forward_history`. As the issue notes, the missing `AbstractCapability` base fields are #554's territory and are not addressed here.
- **`ManagedPrompt`** (`logfire/_managed_prompt.py`): resolving `Logfire` at runtime would only move the failure into the nullable-drop tracked by #553, so it gets a `from_spec` override instead, covering the spec-serializable options (`name` and `default` required, plus `label`, `targeting_key`, `attributes`, `render_template`, and the `id`/`description`/`defer_loading` base fields). `logfire_instance` and the callable forms are not spec-serializable and are documented as such in the override's docstring.
- **`SummarizingCompaction`**: already resolves on current main (it imports `Model` at runtime since the durable-summarization work), so this PR only pins the published property set with a regression test.

Regression tests assert the exact `spec_params_<Name>` property sets via `AgentSpec.model_json_schema_with_capabilities`, following the shape of the existing `Memory` spec-schema test, plus `from_spec` construction and `Agent.from_spec` loading coverage for `ManagedPrompt`.

Note: #546 (which adds the `tests/test_capability_specs.py` guard whose xfails point at this issue) is not yet merged, so there are no `_NO_SCHEMA_ENTRY` lines to delete here; whichever lands second should reconcile the tracker map.

Fixes #552.

## How I verified

- `uv run --no-sync pytest -p no:cacheprovider tests/advisor tests/compaction tests/logfire_variables tests/pydantic_ai_docs` -- 662 passed, 1 skipped
- `PYRIGHT_PYTHON_IGNORE_WARNINGS=1 uv run --no-sync pyright` on all 7 changed files -- 0 errors, 0 warnings
- Reproduced the issue before the fix (all four capabilities' entries were a bare `{"const": ...}`) and confirmed all four publish their fields after it.

## Checklist

### DIFF
--- a/pydantic_ai_harness/pydantic_ai_docs/_deprecated.py
+++ b/pydantic_ai_harness/pydantic_ai_docs/_deprecated.py
@@ -4,6 +4,10 @@ from __future__ import annotations
 
 from dataclasses import dataclass
 
+# `@dataclass` rebuilds `__init__` in this module, so the inherited `local_docs_path: Path`
+# annotation must resolve against these globals when the agent-spec schema is built (#552).
+from pathlib import Path  # noqa: F401  # pyright: ignore[reportUnusedImport]
+
 from pydantic_ai.tools import AgentDepsT
 
 from pydantic_ai_harness.pydantic_ai_docs._capability import PydanticAIDocs
--- a/pydantic_ai_harness/advisor/_capability.py
+++ b/pydantic_ai_harness/advisor/_capability.py
@@ -10,7 +10,16 @@ from pydantic_ai import Agent
 from pydantic_ai.capabilities import ModelSelection, NativeOrLocalTool
 from pydantic_ai.exceptions import ModelRetry, UnexpectedModelBehavior, UserError
 from pydantic_ai.messages import ModelResponse
-from pydantic_ai.models import ModelRequestContext, parse_model_id
+
+# `KnownModelName` and `Model` back core's `ModelSelection` alias, which is the string
+# `'Model | KnownModelName | str'` evaluated against this module's globals when the
+# agent-spec schema reads `__init__` annotations, so both must be imported at runtime (#552).
+from pydantic_ai.models import (
+    KnownModelName,  # noqa: F401  # pyright: ignore[reportUnusedImport]
+    Model,  # noqa: F401  # pyright: ignore[reportUnusedImport]
+    ModelRequestContext,
+    parse_model_id,
+)
 from pydantic_ai.native_tools import AdvisorTool
 from pydantic_ai.settings import ModelSettings
 from pydantic_ai.tools import AgentDepsT, AgentNativeTool, RunContext
--- a/pydantic_ai_harness/logfire/_managed_prompt.py
+++ b/pydantic_ai_harness/logfire/_managed_prompt.py
@@ -208,3 +208,36 @@ class ManagedPrompt(AbstractCapability[AgentDepsT]):
                 return await handler()
             finally:
                 self._resolved.reset(token)
+
+    @classmethod
+    def from_spec(
+        cls,
+        name: str,
+        default: str,
+        *,
+        label: str | None = None,
+        targeting_key: str | None = None,
+        attributes: Mapping[str, Any] | None = None,
+        render_template: bool = False,
+        id: str | None = None,
+        description: str | None = None,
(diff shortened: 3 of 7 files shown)


mpfaffenberger:
@coderabbitai full review

mpfaffenberger:
@codex review

mpfaffenberger:
@douwebot review

mpfaffenberger:
Desirable - missing Capabilities and features from spec.

pydantic_ai_harness/pydantic_ai_docs/_deprecated.py
```diff
@@ -4,6 +4,10 @@
 
 from dataclasses import dataclass
 
+# `@dataclass` rebuilds `__init__` in this module, so the inherited `local_docs_path: Path`
+# annotation must resolve against these globals when the agent-spec schema is built (#552).
+from pathlib import Path  # noqa: F401  # pyright: ignore[reportUnusedImport]
```
adtyavrdhn, line 9: This looks weird?

adtyavrdhn:
Can you just have a look at the path thing, looks good to me otherwise

pydantic_ai_harness/pydantic_ai_docs/_deprecated.py
mpfaffenberger: This is absolutely insane.

PR #758, opened 2026-09-01
Shield the cancel+`gather` teardown in `guardrails` parallel mode and `MontyExecutor`

### LINKED ISSUE #559: Shield cancel+gather teardown in guardrails parallel mode and _monty_exec
### Summary

Two capabilities that spawn raw `asyncio.create_task(...)` tasks tied to a run — `guardrails/_capability.py` (parallel `wrap_model_request`) and `_monty_exec.py` (CodeMode's dispatched work in `_pending`) — cancel and `await asyncio.gather(...)` their tasks in a `finally`, but the await is **not shielded**.

### Details

Pydantic AI first-party run cancellation (`cancel_run()` / `CancellationToken`, pydantic/pydantic-ai#6497) cancels the run's driving task, exactly like an external `asyncio.wait_for` timeout. Under a **single** cancellation the teardown here completes fine (the cancel loop runs before the await; verified). But a **racing second** cancellation during teardown (first-party + external at once, or Python 3.10 where the two can't be disambiguated) can interrupt the `gather`, so the run unwinds before the drain finishes. Nothing runs forever (the tasks are already `cancel()`'d), so this is consistency hardening, not a runaway leak.

### Fix

Wrap the cancel+`gather` teardown in `anyio.CancelScope(shield=True)`, matching pydantic/pydantic-ai#7095's stream-teardown precedent. `experimental/acp` is the in-repo reference pattern (separate task + shielded cleanup, aware that a raw `task.cancel()` pierces shields). Do this once the equivalent fix lands for `BackgroundTools` (#222), for a consistent teardown idiom across the raw-`create_task` capabilities.

_Filed by Claude Code on behalf of @DouweM._

mpfaffenberger:
## Summary

Pydantic AI run cancellation (`cancel_run()` / an external timeout) can arrive while an enclosing anyio cancel scope is already cancelled, and a cancelled anyio scope re-delivers cancellation to its tasks on every event-loop cycle. Two teardowns awaited their cancelled tasks without a shield:

- `guardrails/_capability.py`, parallel `wrap_model_request`: the `finally` drain over the guard and handler tasks.
- `_monty_exec.py`, `MontyExecutor.run`: the cleanup `gather` over still-pending dispatched tasks (shared by `CodeMode` and `DynamicWorkflow`).

Under that repeated delivery, each cycle either aborts the drain outright (the tasks outlive the call mid-unwind) or is forwarded through the `gather` into every task, breaking any await their cancellation cleanup performs. Both drains are now wrapped in `anyio.CancelScope(shield=True)`, matching the stream-teardown precedent in pydantic/pydantic-ai#7095. As `experimental/acp` documents, the shield holds for anyio-scope cancellation; a raw repeated `Task.cancel()` can still pierce it.

The new tests are event-sequenced (no sleep-based races): each drives the teardown from inside a cancelled `anyio.CancelScope`, holds the task's cancellation handler on an event, and asserts that the caller stays blocked until the handler is released, that the handler's awaits survive (the Monty test counts re-cancels and requires zero), and that the work has fully unwound when the caller returns. The Monty test reaches the executor's cleanup through the sequential-barrier path, the one place a dispatched task is still running in `_pending` when the teardown owns it. Both tests fail against the unshielded code. The stale docstring on `test_no_dangling_tasks_on_outer_cancellation` (which recorded that no shield was needed) now points at the case where it is.

Fixes #559

## How I verified

- `uv run --no-sync pytest -p no:cacheprovider tests/guardrails tests/code_mode tests/dynamic_workflow` -- 564 passed
- `PYRIGHT_PYTHON_IGNORE_WARNINGS=1 uv run --no-sync pyright pydantic_ai_harness/_monty_exec.py pydantic_ai_harness/guardrails/_capability.py tests/code_mode/test_code_mode.py tests/guardrails/test_input_guardrail.py` -- 0 errors
- Regression check: with the two source files reverted, both new tests fail (guardrails: the runner returns while the guard is still unwinding; code_mode: the pending task is re-cancelled every cycle). With the fix, both pass 10/10 repeated runs.

### DIFF
--- a/pydantic_ai_harness/_monty_exec.py
+++ b/pydantic_ai_harness/_monty_exec.py
@@ -22,6 +22,8 @@ from collections.abc import Callable, Container, Coroutine
 from dataclasses import dataclass, field
 from typing import Any
 
+import anyio
+
 try:
     from pydantic_monty import (
         CollectString,
@@ -131,8 +133,15 @@ class MontyExecutor:
                 # point; await them so dispatched work (e.g. sub-agent runs mutating shared
                 # usage) has fully unwound before this returns. `return_exceptions=True` keeps
                 # one task's teardown error from masking the original exception, and the
-                # results are deliberately discarded.
-                await asyncio.gather(*cancelled, return_exceptions=True)
+                # results are deliberately discarded. Shielded: run cancellation can land here
+                # with an enclosing anyio scope already cancelled, and that scope re-cancels
+                # its tasks on every event-loop cycle -- each delivery either aborts this await
+                # outright (abandoning the tasks mid-unwind) or is forwarded through the
+                # `gather` into every task, breaking any await their cleanup performs. The
+                # shield holds for anyio-scope cancellation; a raw second `Task.cancel()` can
+                # still pierce it.
+                with anyio.CancelScope(shield=True):
+                    await asyncio.gather(*cancelled, return_exceptions=True)
         return state
 
     async def _handle_function(self, snapshot: FunctionSnapshot) -> MontyState:
--- a/pydantic_ai_harness/guardrails/_capability.py
+++ b/pydantic_ai_harness/guardrails/_capability.py
@@ -34,6 +34,7 @@ from collections.abc import Awaitable, Callable, Sequence
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any
 
+import anyio
 from pydantic_ai.capabilities import AbstractCapability, CapabilityOrdering, Instrumentation, WrapModelRequestHandler
 from pydantic_ai.exceptions import ModelRetry, SkipModelRequest, UserError
 from pydantic_ai.messages import ModelMessage, ModelResponse, TextPart, UserPromptPart
@@ -301,7 +302,16 @@ class InputGuardrail(AbstractCapability[AgentDepsT]):
                 if not task.done():
                     task.cancel()
 
-            await asyncio.gather(guard_task, handler_task, return_exceptions=True)
+            # Run cancellation (`cancel_run()`, `anyio.fail_after`) can land here with an
+            # enclosing anyio scope already cancelled, and that scope re-cancels its tasks on
+            # every event-loop cycle -- each delivery either aborts the drain below outright
+            # (abandoning the guard and handler tasks mid-unwind) or is forwarded through the
+            # `gather` into both tasks, breaking any await their cleanup performs. Shield the
+            # drain so both tasks fully unwind before the cancellation propagates. The shield
+            # holds for anyio-scope cancellation; a raw second `Task.cancel()` can still
+            # pierce it.
+            with anyio.CancelScope(shield=True):
+                await asyncio.gather(guard_task, handler_task, return_exceptions=True)
 
 
 @dataclass
--- a/pydantic_ai_harness/guardrails/_capability.py
+++ b/pydantic_ai_harness/guardrails/_capability.py
@@ -310,6 +310,16 @@ class InputGuardrail(AbstractCapability[AgentDepsT]):
             # drain so both tasks fully unwind before the cancellation propagates. The shield
             # holds for anyio-scope cancellation; a raw second `Task.cancel()` can still
             # pierce it.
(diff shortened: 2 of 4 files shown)


pydantic_ai_harness/guardrails/_capability.py
```diff
@@ -301,7 +302,16 @@ async def run_handler() -> ModelResponse:
                 if not task.done():
                     task.cancel()
 
-            await asyncio.gather(guard_task, handler_task, return_exceptions=True)
+            # Run cancellation (`cancel_run()`, `anyio.fail_after`) can land here with an
```
adtyavrdhn, line 305: Should we also make this an anyio task group in that case https://anyio.readthedocs.io/en/stable/cancellation.html#shielding in that case.

mpfaffenberger:
@coderabbitai full review

mpfaffenberger:
@douwebot review

mpfaffenberger:
@codex review

pydantic_ai_harness/_monty_exec.py
```diff
@@ -131,8 +133,15 @@ async def run(self, state: MontyState) -> MontyComplete:
...
-                # results are deliberately discarded.
-                await asyncio.gather(*cancelled, return_exceptions=True)
+                # results are deliberately discarded. Shielded: run cancellation can land here
+                # with an enclosing anyio scope already cancelled, and that scope re-cancels
+                # its tasks on every event-loop cycle -- each delivery either aborts this await
+                # outright (abandoning the tasks mid-unwind) or is forwarded through the
+                # `gather` into every task, breaking any await their cleanup performs. The
+                # shield holds for anyio-scope cancellation; a raw second `Task.cancel()` can
+                # still pierce it.
+                with anyio.CancelScope(shield=True):
```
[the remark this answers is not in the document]
mpfaffenberger: Thanks for flagging this. I’m dismissing it for this change because fully unwinding dispatched work before returning is the teardown contract here: those tasks can still be mutating shared usage/state, so a finite grace period would knowingly restore the task-abandonment bug after an arbitrary timeout. The scope shield only protects against enclosing anyio cancellation; a raw second `Task.cancel()` remains able to pierce it as the code comment notes. If we want a bounded-shutdown policy, that should be designed separately with an explicit answer for what happens to still-mutating work.

pydantic_ai_harness/guardrails/_capability.py
mpfaffenberger: > This comment was posted by Claude Code on behalf of Mike.

Good question — I worked it through, and a task group makes this specific failure worse rather than better, for three reasons:

1. **The children being raw `asyncio` tasks is what confines the bug to the host task.** anyio only delivers (and re-delivers) scope cancellation to tasks its scopes track. The guard and handler tasks are untracked, so the cancelled enclosing scope can only hammer *this* function's task — which the shielded drain absorbs — and each child sees exactly one `.cancel()` and unwinds cleanly. As task-group children they'd be anyio-tracked, and the cancelled outer scope would re-cancel them inside their own cleanup awaits on every cycle: the exact #559 failure, relocated one level down where no shield is standing.

2. **A task group can't shield only its drain.** The wait-for-children happens in `__aexit__`, so the only lever is `tg.cancel_scope.shield = True`, which shields the whole race — `cancel_run()` / `fail_after` arriving mid-race would then be unable to interrupt at all, instead of being held only for the duration of the drain as it is today.

3. **anyio ≥ 4 wraps child exceptions in `BaseExceptionGroup`.** The guard's verdicts propagate as exceptions — `SkipModelRequest` for `block` has to reach pydantic-ai's run machinery unwrapped — so we'd need unwrap-and-reraise plumbing plus a result container (`start_soon` returns nothing). That reintroduces more hand-rolled machinery than the task group removes.

Recorded a condensed version of this at the drain site in d6e2cb21 so the decision lives in the code rather than only in this thread.

mpfaffenberger: Number 2 is incorrect according to Aditya. Double / Triple Check this.

PR #761, opened 2026-09-01
Run `ShellToolset` and `FileSystemToolset` against `ctx.workspace`

adtyavrdhn:
Runs the `ShellToolset` and `FileSystemToolset` against the run's workspace (`ctx.workspace`) instead of the host filesystem, so both tools act on whatever workspace a run attaches -- local, a container, or a provider-backed environment -- and share it with the rest of the run.

Rebased onto the workspace-migrated base and repinned core to pydantic/pydantic-ai#6492. The filesystem tools delegate every read, write, and directory listing to `ctx.workspace` while keeping their own root confinement and allow/deny/protected policy, and they still emit the `FileReadEvent`, `FileWrittenEvent`, and `DirectoryListedEvent` traversal events that `RepoContext` subscribes to. Binary classification and per-line capping now come from the workspace's bounded `read_file`.


gabrycina:
I reconciled this branch with the current `codex/sandbox-migration` base in #832. The companion PR migrates the implementation to the current `Workspace` APIs while preserving remote FileSystem/Shell behavior, toolset IDs, and per-run shell state. Local Ruff and focused Pyright checks passed, along with 947 focused tests (2 skipped). If the approach looks right, merging #832 into `sandbox-toolsets` will update this PR and trigger fresh CI here.

DouweM:
## Review notes (Douwe, written up by Claude)

**This PR now conflicts with its base and I did not resolve it**, because it needs your decisions: `main` rewrote `FileSystemToolset` and the Shell toolset around host paths at the same time this PR rewrote them around `ctx.workspace` (16 files; `filesystem/_toolset.py` alone is 42 hunks). `main` added fd-level hash-checked writes, batch `Replacement` edits, character-bounded read windows, `FileChangeRequestEvent` cancellation, ripgrep-backed `list_files`/`grep`, and a persistent `shell` tool with a host-side supervisor. Porting those onto the workspace raises questions a merge can't settle: does `rg` run through `workspace.run` when a remote environment may not have it; do the fd-level write guarantees survive; what does a persistent shell with PIDs and log files mean on a remote workspace?

Other things to fix here:
1. **A write against a `ReadOnlyWorkspace` aborts the whole run.** `ReadOnlyWorkspace` raises `UserError` and both toolsets' `_recoverable` re-raise it, so the model never hears "this workspace is read-only". Nothing connects `FileSystem(read_only=True)` to the workspace's actual state either; the capability should hide write tools when the workspace is read-only, and a refused write should reach the model.
2. **Reader-tool duplication.** `read_tool_result`, `read_pyai_docs`, `inventory_agent_context` and the memory tools overlap `read_file`/`search_files`/`list_directory`/`find_files`, and no capability can tell whether file tools are attached. We need a small "file tools present, rooted here, writable or not" protocol discoverable through `ctx.capabilities` (same shape as `TranscriptHandleProvider`), so others can drop their redundant tool in `prepare_tools`. Part of the metadata proposal linked from #522.
3. Current `coverage` misses: `filesystem/_toolset.py:113` and `:594-595`, `tests/filesystem/test_filesystem.py:196`.

adtyavrdhn:
Folded into #522, which now carries the whole workspace port (this Shell/FileSystem work included) so there is one PR for it. The decisions listed here are carried over in #522's body.

PR #774, opened 2026-09-02
Name the internal `Agent` in `SummarizingCompaction`, `ToolOutputLimits`, `SystemReminders`, `Advisor` and `PydanticAIChatModel` so Logfire attributes their runs

mpfaffenberger:
## Why

A user tracking summarization spend in Logfire reported the `SummarizingCompaction` summarizer run showing up with `agent_name = self`. Core infers an unnamed `Agent`'s name by scanning the caller's frame locals for the agent object; inside a capability method (and especially under a durability wrapper) that lands on whatever the frame bound first, so the run is unattributable.

## What

- Pass `name=` explicitly for every `Agent` a capability constructs internally, set to the capability's snake_case name:
  - `SummarizingCompaction` -> `summarizing_compaction`
  - `ToolOutputLimits` (`Summarize` action) -> `tool_output_limits`
  - `SystemReminders` (`LLMReminder`) -> `system_reminders`
  - `Advisor` (local mode) -> `advisor`
  - `PydanticAIChatModel` (browser-use) -> `browser_use`
- Shared `instrument_all_agents` fixture + `agent_run_names(capfire)` helper in `tests/conftest.py`; per-agent `instrument=` on the outer agent does not reach the inner one, so `Agent.instrument_all()` is the only way to observe those spans. One test per capability asserts the run span's `agent_name`.
- Write the rule into `agent_docs/capability-authoring.md` and `agent_docs/review-checklist.md` so new capabilities follow it.
- Document the `agent_name` filter in the compaction README and `docs/compaction.md` (kept in parity).

## Verification

```
uv run --no-sync ruff format --check .   # clean
uv run --no-sync ruff check .            # clean
pyright on touched files                 # 0 errors
pytest tests/compaction tests/tool_output_limits tests/system_reminders tests/advisor tests/browser_use
# 959 passed, 1 skipped
```

The new compaction test was verified to fail before the fix (`['agent', 'outer']`) and pass after.

### DIFF
--- a/agent_docs/capability-authoring.md
+++ b/agent_docs/capability-authoring.md
@@ -129,6 +129,24 @@ warnings where practical.
   editors, and stack traces (`modal_sandbox` is the reference; `filesystem` is
   0-based pending migration).
 
+### Internal Agents Carry The Capability's Name
+
+A capability that constructs its own `Agent` (a summarizer, an advisor, a
+reminder generator) passes `name=` explicitly, set to the capability's
+snake_case name: `Agent(model, name='summarizing_compaction', ...)`. That name
+is the `agent_name` on the run span, which is what Logfire groups runs and
+costs by.
+
+Without it, core infers the name from the caller's frame locals. Inside a
+capability method that lands on `agent`, `self`, or whatever a durability
+wrapper happened to bind, so the run is unattributable in tracing. Matching
+the capability name lets a reader trace spend straight back to the
+capability without knowing its internals.
+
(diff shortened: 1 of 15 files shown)


adtyavrdhn:
Let us make a rule out of it and find out why we missed it too

mpfaffenberger:
> Let us make a rule out of it and find out why we missed it too

`capability_authoring.md` now has this as guidance. Did you have something else in mind? Can we merge this?

Issue #879, opened 2026-09-16
Make `Coder` file-read truncation explicit and reliably resumable


Quiyyy:
### Capability

`Coder`

### pydantic-ai-harness version

Source checkout at `407f03326e5aed7371aaf2bde86b3967d074be78`

### pydantic-ai version

`pydantic-ai-slim 2.42.0`

### Environment

Python 3.14.6, Windows 11

### Bug Description

`Coder`'s `read_file` uses line-based offsets, but its character budget can truncate a line midway. The response does not identify the incomplete line or provide a precise continuation offset. A caller continuing after the last displayed line can therefore skip unread content.

The implementation slices each rendered line to the remaining budget, then appends the same continuation hint whether it reached an output limit or EOF. [Relevant source](https://github.com/pydantic/pydantic-ai-harness/blob/407f03326e5aed7371aaf2bde86b3967d074be78/pydantic_ai_harness/coder/_toolset.py#L52-L87).

### Minimal Reproduction

This uses `FunctionModel` to exercise the public `Agent` interface without provider calls.

```python
import json
from collections.abc import AsyncIterator
from pathlib import Path
from tempfile import TemporaryDirectory

from pydantic_ai import Agent
from pydantic_ai.messages import ModelMessage, ToolReturnPart
from pydantic_ai.models.function import AgentInfo, DeltaToolCall, FunctionModel

from pydantic_ai_harness.coder import Coder


def read_file(workspace: Path, offset: int = 0) -> str:
    async def respond(
        messages: list[ModelMessage], info: AgentInfo
    ) -> AsyncIterator[str | dict[int, DeltaToolCall]]:
        if any(isinstance(part, ToolReturnPart) for message in messages for part in message.parts):
            yield 'done'
        else:
            args = json.dumps({'path': 'example.txt', 'offset': offset})
            yield {0: DeltaToolCall(name='read_file', json_args=args)}

    agent = Agent(FunctionModel(stream_function=respond), capabilities=[Coder(workspace)])
    result = agent.run_sync('Read the file.')
    return next(
        str(part.content)
        for message in result.all_messages()
        for part in message.parts
        if isinstance(part, ToolReturnPart) and part.tool_name == 'read_file'
    )


with TemporaryDirectory() as directory:
    workspace = Path(directory)
    (workspace / 'example.txt').write_bytes(b'A' * 40_000 + b'\n' + b'B' * 40_000 + b'\nTHIRD\n')
    first = read_file(workspace)
[...]

PR #896, opened 2026-09-17
Give `FileSystem` a `max_read_chars` budget so `Coder` reads page without skipping lines

mpfaffenberger:
## Summary

Reported by @Quiyyy in #879 and fixed for the old private `Coder` reader in #892. Since #895 composes `Coder` from the public `FileSystem`, that diff no longer applies, so this moves the fix into `FileSystem` with @Quiyyy credited as co-author.

The bug reproduces in the composition: a 2,000-line `read_file` of 70-character lines exceeds `Coder`'s 64,000-character tool-output cap, which removes lines 334 to 1510 from the middle of the window while the footer still says `Use offset=2000 to continue reading`, so paging by `offset` skips content.

## Change

- `FileSystem` gains `max_read_chars: int | None = None`. With it set, the window ends on the last complete numbered line that fits, the continuation hint names the first line not shown, and a line too long for the window is named with a skip offset (`Use offset=N to skip it, or a shell byte range to inspect it`). Validated positive like the other caps. Default `None` is no change for existing users.
- `Coder` sets `max_read_chars=60000`, under its 64,000-character output cap, so the cap never touches a read. `MAX_READ_CHARS` and `MAX_OUTPUT_CHARS` are now named constants.
- Both docs pages of both capabilities updated.

## Not changed from #892

The zero-based `offset` contract stays as shipped; the cursor discussion in #879 remains separate, as #892 recommended.

## Verification

Ruff, scoped Pyright, `tests/coder`, `tests/filesystem`, `tests/test_coder.py`, `tests/test_docs_parity.py`; focused branch coverage on `filesystem` and `coder` at 100%. Regressions: `TestMaxReadChars` (window ends on a complete line, `limit` still applies, oversized line named and skippable, validation) and `test_long_reads_page_without_gaps` through `Agent(capabilities=[Coder(...)])`.

Refs #879. Supersedes #892.

Co-authored-by: Quiyyy <120462130+Quiyyy@users.noreply.github.com>


mpfaffenberger:
Closing: the fix lives in Quiyyy's #892 instead, updated on top of main.

Issue #875, opened 2026-09-16
Add a standalone `pydantic-clai2` terminal client


mpfaffenberger:
Add a separately installable terminal client with Coder as its default agent, support for custom agents, retained conversation history, Termflow streaming output and thinking, and CLAI branding based on Code Puppy. Terminal dependencies belong to the standalone package rather than the Harness base install.

PR #928, opened 2026-09-18
Share the root `uv` workspace and `Pyright` configuration with `pydantic-clai2`

Kludex:
Move `pydantic-clai2` into the root workspace, remove its lockfile and duplicate development configuration, and typecheck both packages with the root `Pyright` configuration.

Keep Harness compatible with Python 3.10 by retaining its slim, all-extras, and lowest-versions CI jobs. Python 3.10 installs use shared-lock version ceilings and the root provider overrides, allowing compatible dependency fallbacks without unrelated upgrades. Workspace installs require Python 3.11+ because of CLAI. Run CLAI tests and coverage in the main CI workflow, with asyncio selected when the shared development environment includes Trio.

Related: https://github.com/pydantic/pydantic-ai-harness/issues/875.

Validated with focused Pyright, repository-wide Ruff, actionlint, lock checks, CLAI tests on Python 3.11 and 3.14, and focused Harness tests on Python 3.10 with lowest direct dependencies.

## AI Disclaimer

This PR was developed with the assistance of either Claude or Codex. I've reviewed and verified the changes.


.github/workflows/clai2.yml
```diff
@@ -23,10 +23,7 @@ jobs:
       - uses: astral-sh/setup-uv@cec208311dfd045dd5311c1add060b2062131d57 # v8.0.0
         with:
           python-version: ${{ matrix.python }}
-      - run: uv sync --locked
-      - run: uv run --no-sync ruff check .
-      - run: uv run --no-sync ruff format --check .
-      - run: uv run --no-sync pyright src tests
+      - run: uv sync --locked --package pydantic-clai2
```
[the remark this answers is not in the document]
Kludex: Updated both CLAI jobs to sync all workspace packages with the shared `dev` group. Verified from the CLAI subdirectory with a fresh Python 3.11 environment: `coverage` and `pytest` are installed, and all 362 tests pass.

.github/workflows/clai2.yml
Kludex, line 1: We don't need this.

pydantic-clai2/pyproject.toml
Kludex, line 10: Change the required version to >=3.10 to align with this project please.

tests/test_cassettes.py
```diff
@@ -74,7 +74,7 @@ def test_cassettes_discovered() -> None:
     assert sum(1 for _ in _cassettes()) >= 7
 
 
-@pytest.mark.parametrize('path', _cassettes())
+@pytest.mark.parametrize('path', list(_cassettes()))
```
Kludex: Why are we doing this "list()" everywhere? please remove those unnecessary changes.

pydantic-clai2/pyproject.toml
Kludex: I tried `>=3.10`: `uv lock --dry-run` fails because `termflow-md==0.9.2` requires Python >=3.11, and no newer release is available. Lowering CLAI alone therefore makes the whole workspace unresolvable. I recommend retaining CLAI 3.11 and Harness 3.10 for this PR; full CLAI 3.10 support needs a Termflow backport/replacement plus its own runtime compatibility changes. Would you prefer that separate backport instead?

tests/test_cassettes.py
Kludex: The unlocked Python 3.10 installs picked pytest 9.1, which rejects generator parameters under our warnings-as-errors setting. I will revert these unrelated test edits and keep the test runner aligned with the shared lock instead.

.github/workflows/clai2.yml
Kludex: Agreed. I will move the CLAI test and coverage jobs into the main CI workflow and remove the separate workflow, preserving the coverage check.

Kludex: Removed the separate workflow. CLAI tests and its coverage gate now run in the main workflow and are required by `check`.

tests/test_cassettes.py
Kludex: Reverted all four `list()` edits; those test files are unchanged in the PR. Python 3.10 compatibility and coverage now use the existing lowest-versions job, avoiding an unrelated latest-dependency upgrade in this workspace change.

.github/workflows/main.yml
```diff
@@ -157,7 +212,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.10', '3.11', '3.12', '3.13', '3.14']
+        python-version: ['3.11', '3.12', '3.13', '3.14']
```
Kludex: We should continue to support 3.10!

Kludex: Restored Python 3.10 to both the slim and all-extras matrix; the lowest-versions job remains too. These installs now use the shared lock as version ceilings plus the root provider overrides, allowing 3.10-compatible dependency fallbacks without unrelated upgrades. Verified the previously failing cases locally: 596 passed, 29 skipped.

.github/workflows/main.yml
```diff
@@ -217,11 +290,16 @@ jobs:
       # The lock validation must be off here: this job deviates from the lock
       # on purpose, and under UV_LOCKED=1 the lowest-direct resolution fails
       # because uv refuses to write the lowest-resolved lockfile it produces.
-      - run: uv sync --resolution lowest-direct --group dev --all-extras
+      - if: matrix.python-version != '3.10'
+        run: uv sync --resolution lowest-direct --group dev --all-extras
         env:
           UV_LOCKED: '0'
+      - if: matrix.python-version == '3.10'
```
Kludex, line 297: those bunch of 3.10 conditionals shouldn't be needed.

PR #932, opened 2026-09-18
Separate CLAI2 tool calls with blank lines and require tmux UX verification

mpfaffenberger:
## Summary
- Require testing added or modified CLAI2 UX features in a fresh tmux window with `uv run clai2`, checking the terminal behavior against the user request before opening a PR.
- Print a blank line after every tool-call summary, including shell, file tools, grep, and generic tool calls, regardless of the detailed-output setting.
- Keep detailed output and plugin rendering unchanged; update the user docs and regression assertions.

## Verification
- 407 CLAI2 tests passed.
- Ruff check and format check passed; Pyright reported zero errors.
- Launched the changed checkout using `uv run clai2` in a fresh tmux session/window (`biscuit-clai-spacing:ux`, 120x40).
- Verified sequential shell, list_files, read_file, grep, write_file, and edit_file summaries each have a following blank line, including before the final assistant response. File operations used a temporary file outside the repository.
- Enabled detailed output and verified shell output/completion and grep results retain correct spacing; restored summary mode afterward.

This updates the existing PR from the interrupted docs-only attempt; the code change was tested in tmux before this push.


mpfaffenberger:
Added the requested color inversion: tool names use pink ACCENT, arguments and bullet markers use grey MUTED. All built-in and generic tool headers now share a formatter; blank-line spacing is retained. Updated README and plugin docs. Verification: 421 tests pass, Ruff and Pyright pass. Launched `uv run clai2` with truecolor in fresh tmux window `biscuit-clai-colors:ux`; shell, read_file, and grep rendered with pink names and grey arguments. Captured terminal SGR confirms RGB 229/32/233 for names and 143/136/142 for arguments. Regression tests check every supported tool and generic names with detailed output both on and off.

PR #934, opened 2026-09-18
Stop printing the plugin list when closing `/plugins`

mpfaffenberger:
## Summary
- Return quietly from the `/plugins` menu instead of automatically running `/plugins list`.
- Keep explicit `/plugins list` output unchanged and menu actions applied immediately.
- Add regression tests for closing with and without changes, including an empty menu, and update the user docs.

## Verification
- Repository-wide Ruff format and lint checks passed.
- Focused strict Pyright passed with zero errors or warnings.
- Plugin menu and loader tests: 19 passed.
- Fresh tmux session running `uv run clai2` with an isolated settings database: verified Esc, Enter, Q, and Ctrl-C close without a listing; disabling a plugin persists without output on close; explicit `/plugins list` still prints the updated state.


mpfaffenberger:
CI note: the CLAI test job fails in all 14 parametrizations of `tests/test_compact_tools.py::test_tool_header_colors`, which expect two trailing newlines but receive one. I reproduced the same 14 failures in a detached worktree at the unchanged base commit `0dc5f034`, with `PYTHONPATH` pointing to that worktree’s CLAI source. This PR does not modify the renderer or those tests. The failure is pre-existing, so I am leaving unrelated tool-header behavior out of this menu fix.

Failing CI job: https://github.com/pydantic/pydantic-ai-harness/actions/runs/35352162557/job/105622638027

The 19 focused plugin-menu and loader tests, Ruff checks, strict Pyright, and fresh tmux checks all pass.

mpfaffenberger:
Fixed the pre-existing failing checks in the follow-up commit. The renderer already emits two trailing newlines; Rich `Text.from_ansi()` removes the final newline. The test now asserts spacing on the raw terminal output and retains all decoded color assertions. No rendering behavior changed. All 40 focused compact-tool, plugin-menu, and plugin-loader tests pass, as do repository-wide Ruff checks and focused strict Pyright.

Issue #944, opened 2026-09-18
Restore sub-agent delegation to the default `Coder`, behind a flag


DouweM:
`Coder` stopped bundling `SubAgents` in #870, which removed both planning and delegation. Planning should stay out. Delegation should come back, behind a boolean defaulting to `True`.

The reasoning is backward compatibility rather than benchmarks. The bar a change to a public default has to clear is being reasonably confident that across general use cases it works out for the better. Planning clears that -- it was worth a couple of percent at most. Delegation does not: it was removed without that confidence.

### Scope

- `Coder` gains `sub_agents: bool = True`.
- The delegate is the same `Coder` the caller configured, built with `sub_agents=False`. That flag is what terminates the recursion.
- Composition order follows the docs: instructions, file system, shell, repo context, sub-agents, then the plumbing the agent never calls directly. `RepairToolArguments` moves last; it hooks a different part of the lifecycle from everything else in the chain, so its position in our own list carries no meaning, and it needs no `get_ordering`. Where ordering matters is composition with a user capability overriding the same hooks, and there the user picks.
- `docs/coder.md` and `pydantic_ai_harness/coder/README.md` stay in sync, and the "no magic" sections on the harness README and docs home regain `SubAgents`.

Decided in the 2026-09-18 maintainer huddle.

Issue #947, opened 2026-09-18
Host security capabilities do not reach sub-agent runs, so `delegate_task` bypasses parent-level tool policy


DouweM:
### What happens

A capability bound at the agent level alongside a harness -- an approval gate, a tool guardrail, an audit hook -- applies to the parent run. A delegation is a separate run, and the parent's other capabilities are not forwarded into it. The parent-level hook therefore sees the `delegate_task` call and none of the tool calls the delegate makes inside it.

Measured with a capability implementing `before_tool_execute`, bound next to `Coder` on the same agent:

```
sub_agents=False: guardrail saw ['shell']
sub_agents=True:  guardrail saw ['delegate_task']
```

In the second case the delegate ran a shell command the parent-level hook never observed and could not refuse.

### Why it matters more now

This follows from sub-agent isolation rather than from any one harness, and `docs/subagents.md` already documents that tools contributed by the parent's capabilities are not inherited, pointing at `shared_capabilities`. What changed is the default: #946 gives the default `Coder` a delegate that is itself a full `Coder`, with unrestricted shell and filesystem access, so any host relying on parent-level tool policy now has a gap unless it passes `sub_agents=False`.

For contrast, the delegate `Coder` carried before #870 was `FileSystem(read_only=True)` plus `RepoContext` -- no shell and no writes -- so the same gap existed but had far less reach.

### The awkward part

A capability cannot see what the host bound next to it. `SubAgents.shared_capabilities` is the right mechanism, but `Coder` has no way to discover an approval capability a user registered alongside it and forward it. Closing this properly likely needs core to expose sibling or run-level capabilities to a capability building a child run, rather than each harness reimplementing propagation.

### Options

- Document the boundary and leave propagation to the host (#946 does the documenting).
- Give `Coder` a way to accept capabilities to forward, e.g. `sub_agent_capabilities=[...]` passed through to `shared_capabilities`.
- Propose the core change so any capability composing a child run can forward the host's security capabilities by default.

Raised by automated review on #946 and confirmed by measurement.

DouweM:
### Delegating to the running agent closes this, and is worth considering over forwarding

The framing that matters: people expect guardrails added to the top-level agent to cascade into delegated runs, and they do not. Forwarding capabilities cannot fix that in general -- `SubAgentToolset.get_wrapper_toolset` already documents why:

[...]

Issue #945, opened 2026-09-18
`Coder`'s sub-agent delegate inherits `repo_context=False` and loses the repository instructions


DouweM:
### What happens

`Coder(repo_context=False)` passes that flag through to its sub-agent delegate, so the delegate is built without `RepoContext` and never sees `AGENTS.md` or the repository structure.

### Why the pass-through is wrong here

`repo_context=False` exists for one reason: a host that binds its own `RepoContext` at the agent level would otherwise load the instruction files twice. That double-load cannot happen in a sub-agent run. Each delegation is a fresh, isolated run of a separate `Agent`, and the host's `RepoContext` is bound to the parent agent, not to the delegate. So the flag suppresses a conflict that does not exist there, and the only effect is that the delegate works without the repository's own instructions.

Only the top-level `Coder` is the one that might have its `RepoContext` slotted out by a host.

### Where it bites

`pydantic-clai2` is the case in the tree today. `DEFAULT_PLUGINS` in `pydantic-clai2/src/pydantic_clai2/_app.py` registers:

```python
PluginSettings(
    id='coder',
    factory='pydantic_ai_harness.coder:Coder',
    settings={'unrestricted_filesystem': True, 'repo_context': False},
)
```

because clai2's own `repo_context` plugin binds one. The parent agent gets the repository instructions; every delegation runs without them.

### Fix

Build the delegate with `repo_context=True` unconditionally, rather than inheriting the caller's value. `docs/coder.md` and `pydantic_ai_harness/coder/README.md` both describe the delegate as "the same `Coder` you configured", so that sentence needs the exception spelled out.

Introduced in #944.

DouweM:
@mpfaffenberger will be removing the standalone `repo_context` plugin from clai2 and the `repo_context` flag from `Coder`.

PR #950, opened 2026-09-18
Restore sub-agent delegation to the default `Coder` behind a `sub_agents` flag

DouweM:
## Summary

Replaces #946, which bundled this with the composition reorder and docs fixes. Those are now #949, and this targets that branch so the diff here is delegation only. It will retarget to `main` once #949 merges.

> [!IMPORTANT]
> **Not ready to merge.** See "The open question" below: what top-level policy should mean for a sub-agent run (#947) is unresolved, and this PR decides it by default in a direction nobody is confident about yet. Parked for the delegation-rules discussion.

`Coder` stopped bundling `SubAgents` in #870, which removed planning and delegation together. Planning stays out. This puts delegation back behind `sub_agents: bool = True`.

**The delegate is the same `Coder` the caller configured** -- same workspace, `instructions=`, and filesystem scope -- built with `sub_agents=False`. That flag is what terminates the recursion. It carries no model of its own, so each delegation runs on the parent run's model, and `agent_folders=None` keeps the roster to that one delegate rather than picking up disk agents.

## The open question

A capability the host binds next to `Coder` -- an approval gate, a tool guardrail, an audit hook -- applies to the parent run. A delegation is a separate run, and the parent's other capabilities are not forwarded into it. Measured with a `before_tool_execute` capability bound alongside `Coder`:

```
sub_agents=False: guardrail saw ['shell']
sub_agents=True:  guardrail saw ['delegate_task']
```

The delegate ran a command the parent-level hook never observed and could not refuse.

This is not incidental. `SubAgentToolset.get_wrapper_toolset` documents why forwarding cannot fix it in general: capability toolsets are bound to capability instances registered in the parent run, so they cannot be carried into a foreign run. `shared_capabilities` is the supported mechanism, but `Coder` cannot discover a guardrail a user registered beside it.

Two things make this worth pausing on rather than documenting around:

- **The delegate here is not what #870 removed.** That was a read-only `explorer` (`FileSystem(read_only=True)` plus `RepoContext`, no shell, no writes). A recursive `Coder` with unrestricted shell has a different risk profile, so the backward-compatibility argument that motivated restoring delegation does not by itself cover this delegate.
- **Delegating to the running agent would close it.** A run of the same agent re-registers the same capabilities, so the hooks fire. Measured: `['delegate_task', 'shell']`. It needs an explicit depth cap (self-delegation recursed to depth 11 unbounded; a `ContextVar` held it at 2) and an answer for what should not cascade -- `AskUser`, `output_type`, the system-prompt listing. Recorded on #947.

The Sub-agents section in the docs states the boundary and names `sub_agents=False` as the way out.

## Also here

- `## Six tools` becomes `## Tools` with a `delegate_task` row, and the stale "Coder does not include planning, delegation, or the run-scoped `run_command` family" line is corrected.
- Both "no magic, it's capabilities all the way down" sections and both Harnesses table rows regain `SubAgents`.
- `TERMINAL_BENCH.md` notes that later releases benchmark a different composition, and that `sub_agents=False` reproduces the six-tool generation.

## Tests

`test_the_delegate_does_not_delegate_further` drives a real `delegate_task` call through a `FunctionModel` and records the tool set of every request. The delegate carries no model, so it runs on the parent's and both requests land in the same function; the test asserts the sub-agent's run sees `edit_file` but not `delegate_task`. The recursion guard is tested rather than assumed.

`test_coder_and_researcher_compose` expected the combined roster to be `['researcher']`, which held only because `Coder` had no `SubAgents` to contribute. Both harnesses bring one now, so they combine into a single capability carrying each one's delegate.

Full suite: 6,632 passed, 72 skipped. Ruff and Pyright clean.

## Known, not fixed here

- #945: the delegate inherits `repo_context=False` and so loses the repository instructions, which affects `pydantic-clai2` today.
- `CombinedCapability([Coder(), Coder()])` raises `Duplicate sub-agent name 'coder'`. The pre-#870 `Coder` collided the same way with its `explorer`, and the error names the fix.

## Linked Issue

Closes #944

## Checklist


DouweM:
Superseded by #1061: the delegate is now the running agent (`SubAgents(include_self=True)`), so capabilities bound next to `Coder` also cover delegated runs (#947).

Issue #966, opened 2026-09-19
Complete the remaining `CLAI2` local usability checklist


Kludex:
Implement the remaining local checklist items as stacked PRs:

- [ ] Correct fenced code highlighting.
- [ ] Add a `/theme` command.
- [ ] Ship MCP support as a built-in plugin.
- [ ] Offer a web interface alternative.
- [ ] Ship native OS notifications as a built-in plugin, enabled by default, including macOS.
- [ ] Keep conversation context visible during questions and replace Space + Enter selection.
- [ ] Show update availability notices.

The screenshot item is excluded. Existing completed items (resume, reload, cancellation, and the persistent editor) are not part of this stack.

Kludex:
The local `kludex/clai2-builtin-mcp` implementation now references this issue from `pydantic-clai2/pyproject.toml` for the built-in MCP runtime dependency. The plugin must remain usable without an extra install, while default startup/status must not read or connect to repository MCP configuration. Outer AnyIO cancellation has a core-only subprocess-leak reproducer tracked in https://github.com/pydantic/pydantic-ai/issues/8548; the local implementation retains a strict xfail and documents the limitation rather than duplicating core cleanup. No branch push or PR yet.

Kludex:
The first four stacked PRs are implemented and pass CI: [highlighting](https://github.com/pydantic/pydantic-ai-harness/pull/967), [inline questions](https://github.com/pydantic/pydantic-ai-harness/pull/970), [native notifications](https://github.com/pydantic/pydantic-ai-harness/pull/971), and [themes](https://github.com/pydantic/pydantic-ai-harness/pull/972). Review findings are addressed. MCP, update notices, and the browser interface remain in progress; MCP will be a built-in plugin.

Kludex:
All seven stacked Harness PRs have been implemented and pass CI: [highlighting](https://github.com/pydantic/pydantic-ai-harness/pull/967), [inline questions](https://github.com/pydantic/pydantic-ai-harness/pull/970), [native notifications](https://github.com/pydantic/pydantic-ai-harness/pull/971), [themes](https://github.com/pydantic/pydantic-ai-harness/pull/972), [MCP](https://github.com/pydantic/pydantic-ai-harness/pull/974), [update notices](https://github.com/pydantic/pydantic-ai-harness/pull/975), and [browser UI](https://github.com/pydantic/pydantic-ai-harness/pull/977).

[...]

PR #977, opened 2026-09-19
Offer the core browser interface for configured `CLAI2` agents

Kludex:
Part 7 of the CLAI2 checklist stack, based on [#975](https://github.com/pydantic/pydantic-ai-harness/pull/975). Refs #966, including the optional web dependencies.

Add `clai2 --web` using Pydantic AI's existing browser UI and Uvicorn, bound only to `127.0.0.1`. Preserve model/configuration precedence, saved model settings, tool retries, credentials, and compatible plugin capabilities. Agent runs share one execution slot; resources close in the ASGI lifespan before Uvicorn re-raises Ctrl-C.

Terminal-only plugins are omitted without changing preferences. Unsupported turn hooks reject startup rather than losing guards. Browser history is separate from CLAI SQLite sessions, and core's current web API uses its default 50-request limit; unsupported resume/request-limit options fail explicitly. A real HTTP/stdio test verifies cleanup when the browser disconnects during a tool call. The separate upstream cancellation limitation remains documented for direct `Agent.run()` embeddings.

Verification: all 655 CLAI tests pass with the documented MCP xfail, including real HTTP streams, configured plugin tools, serialization, host/content-type/CORS checks, startup failures, and real SIGINT teardown. Ruff and CLAI Pyright pass. Fresh tmux startup and headless Chrome checks verified desktop/mobile UI, real streamed replies and plugin tool output, no JavaScript errors or mobile overflow, and async plugin cleanup on Ctrl-C.

## AI Disclaimer

This PR was developed with the assistance of either Claude or Codex. I've reviewed and verified the changes.


Kludex:
The requested browser surface is opt-in and reuses core rather than adding another agent runtime. The initial head passed all CI checks and review raised no correctness findings; the follow-up adds a real HTTP/MCP disconnection regression and clarifies that the known direct Agent.run cancellation issue does not reproduce on this streaming path.

Kludex:
The browser/MCP integration regression now lives in `pydantic-clai2/tests/mcp/` alongside the other MCP tests. All 22 focused tests pass, with the separate known core cancellation xfail unchanged.

PR #987, opened 2026-09-20
Expose stable `Planning` IDs and reject unknown fields

### LINKED ISSUE #986: `Planning` hides generated step IDs after accepting unknown fields
## Bug

`Planning.write_plan` can accept a model-generated `task_id` field on each item, silently discard it, generate a different internal id, and then render only ordinal numbers in both the tool result and the tail reminder.

That combination makes the ordinal look like the id accepted by `update_task_status`.

## Observed failure

In a real agent trace, the model called `write_plan` with four items carrying `task_id` values `"1"` through `"4"`. The call succeeded and returned:

```text
Plan updated: 4 step(s).

1. [~] ...
2. [ ] ...
3. [ ] ...
4. [ ] ...
```

The store had generated unrelated eight-character ids. The model then called:

```json
{"task_id":"1","status":"completed"}
{"task_id":"2","status":"in_progress"}
```

Both calls returned `Step with id '...' not found.` The model retried the same ordinal ids later and failed again. Its reasoning explicitly treated the reminder's `1`, `2`, `3`, and `4` as ids.

## Root cause

`PlanItem` uses Pydantic's default extra-field behavior, so the unexpected `task_id` is ignored at runtime. Separately, the compact renderer used by `write_plan` and the injected reminder omits `PlanItem.id`, even though granular tools require it.

## Expected behavior

- Unknown fields in plan input objects should produce a tool-validation retry instead of being discarded.
- Every model-facing plan rendering should show each step's stable id.
- A regression test should cover the full `write_plan` validation, generated-id return, and `update_task_status` flow through an `Agent`.

tigrisa:
## What

- reject unknown fields in `PlanItem` and `PlanStatusUpdate` instead of silently discarding them
- include each stable step id in the compact plan rendering used by `write_plan` results and tail reminders
- update model guidance and both Planning docs to describe the ID contract
- add an end-to-end `Agent` regression that exercises validation retry, generated-ID discovery, and `update_task_status`

## Real failure

A downstream agent trace showed this sequence:

1. The model called `write_plan` with four items carrying `task_id` values `"1"` through `"4"`.
2. `PlanItem` silently ignored those unknown fields and generated unrelated eight-character ids.
3. `write_plan` returned only ordinal lines such as `1. [~] ...`; the tail reminder used the same rendering.
4. The model explicitly interpreted `1`, `2`, `3`, and `4` as ids and called `update_task_status` with `"1"` and `"2"`.
5. The updates returned `Step with id '1' not found.` and `Step with id '2' not found.` The model retried the same values and failed again.

This was a contract mismatch, not just a weak model choice: input validation accepted a field that looked like the downstream identifier, while the two automatic outputs hid the generated identifier that downstream tools require.

## Why this fix

`ConfigDict(extra='forbid')` turns the malformed `write_plan` call into Pydantic AI's normal tool-validation retry. Including `[item.id]` in the shared compact renderer makes the generated id visible immediately and keeps the `write_plan` result and automatic reminder consistent. Ordinal numbers remain as display positions only.

No storage format, tool name, or public function signature changes.

## Verification

- `uv run --no-sync pytest -p no:cacheprovider tests/planning` (`148 passed`)
- `uv run --no-sync pytest -p no:cacheprovider tests/test_doc_snippets.py tests/test_docs_parity.py` (`817 passed, 22 skipped`)
- `uv run --no-sync ruff format --check .`
- `uv run --no-sync ruff check .`
- `PYRIGHT_PYTHON_IGNORE_WARNINGS=1 uv run --no-sync pyright pydantic_ai_harness/planning tests/planning`
- required docs parity reviewer: PASS

Fixes #986


tigrisa:
Thanks for applying the review suggestion. I think this change conflicts with the scope we agreed in #986: strict validation and immediate `write_plan` ID visibility stay together, while reminder ID visibility is considered separately.

The current implementation, docs, and end-to-end test keep the reminder ordinal-only, but commit `620f7878` changes the unit test to require `[step-x]`, so the Planning suite will fail once CI is approved.

May I revert just that assertion to `assert 'step-x' not in reminder_text` and push a follow-up? If you intended to revisit reminder visibility instead, I can update the implementation and docs to match.

mpfaffenberger:
> Thanks for applying the review suggestion. I think this change conflicts with the scope we agreed in #986: strict validation and immediate `write_plan` ID visibility stay together, while reminder ID visibility is considered separately.
> 
> The current implementation, docs, and end-to-end test keep the reminder ordinal-only, but commit `620f7878` changes the unit test to require `[step-x]`, so the Planning suite will fail once CI is approved.
> 
> May I revert just that assertion to `assert 'step-x' not in reminder_text` and push a follow-up? If you intended to revisit reminder visibility instead, I can update the implementation and docs to match.

yes

